PR #8, opened 2025-06-16
Add Support for NVIDIA AI Workbench

nv-edwli:
Adds AIWB-specific specification files
  * /.project/ directory
  * /requirements.txt
Adds AIWB as an option under /deploy/
  * workbench/docker-compose.yaml
  * workbench/nginx.conf.template
  * workbench/README.md

### DIFF
--- a/deploy/workbench/README.md
+++ b/deploy/workbench/README.md
@@ -1,0 +1,71 @@
+# Quickstart for NVIDIA AI Workbench 
+
+The AI-Q NVIDIA Research Assistant blueprint allows you to create a deep research assistant that can run on-premise, allowing anyone to create detailed research reports using on-premise data and web search. 
+
+> **Note**
+> This app runs in [NVIDIA AI Workbench](https://docs.nvidia.com/ai-workbench/user-guide/latest/overview/introduction.html). It's a free, lightweight developer platform that you can run on your own systems to get up and running with complex AI applications and workloads in a short amount of time. 
+
+> You may want to [**fork**](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo#forking-a-repository) this repository into your own account before proceeding. Otherwise you won't be able to fully push any changes you make because this NVIDIA-owned repository is **read-only**.
+
+*Navigating the README*: [Project Overview](#project-overview) | [Get Started](#get-started) | [Customize](#customization) | [License](#license)
+
+*Other Resources*: [:arrow_down: Download AI Workbench](https://www.nvidia.com/en-us/deep-learning-ai/solutions/data-science/workbench/) | [:book: User Guide](https://docs.nvidia.com/ai-workbench/) |[:open_file_folder: Other Projects](https://docs.nvidia.com/ai-workbench/user-guide/latest/quickstart/example-projects.html) | :rotating_light: User Forum (Coming Soon)
+
+## Project Overview
+
+The AI-Q NVIDIA Research Assistant blueprint allows you to create a deep research assistant that can run on-premise, allowing anyone to create detailed research reports using on-premise data and web search. 
+
+The main research agent is written in LangGraph and managed using NVIDIA Agent Intelligence Toolkit (AIQ). The research agent provides a unique deep research capability with these features:
+
+- **Deep Research**: Given a report topic and desired report structure, an agent (1) creates a report plan, (2) searches data sources for answers, (3) writes a report, (4) reflects on gaps in the report for further queries, (5) finishes a report with a list of sources.
+- **Parallel Search**: During the research phase, multiple research questions are searched in parallel. For each query, the RAG service is consulted and an LLM-as-a-judge is used to check the relevancy of the results. If more information is needed, a fallback web search is performed. This search approach ensures internal documents are given preference over generic web results while maintaining accuracy. Performing query search in parallel allows for many data sources to be consulted in an efficient manner.
+- **Human-in-the-loop**: Human feedback on the report plan, interactive report edits, and Q&A with the final report.
+- **Data Sources**: Integration with the NVIDIA RAG blueprint to search multimodal documents with text, charts, and tables. Optional web search through Tavily.
+- **Demo Web Application**: Frontend web application showcasing end-to-end use of the AI-Q Research Assistant.
+
+[Read More](../README.md)
+
+## Get Started
+
+Ensure you have satisfied the prerequisites for this Blueprint ([details](../README.md)). 
+
+You should have the [RAG Blueprint](https://github.com/NVIDIA-AI-Blueprints/rag) up and running already. This deep researcher will wrap around your RAG pipeline. 
+
+1. Open NVIDIA AI Workbench. Select a **Location** to work in.
+
+1. **Clone** the project using the repository URL: https://github.com/NVIDIA-AI-Blueprints/aiq-research-assistant. 
+
+1. On the **Project Dashboard**, resolve the yellow unconfigured secrets warning:
+
+   * ``NVIDIA_API_KEY``: NVIDIA API Key generated from build.nvidia.com or NGC (starts with "nvapi-")
+   * ``TAVILY_API_KEY``: Tavily API Key for web search tool calling (starts with "tvly-")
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -0,0 +1 @@
+jupyterlab>3.0
```
slopp, line 1: Does this need to be in the root for workbench? I am worried it will confuse users setting up the project for the first time. If possible I'd love to move it to the workbench directory. If that isn't possible, lets at least add a comment that this is specific to workbench, pointing to the pyproject.toml for the actual project dependencies.

deploy/workbench/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+* Use on-prem NIMs vs. cloud-hosted models
+* Swap models names/endpoints
+* Custom endpoint routes
+* Adjust default report organization prompts
+* And more!
+
+To customize this blueprint, adjust the ``configs`` section under ``deploy/workbench/docker-compose.yaml`` in a code editor and save your changes. When restarting the compose, any configuration changes you make will take effect. 
+
+## License
```
slopp, line 67: Please update to have the same license section as the README

deploy/workbench/docker-compose.yaml
```diff
@@ -0,0 +1,297 @@
+configs:
```
slopp, line 1: Why are we embedding the configs in the compose file? Is this strictly necessary for workbench?

I would prefer if we can mount the config files similar to the existing docker compose for two reasons:

- going forward, we will need to update the configs as we adjust the features of the deep researcher, and I want to avoid the number of places we need to re-write those configs
- users reading any of the non-workbench docs (AIQ / docker getting started / etc) are going to be referred to those config files, it is not intuitive to have separate configs embedded here

We can still use two profiles for easier selection in the workbench, just the AIRA_HOSTED_NIMs env var and the working directory / volume paths would need to change.

slopp (changes requested):
Thank you so much for adding this, I am excited to have workbench support! I believe we can get this into the upcoming release.

I've tagged Allan and Linette for quick reviews as well. 

My main concern is maintaining the workbench config going forward. I left a comments to see if we could simplify the config approach to make that easier. Let me know what is possible with workbench.

requirements.txt
nv-edwli: Unfortunately yes, AIWB does require this to be at the root level of the project repo. 

I can add the following header to the top of the file to make it clear for BP users whether they need this or not. 

``# ==== For NVIDIA AI WORKBENCH Users Only! ====``
``# See pyproject.toml for project dependencies``

deploy/workbench/README.md
nv-edwli: Ack, will do

README.md
```diff
@@ -4,6 +4,10 @@
 
 The AI-Q NVIDIA Research Assistant blueprint allows you to create a deep research assistant that can run on-premise, allowing anyone to create detailed research reports using on-premise data and web search. 
 
+| :exclamation: Important |
```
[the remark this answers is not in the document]
nv-edwli: Ack, will do

deploy/workbench/README.md
```diff
@@ -0,0 +1,69 @@
+# Quickstart for NVIDIA AI Workbench 
+
+The AI-Q NVIDIA Research Assistant blueprint allows you to create a deep research assistant that can run on-premise, allowing anyone to create detailed research reports using on-premise data and web search. 
+
+> **Note**
+> This app runs in [NVIDIA AI Workbench](https://docs.nvidia.com/ai-workbench/user-guide/latest/overview/introduction.html). It's a free, lightweight developer platform that you can run on your own systems to get up and running with complex AI applications and workloads in a short amount of time.
```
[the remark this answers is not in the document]
nv-edwli: Sorry, this should say "Blueprint"--app is a remnant of a previous non-bp project example. 

Thanks for the catch, will update.

deploy/workbench/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+> **Note**
+> This app runs in [NVIDIA AI Workbench](https://docs.nvidia.com/ai-workbench/user-guide/latest/overview/introduction.html). It's a free, lightweight developer platform that you can run on your own systems to get up and running with complex AI applications and workloads in a short amount of time. 
+
+> You may want to [**fork**](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo#forking-a-repository) this repository into your own account before proceeding. Otherwise you won't be able to fully push any changes you make because this NVIDIA-owned repository is **read-only**.
+
+*Navigating the README*: [Project Overview](#project-overview) | [Get Started](#get-started) | [Customize](#customization) | [License](#license)
+
+*Other Resources*: [:arrow_down: Download AI Workbench](https://www.nvidia.com/en-us/deep-learning-ai/solutions/data-science/workbench/) | [:book: User Guide](https://docs.nvidia.com/ai-workbench/) |[:open_file_folder: Other Projects](https://docs.nvidia.com/ai-workbench/user-guide/latest/quickstart/example-projects.html) | :rotating_light: User Forum (Coming Soon)
```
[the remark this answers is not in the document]
nv-edwli: Actually, I plan to have this Developer Forum link ready by release. I just haven't created the topic yet. 

Technically, I can just create the topic thread now and add the live link here, but since the post will be public it may be confusing for users who see the thread on the Forum but can't pull this BP to their AIWB yet. 

Let me know if this is the recommended course of action and I can update this section with the working link.

deploy/workbench/docker-compose.yaml
```diff
@@ -0,0 +1,297 @@
...
+  aira-backend-gpu: 
+    <<: *aira-backend-no-gpu
+    environment:
+      AIRA_HOSTED_NIMS: ${AIRA_HOSTED_NIMS:-false}
+    configs:
+      - source: backend_config_gpu
+        target: /app/configs/config.yml
+    profiles: ["aira-gpu"]
+
+  aira-nginx-no-gpu: &aira-nginx-no-gpu
```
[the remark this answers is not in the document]
nv-edwli: Good call, aligned with this. Will update shortly to consolidate.

deploy/workbench/docker-compose.yaml
nv-edwli: Ah yes, so for context I had initially based this implementation off of the implementation of the VSS Blueprint [here](https://github.com/rmkraus/video-search-and-summarization/blob/main/workbench/compose.yaml). You can see that BP is custom config-heavy and uses a similar config file buried under ``/src/vss-engine/config/*.yaml``, so embedding the config simplifies the user input which conditions which services to run and works around the potential headaches of users needing to go into the src code to manually edit a config file. 

Taking a step back, I'm generally not exactly a huge fan of users needing to do a bunch of manual configurations across a bunch of different places and/or running a bunch of commands before they can get to something usable, eg. compose files, env files, pip installs, config files, etc. so I tend to use AIWB to err on the side of simplifying as much of the setup as reasonably possible for the end user.

In my mind, the ideal experience looks something like:

1. User clones the project to their workbench
2. User adds their API keys + RAG location from the project dashboard when prompted.
3. User selects a compose profile they want to run from the dashboard, either if they have hardware to run with or they don't
4. User starts the compose
5. User opens the UI in a browser and gets started w deep research

I was previously operating under the assumption that the user _needed_ to adjust this config file manually in order to get started which is what I wanted to avoid by embedding it in the compose to utilize variable interpolation, etc. , but I'm seeing that is not actually the case. We can just mount the file as-is and provide a happy path to something usable. In that case, I'm happy to revert.

deploy/workbench/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Customization
+
+You can further customize the blueprint as follows:
+
+* Use on-prem NIMs vs. cloud-hosted models
+* Swap models names/endpoints
+* Custom endpoint routes
+* Adjust default report organization prompts
+* And more!
```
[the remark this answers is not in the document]
nv-edwli: Having a single bullet point in the list doesn't really make sense in the context of this section, is there any way we can reword/consolidate?

slopp (approved):
Thanks! lgtm, @nv-edwli  can you merge today? we're hoping to get a QA RC out by early next week

PR #29, opened 2025-08-28
Replace nginx with custom fastapi routes

efajardo-nv:


AjayThorve:
This looks great, verified it works with aira-frontend-v1.2.0.

Thanks @efajardo-nv

PR #38, opened 2025-09-17
Updating Depreciated AIQ API to newer NAT API

KyleZheng1284:
Updated all files that used the deprecated AIQ CLI to use NAT CLI due to AIQ Deep Research Assistant using NAT 1.2.
Added more detailed documentation on adding custom evaluator.
Minor fix in the ragas_wrapper evaluator to deal with an edge case of seeing an null due to rate limiting

### DIFF
--- a/deploy/entrypoint.sh
+++ b/deploy/entrypoint.sh
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 if [ -z "$AIRA_HOSTED_NIMS" -o "$AIRA_HOSTED_NIMS" = "false" ]; then
-    exec uv run aiq serve --config_file /app/configs/config.yml --host 0.0.0.0 --port 3838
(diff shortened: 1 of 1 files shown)


configs/eval_config.yml
```diff
@@ -57,6 +57,7 @@ llms:
   # that uses local NVIDIA NIM microservices.
   # Update if you are deploying differently or using hosted NVIDIA NIM microservices.
   instruct_llm:
+    _type: nim
```
efajardo-nv, line 60: Duplicate line

efajardo-nv (approved):
LGTM. I built a new backend container from this PR branch, deployed using hosted NIMs, and successfully generated a report.

AjayThorve (changes requested):
https://github.com/KyleZheng1284/aiq-research-assistant/blob/aiq_cli_to_nat/deploy/entrypoint.sh

PR #44, opened 2025-09-29
Bulk upload updates

efajardo-nv:
- Update bulk upload readme to include steps for setting `RAG_INGEST_URL` for helm deployment
- Rename `sync_files2.py` to `zip_to_collection.py` to make it more clear what the script is doing, i.e. syncing zip files to new/existing collections in RAG.
- Script rename will require publishing an updated `load-files` container.
- Update ubuntu base image of `aira-load-files` container to latest version `jammy-20250619`
- Fix syntax error in helm deployment instructions for selecting NIM model profile


data/readme.md
```diff
@@ -26,11 +26,19 @@ docker run \
 
 ## Bulk Upload via Python 
 
-Set the following environment variables based on your RAG deployment:
+Set the `RAG_INGEST_URL` environment variable based on your RAG deployment.
 
+Docker Compose:
 ```bash
 RAG_INGEST_URL="http://ingestor-server:8082" # URL for RAG ingestion server
```
JoyceChenNV, line 33: can't use http://ingestor-server:8082 for docker deploy as well because ingestor-server is a container name.
As we run the script in local host, need to set as export RAG_INGEST_URL="http://localhost:8082

PR #75, opened 2025-12-15
NIM Operator Support

angudadev:


mkhaas:
LGTM

docs/get-started/get-started-helm.md
```diff
@@ -91,6 +91,33 @@ helm install aiq-aira aiq-aira/ \
 --set backendEnvVars.RAG_INGEST_URL=<INGESTOR_SERVER_URL> -n aiq
 ```
 
+##### Deploy with NIM Operator 
+
+The [NVIDIA nim-operator 3.0 and above](https://docs.nvidia.com/nim-operator/latest/install.html) must also be installed and configured in your cluster to ensure that the Nvidia NIM LLM is properly deployed.
```
slopp, line 96: Is this optional or required? 

Why are we installing just one of the NIMs via NIM operator rather than all of them?

slopp (changes requested):

docs/get-started/get-started-helm.md
angudadev: > Is this optional or required?
NIM Operator is a additional option here as depedencies like RAG and nvingest are by default only NIM Operator Helm Deployments with upcoming releases. 

> Why are we installing just one of the NIMs via NIM operator rather than all of them?
because other NIMs are installed from RAG

slopp: ok makes sense

should we wait to land this until the RAG release that uses NIM operator by default? do we need to make any changes to how we pull in the RAG stuff?

angudadev: I agree with that, we can wait for the RAG release so that AIQ will be aligned with NIM Operator. 

At that point NIM Operator is not optional as it will be prerequisite. I will verify the changes and let you know

PR #97, opened 2026-02-25
AIQ UI session storage fixes

exactlyallan:
### Migration Note

Existing `localStorage` session data is not migrated. Dev users should **clear browser storage** (`aiq-chat-store`) before running this update.

### Problem

The chat UI persists session data to `localStorage` (5MB browser quota). Two critical issues were causing excessive storage consumption:

1. **Pruning was dead code on the happy path.** `prunePersistedChatState` — which strips heavy, refetchable fields like `thinkingSteps.content`, `rawPayload`, `reportContent`, `citations`, and all deep research execution data — was only invoked inside a `QuotaExceededError` catch block. Every normal write stored the full un-pruned conversation data, including verbose backend payloads that are never displayed in the UI.

2. **`currentConversation` was serialized twice.** The active session's full conversation object was stored both in the `conversations[]` array and as a separate `currentConversation` field, doubling the storage footprint of the active session when JSON-serialized.

3. **No proactive cleanup.** Old/stale sessions accumulated without bound until the quota hard limit was hit, at which point all sessions were cleared as a last resort.

#### Always-on message pruning (`store.ts`)
- `prunePersistedChatState` now runs on **every** `setItem` call, not just on quota exceeded
- Strips from each message before write:
  - `reportContent`, `citations`, `deepResearchTodos/LLMSteps/Agents/ToolCalls/Files`, `intermediateSteps` — destructured away entirely
  - `thinkingSteps[].content` → set to `''`, `rawPayload` → excluded
  - Deep research thinking steps (`isDeepResearch=true`) → filtered out
  - `planMessages[].text` → capped to 10k chars, `userResponse` → capped to 2k
- `currentConversation` stored as ID string reference; reconstructed from `conversations[]` on read

#### Tiered storage cleanup (`storage-manager.ts`)
- New `ensureStorageCapacity` runs proactively on session create, switch, and after agent responses
- Two-tier cleanup when storage exceeds warning threshold (4.2MB):
  - **Tier 1:** Delete sessions older than 1 day (any user)
  - **Tier 2:** Delete oldest sessions of the current user
- Protected sessions (current + active deep research jobs) are never deleted
- Max 10 deletions per cleanup pass with target of 3.5MB

#### Message pruning utilities (`prune-message-for-storage.ts`)
- `pruneMessageForStorage` — removes heavy refetchable fields, strips thinking step content
- `stripThinkingStepsForStorage` — keeps only display fields (id, displayName, functionName, category, timestamp, flags)
- `prunePlanMessages` — caps text content to prevent oversized plan data

#### UI updates
- SessionsPanel footer shows storage quota usage percentage
- Note text: *"Sessions and files are saved for a limited time before automatic deletion."*
- Minor CSS and component fixes

#### Housekeeping
- Removed unused Helm deployment charts (`deploy/helm/`)
- Updated notebooks and README
- Removed stale auth config test

### Test Results

All **1086 tests pass** across 66 test files (1 pre-existing skip), including:
- 15 tests for `prune-message-for-storage` (field stripping, content capping, rawPayload exclusion)
- 20 tests for `storage-manager` (tiered cleanup, capacity checks, protected sessions)
- 326 tests for the chat store and components

### DIFF
--- a/frontends/ui/src/features/chat/lib/prune-message-for-storage.spec.ts
+++ b/frontends/ui/src/features/chat/lib/prune-message-for-storage.spec.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { describe, test, expect } from 'vitest'
-import { pruneMessageForStorage, capString, pruneThinkingSteps, prunePlanMessages } from './prune-message-for-storage'
+import { pruneMessageForStorage, capString, stripThinkingStepsForStorage, prunePlanMessages } from './prune-message-for-storage'
 import type { ChatMessage } from '../types'
 
 describe('prune-message-for-storage', () => {
@@ -79,7 +79,6 @@ describe('prune-message-for-storage', () => {
 
       const pruned = pruneMessageForStorage(message)
 
-      // Essential fields kept
       expect(pruned.thinkingSteps).toBeDefined()
       expect(pruned.thinkingSteps).toHaveLength(1)
       expect(pruned.planMessages).toBeDefined()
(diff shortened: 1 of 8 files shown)


AjayThorve (approved):
LGTM

PR #109, opened 2026-02-27
Fix backend silently dropping WebSocket messages after DB pool failure

AjayThorve:
-   Add DB connectivity check to /health endpoint so Kubernetes liveness
-   probes restart the pod when the asyncpg connection pool gets stuck.
-   Add pool_recycle=1800 to SQLAlchemy engines to proactively replace
-   stale connections. Enable backend health probes in default helm values.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -91,6 +91,10 @@
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
     },
     {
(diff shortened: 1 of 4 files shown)


cdgamarose-nv:
Looks good to me!

cdgamarose-nv (approved):

PR #111, opened 2026-02-27
Strip dev dependencies from frontend prod container

drobison00:
Strip dev dependencies from frontend production container

Moves typescript from devDependencies to dependencies in package.json so it is available at runtime (required to parse next.config.ts), then adds a npm ci --omit=dev prune step to the Dockerfile builder stage immediately after npm run build.

This ensures the final production image only contains runtime dependencies (~380 packages vs ~823 previously), removing language-subtag-registry and all other dev-only packages from the shipped container.

Changes:

frontends/ui/package.json — move typescript@5.9.3 to dependencies
frontends/ui/package-lock.json — updated lockfile
frontends/ui/deploy/Dockerfile — add RUN npm ci --omit=dev --ignore-scripts after build step
Motivation: language-subtag-registry (CC0) and related ESLint tooling are dev-only dependencies pulled in transitively by eslint-config-next → eslint-plugin-jsx-a11y. Shipping them in the production container creates unnecessary license compliance overhead and bloat.


exactlyallan (approved):
Tested and ran locally and seems to work fine.

PR #113, opened 2026-03-02
Fix chat responses

pastorsj:
## What does this PR do?
* Switch from str response to ChatResponse return type


cdgamarose-nv (approved):
Works as expected. Good to merge

PR #114, opened 2026-03-02
Fix: Stale workflow responses leaking to reconnected WebSocket clients

exactlyallan:
### Problem

When a WebSocket connection is broken mid-request (network drop, browser
refresh, tab close), the frontend correctly marks the in-progress
`chatThinking` as **Interrupted**. However, responses from the original
request could still arrive — sometimes after a second `chatThinking`
cycle had already started — causing three distinct bugs:

1. **Ghost responses** — the stale workflow's final `SYSTEM_RESPONSE`
   arriving on the new socket, injecting a response belonging to the
   previous (cancelled) request.
2. **Interleaved thinking steps** — intermediate steps from the old
   workflow appearing inside a new request's `chatThinking` block.
3. **Premature stream termination** — the stale `COMPLETE` signal
   resetting `isStreaming = false`, cutting off the new request's
   in-progress thinking display.

---

### Root Cause

NAT's `WebSocketMessageHandler.process_workflow_request()` launches the
LLM workflow as a background `asyncio.create_task` and returns
immediately. When the socket disconnects, the `run()` loop exits — but
**the background task is never cancelled**. It keeps running and emitting
messages.

The `WebSocketSessionRegistry` (a module-level singleton) maps
`conversation_id → current WebSocket`. When the client reconnects with
the same `conversation_id`, the new socket replaces the old one in the
registry. The stale task's next `registry.send()` call finds the **new**
socket and delivers to it.

```
Client disconnects  →  run() loop exits  →  registry cleared
                                            ↕  task still running!
Client reconnects   →  new socket registered
Stale task resumes  →  registry.send() finds new socket  →  DELIVERED ❌
```

> **Note on intermediate steps:** NAT sets `parent_id` on
> `SYSTEM_INTERMEDIATE` messages using an **internal step ID**
> (`get_intermediate_step_parent_id()`), not the user message ID. This
> makes `parent_id`-based correlation impossible for thinking steps and
> is why the `isStreaming` guard is used instead.

---

#### Backend — `frontends/aiq_api/src/aiq_api/websocket_reconnect.py`

**1. Cancel workflow task on disconnect**

`run()` now calls `_cancel_running_workflow()` and
`registry.cancel_workflow_task()` on every disconnect:

```python
except (asyncio.CancelledError, WebSocketDisconnect):
    await _registry.clear_socket(self._conversation_id, self._socket)
    await _registry.cancel_workflow_task(self._conversation_id)
    self._cancel_running_workflow()
    break
```

**2. Registry-level task tracking**

`WebSocketSessionRegistry` gains `_workflow_tasks: dict[str, asyncio.Task]`
and two new methods:

- `set_workflow_task(conversation_id, task)` — registers the task;
  auto-cancels any previously registered task for the same conversation
  (handles the race where a reconnected client sends a new request before
  the old task is cancelled).
- `cancel_workflow_task(conversation_id)` — cancels and removes the task.

**3. Register task after launch**

After `super().process_workflow_request()` fires the NAT background task,
the new task is registered in the registry:

```python
task = self._running_workflow_task
if task is not None and not task.done():
    await _registry.set_workflow_task(conversation_id, task)
```

---

#### Frontend — `frontends/ui/src/adapters/api/websocket-client.ts`

- Added `activeParentId: string | null` — tracks the `id` of the last
  user message sent; updated on every `sendMessage()` call.
- `onResponse` and `onIntermediateStep` callbacks now receive an optional
  `parentId` argument populated from `message.parent_id`.

---

#### Frontend — `frontends/ui/src/features/chat/hooks/use-websocket-chat.ts`

Two independent stale-message guards:

| Guard | Applies to | Mechanism |
|---|---|---|
| `parent_id` correlation | `onResponse` | `SYSTEM_RESPONSE.parent_id` reliably equals the client's user message ID. Drops responses where `parent_id !== activeParentId`. |
| `isStreaming` guard | `onResponse` (`isFinal`) and `onIntermediateStep` | If `isStreaming` is `false` when a completion or step arrives, the request was already interrupted — discard it. |

---

### Tests

**`tests/aiq_agent/async_api/test_websocket_reconnect.py`** — 4 new tests:

| Test | Verifies |
|---|---|
| `test_handler_run_cancels_workflow_on_disconnect` | Disconnect cancels the in-flight workflow task |
| `test_registry_set_workflow_task_cancels_stale` | Registering a new task cancels the previous one for the same conversation |
| `test_registry_cancel_workflow_task` | `cancel_workflow_task()` cancels and removes the tracked task |
| `test_registry_cancel_workflow_task_noop_for_missing` | No-op for missing/`None` conversation IDs |

**`frontends/ui/src/features/chat/hooks/use-websocket-chat.spec.ts`** —
Updated 4 existing tests to set `mockStoreState.isStreaming = true` before
firing intermediate step callbacks, reflecting the real-app invariant that
`isStreaming` is always `true` while thinking steps are in flight.

---

### Test Results

| Suite | Result |
|---|---|
| `pytest` (Python, 703 collected) | ✅ 701 passed, 2 skipped |
| `npm test` (UI Vitest, 1126 collected) | ✅ 1125 passed, 1 skipped |
| `npm run lint` | ✅ 0 errors |
| `npm run type-check` | ✅ Pass |
| `pre-commit run --all-files` | ✅ All hooks passed |

---

### Defence-in-Depth Summary

The fix operates at two independent layers. Either one prevents the bug;
together they provide belt-and-suspenders coverage against the race window
between disconnect detection and task cancellation.

```
Layer 1 (Backend)
  Cancel asyncio.Task on disconnect
  → stale workflow stops producing messages immediately

Layer 2 (Frontend)
  isStreaming guard + parent_id guard
  → even if a stale message slips through the race window,
    it is silently discarded before touching UI state
```

### DIFF
--- a/docs/notebooks/1_Deep_Researcher_Web_Search.ipynb
+++ b/docs/notebooks/1_Deep_Researcher_Web_Search.ipynb
@@ -60,7 +60,19 @@
    "cell_type": "markdown",
    "id": "4629edb9",
    "metadata": {},
-   "source": "<a id=\"prereqs\"></a>\n## Prerequisites\n\n- **Python 3.11–3.13**\n- **[uv](https://github.com/astral-sh/uv)** package manager\n- **LLM API key** – For your chosen provider (required for LLM inference):\n    - **NVIDIA API Key** from [NVIDIA AI](https://build.nvidia.com/) (for NIM models)\nOptional requirements:\n- **Web search API key**:\n  - **Tavily** – [tavily.com](https://tavily.com) (for general web search)\n- **Node.js 22+ and npm** (optional, for web UI mode)"
+   "source": [
+    "<a id=\"prereqs\"></a>\n",
+    "## Prerequisites\n",
+    "\n",
+    "- **Python 3.11–3.13**\n",
+    "- **[uv](https://github.com/astral-sh/uv)** package manager\n",
+    "- **LLM API key** – For your chosen provider (required for LLM inference):\n",
+    "    - **NVIDIA API Key** from [NVIDIA AI](https://build.nvidia.com/) (for NIM models)\n",
+    "Optional requirements:\n",
+    "- **Web search API key**:\n",
+    "  - **Tavily** – [tavily.com](https://tavily.com) (for general web search)\n",
+    "- **Node.js 22+ and npm** (optional, for web UI mode)"
+   ]
   },
   {
    "cell_type": "markdown",
--- a/frontends/aiq_api/src/aiq_api/websocket_reconnect.py
+++ b/frontends/aiq_api/src/aiq_api/websocket_reconnect.py
@@ -51,11 +51,12 @@ logger = logging.getLogger(__name__)
 
 
 class WebSocketSessionRegistry:
-    """Keep track of active sockets and pending HITL responses."""
+    """Keep track of active sockets, pending HITL responses, and running workflow tasks."""
 
     def __init__(self) -> None:
         self._sockets: dict[str, WebSocket] = {}
         self._pending_interactions: dict[str, asyncio.Future[TextContent]] = {}
+        self._workflow_tasks: dict[str, asyncio.Task] = {}
         self._lock = asyncio.Lock()
 
     async def set_socket(self, conversation_id: str | None, socket: WebSocket) -> None:
@@ -123,6 +124,27 @@ class WebSocketSessionRegistry:
         async with self._lock:
             self._pending_interactions.pop(conversation_id, None)
 
+    async def set_workflow_task(self, conversation_id: str | None, task: asyncio.Task) -> None:
+        """Register the running workflow task, cancelling any stale one first."""
+        if not conversation_id:
+            return
+        async with self._lock:
+            old_task = self._workflow_tasks.get(conversation_id)
+            if old_task is not None and not old_task.done():
+                old_task.cancel()
+                logger.info("Cancelled stale workflow task for conversation %s", conversation_id)
+            self._workflow_tasks[conversation_id] = task
+
+    async def cancel_workflow_task(self, conversation_id: str | None) -> None:
+        """Cancel and remove the workflow task for a conversation."""
+        if not conversation_id:
+            return
+        async with self._lock:
+            task = self._workflow_tasks.pop(conversation_id, None)
+            if task is not None and not task.done():
+                task.cancel()
+                logger.info("Cancelled workflow task for conversation %s", conversation_id)
+
 
 _registry = WebSocketSessionRegistry()
 _installed = False
@@ -166,14 +188,32 @@ class ReconnectableWebSocketMessageHandler(WebSocketMessageHandler):
                             )
             except (asyncio.CancelledError, WebSocketDisconnect):
                 await _registry.clear_socket(self._conversation_id, self._socket)
+                await _registry.cancel_workflow_task(self._conversation_id)
+                self._cancel_running_workflow()
                 break
             except ValidationError as exc:
                 logger.warning("Invalid websocket message payload: %s", str(exc))
+
+    def _cancel_running_workflow(self) -> None:
+        """Cancel the background workflow task spawned by NAT's create_task."""
+        task = self._running_workflow_task
+        if task is not None and not task.done():
+            task.cancel()
+            logger.info(
+                "Cancelled in-flight workflow task for conversation %s",
+                self._conversation_id,
+            )
 
     async def process_workflow_request(self, user_message_as_validated_type: WebSocketUserMessage) -> None:
         """Process user messages and register sockets for reconnect."""
         await _registry.set_socket(user_message_as_validated_type.conversation_id, self._socket)
         await super().process_workflow_request(user_message_as_validated_type)
+        # NAT's parent creates the task via asyncio.create_task and stores it
+        # in self._running_workflow_task. Track it in the registry so a
+        # reconnected handler can cancel a stale workflow for this conversation.
+        task = self._running_workflow_task
+        if task is not None and not task.done():
+            await _registry.set_workflow_task(user_message_as_validated_type.conversation_id, task)
 
     async def create_websocket_message(
         self,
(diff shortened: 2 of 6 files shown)


docs/notebooks/1_Deep_Researcher_Web_Search.ipynb
```diff
@@ -60,7 +60,19 @@
...
+   "source": [
+    "<a id=\"prereqs\"></a>\n",
+    "## Prerequisites\n",
+    "\n",
+    "- **Python 3.11–3.13**\n",
+    "- **[uv](https://github.com/astral-sh/uv)** package manager\n",
+    "- **LLM API key** – For your chosen provider (required for LLM inference):\n",
+    "    - **NVIDIA API Key** from [NVIDIA AI](https://build.nvidia.com/) (for NIM models)\n",
+    "Optional requirements:\n",
+    "- **Web search API key**:\n",
```
AjayThorve, line 72: was the notebook change intentional?

AjayThorve:
@exactlyallan On client disconnect mid-workflow for shallow, the workflow still finishes in the backend. The stale messages no longer arrive, but the worklfow should ideally also close out as per the changes, right?

Claude's root cause analysis:

```ai
Backend workflow task cancellation is a no-op -- self._running_workflow_task is always None at runtime, so the entire registry-based cancellation mechanism never executes.

The NAT parent class (message_handler.py:190-196) assigns the attribute via method chaining:


self._running_workflow_task = asyncio.create_task(...).add_done_callback(_done_callback)
Task.add_done_callback() returns None, so self._running_workflow_task is immediately None. As a result:

process_workflow_request reads None, so set_workflow_task() is never called -- the registry never holds a task
On disconnect, cancel_workflow_task() is a no-op (empty registry) and _cancel_running_workflow() is a no-op (None attribute)
The test test_handler_run_cancels_workflow_on_disconnect passes only because it manually pre-seeds both handler._running_workflow_task and the registry, bypassing the broken production path
The frontend guards (isStaleMessage + isStreaming) still provide real protection, so the bug is mitigated client-side. But the backend cancellation -- which is the primary mechanism described in the PR -- does nothing until the NAT upstream bug is fixed.

Affected lines: [websocket_reconnect.py:207-216](vscode-webview://1d3a64quisdqf9u4n82mi90ntr849vr4gpfdq1hq46p1i1lduvf4/frontends/aiq_api/src/aiq_api/websocket_reconnect.py#L207-L216) (task read + registry call), [websocket_reconnect.py:197-205](vscode-webview://1d3a64quisdqf9u4n82mi90ntr849vr4gpfdq1hq46p1i1lduvf4/frontends/aiq_api/src/aiq_api/websocket_reconnect.py#L197-L205) (_cancel_running_workflow), [test_websocket_reconnect.py:543-565](vscode-webview://1d3a64quisdqf9u4n82mi90ntr849vr4gpfdq1hq46p1i1lduvf4/tests/aiq_agent/async_api/test_websocket_reconnect.py#L543-L565) (test that masks the issue)

NAT source: .venv/lib/python3.13/site-packages/nat/front_ends/fastapi/message_handler.py:190-196
```

Seems like this really is a NAT issue

AjayThorve:
this would need https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1744 to be merged

docs/notebooks/1_Deep_Researcher_Web_Search.ipynb
exactlyallan: It happens in pre-commit checks to clear notebook output - ends up always editing notebook.

frontends/aiq_api/src/aiq_api/websocket_reconnect.py
```diff
@@ -166,14 +188,32 @@ async def run(self) -> None:
...
     async def process_workflow_request(self, user_message_as_validated_type: WebSocketUserMessage) -> None:
         """Process user messages and register sockets for reconnect."""
         await _registry.set_socket(user_message_as_validated_type.conversation_id, self._socket)
         await super().process_workflow_request(user_message_as_validated_type)
+        # NAT's parent creates the task via asyncio.create_task and stores it
+        # in self._running_workflow_task. Track it in the registry so a
+        # reconnected handler can cancel a stale workflow for this conversation.
+        task = self._running_workflow_task
+        if task is not None and not task.done():
+            await _registry.set_workflow_task(user_message_as_validated_type.conversation_id, task)
```
exactlyallan, line 217: **Backend workflow task cancellation is a no-op at runtime**

`self._running_workflow_task` is always `None` in production, so the entire registry-based cancellation mechanism never executes.

The NAT parent class (`message_handler.py:190-196`) assigns the attribute via method chaining:

```python
self._running_workflow_task = asyncio.create_task(...).add_done_callback(_done_callback)
```

`Task.add_done_callback()` returns `None`, so `self._running_workflow_task` is immediately `None`. As a result:

1. `process_workflow_request` reads `None`, so `set_workflow_task()` is never called — the registry never holds a task.
2. On disconnect, `cancel_workflow_task()` is a no-op (empty registry) and `_cancel_running_workflow()` is a no-op (`None` attribute).
3. The test `test_handler_run_cancels_workflow_on_disconnect` passes only because it manually pre-seeds both `handler._running_workflow_task` and the registry, bypassing the broken production path.

The frontend guards (`isStaleMessage` + `isStreaming`) still provide real protection, so the bug is mitigated client-side. But the backend cancellation — which is the primary mechanism described in the PR — does nothing until the NAT upstream bug is fixed.

**Affected lines:**
- `websocket_reconnect.py:207-216` — task read + registry call
- `websocket_reconnect.py:197-205` — `_cancel_running_workflow`
- `test_websocket_reconnect.py:543-565` — test that masks the issue

**NAT source:** `.venv/lib/python3.13/site-packages/nat/front_ends/fastapi/message_handler.py:190-196`

This is an upstream NAT issue — see [NeMo-Agent-Toolkit#1744](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1744).

AjayThorve (approved):
Works partially, backend issue is dependent on NAT upstream feature, which would be resolved later. Thanks for adding the TODO

PR #116, opened 2026-03-03
Structured metadata

KyleZheng1284:
- Tools (Tavily, Serper, knowledge layer) now embed typed url/title/citation_key data as a JSON comment in their string output. The middleware reads it directly instead of regex-parsing. Falls back to regex for tools without metadata

- Metadata stripped before LLM -- The JSON comment is removed from the content before the LLM sees it, so agent behavior is unchanged.

- On top of Ajays current citation branch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,7 +140,6 @@ jobs:
             ./frontends/benchmarks/intent_classifier \
             ./sources/tavily_web_search \
             ./sources/google_scholar_paper_search \
-            ./sources/eci
           do
             if [ -d "$pkg" ]; then
               uv pip install -e "$pkg"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,7 +139,7 @@ jobs:
             ./frontends/benchmarks/freshqa \
             ./frontends/benchmarks/intent_classifier \
             ./sources/tavily_web_search \
-            ./sources/google_scholar_paper_search \
+            ./sources/google_scholar_paper_search
           do
             if [ -d "$pkg" ]; then
               uv pip install -e "$pkg"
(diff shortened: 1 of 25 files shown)


AjayThorve:
Thanks for the effort on this @KyleZheng1284 ! I reviewed the structured metadata approach and while the idea of explicit source encoding is clean, I have concerns about the maintenance burden:

- Every future tool must remember to call `encode_source_metadata()` — if they don't, it silently falls back to regex, so we end up maintaining both codepaths anyway
- The current `extract_sources_from_tool_result()` → `verify_citations() → sanitize_report()` pipeline already handles the cases this targets (URL corruption, hallucination, body URL leakage) without requiring any tool-side changes
- New tools automatically get source extraction for free with the regex approach — zero opt-in needed

The tests you wrote for `SourceRegistryMiddleware` are solid though — happy to pull those in separately if you want to open a smaller PR with just the test coverage additions.

AjayThorve:
> Thanks for the effort on this @KyleZheng1284 ! I reviewed the structured metadata approach and while the idea of explicit source encoding is clean, I have concerns about the maintenance burden:
> 
> * Every future tool must remember to call `encode_source_metadata()` — if they don't, it silently falls back to regex, so we end up maintaining both codepaths anyway
> * The current `extract_sources_from_tool_result()` → `verify_citations() → sanitize_report()` pipeline already handles the cases this targets (URL corruption, hallucination, body URL leakage) without requiring any tool-side changes
> * New tools automatically get source extraction for free with the regex approach — zero opt-in needed
> 
> The tests you wrote for `SourceRegistryMiddleware` are solid though — happy to pull those in separately if you want to open a smaller PR with just the test coverage additions.

@KyleZheng1284 i pulled some tests from SourceRegistryMiddleware and push those to my PR, thanks for this!

PR #117, opened 2026-03-04
feat: add more example configs

cdgamarose-nv:


AjayThorve (approved):
Verified all configs work as expected

PR #119, opened 2026-03-04
Bug fixes aiq-agent

AjayThorve:
This PR fixes a couple of issues:

- PostgreSQL checkpointer enforces NOT NULL on thread_id in checkpoint_blobs, but requests without a conversation-id header leave it as None. SQLite is permissive so this only fails in Docker deployments using Postgres.

- NAT provides `periodic_cleanup` as a Dask task for job_info expiry but it was never submitted. The job_events table (managed by our EventStore) had cleanup methods defined but no caller.


KyleZheng1284 (approved):
LGTM

cdgamarose-nv (approved):
Looks good!

PR #127, opened 2026-03-05
Fix for reduced-motion breaking panel show/hide bug

exactlyallan:
## Summary

- **Fix SessionsPanel (and all KUI SidePanels) broken under `prefers-reduced-motion: reduce`**: KUI's slide animations are gated behind `@media (prefers-reduced-motion: no-preference)` with `animation-fill-mode: both` persisting the final transform. When reduced motion is enabled, animations are skipped entirely and force-mounted panels stay visible at `transform: none`. Added CSS fallback rules in `globals.css` that directly apply the correct end-state transforms (`translate(-100%)` / `translate(100%)`) under reduced motion.
- **Respect reduced motion for inline JS transitions**: Created a `useReducedMotion()` hook and applied it to `ResearchPanel` and `MainLayout` so their inline `style` transitions become instant when the user prefers reduced motion. The hook initializes as `false` to avoid SSR hydration mismatches.
- **Add min-width 768px viewport guard**: Instead of a full mobile-responsive layout, enforce a minimum width on `html/body` and the root layout container so the browser shows a horizontal scrollbar when the viewport is narrower than 768px, preventing nav/UI from being cropped.

## Test plan

- [ ] Enable "Reduce Motion" in OS accessibility settings and verify:
  - SessionsPanel opens and closes correctly (no stuck panel)
  - DataSourcesPanel and SettingsPanel open/close correctly
  - ResearchPanel expands/collapses instantly (no animation)
  - Collapsible/accordion sections expand and collapse
- [ ] Disable "Reduce Motion" and verify all panel animations still work normally
- [ ] Resize browser window below 768px and verify horizontal scrollbar appears instead of layout breakage
- [ ] Verify no hydration mismatch warnings in the browser console

Made with [Cursor](https://cursor.com)

### DIFF
--- a/frontends/ui/src/app/globals.css
+++ b/frontends/ui/src/app/globals.css
@@ -64,6 +64,7 @@
 html,
 body {
   height: 100%;
+  min-width: 768px;
   margin: 0;
   padding: 0;
 }
--- a/frontends/ui/src/app/globals.css
+++ b/frontends/ui/src/app/globals.css
@@ -72,9 +72,49 @@ body {
   height: 100%;
 }
 
-/* Smooth scrolling */
-html {
-  scroll-behavior: smooth;
+/* Smooth scrolling — only when user has no motion preference */
+@media (prefers-reduced-motion: no-preference) {
+  html {
+    scroll-behavior: smooth;
+  }
+}
+
+/* =============================================================================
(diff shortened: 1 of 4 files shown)


exactlyallan:
@AjayThorve Note this will address VDR bug.

AjayThorve (approved):
lgtm

PR #140, opened 2026-03-12
Add MathJax support for report rendering (laTex equations)

exactlyallan:
## Summary

This PR adds MathJax-based math rendering for the UI and exported PDFs while keeping Markdown as the source-of-truth format.

### What changed
- Enabled math rendering in agent response bubbles
- Enabled math rendering in the final report view
- Extended the PDF export pipeline so delimited math renders as equations instead of raw Markdown delimiters
- Added shared math normalization/parsing helpers so the UI renderer and PDF export follow the same delimiter rules

### Supported math delimiters
- Inline: `$...$` and `\(...\)`
- Display: `$$...$$` and `\[...\]`

### Important behavior
- Bare, unwrapped equations are intentionally left as plain text
- Markdown export remains unchanged and preserves the original math markup

### Direct dependencies added
| Package | Version | Why it was added | License |
|---|---:|---|---|
| `mathjax-full` | `^3.2.1` | Core MathJax engine used to render equations, including server-side SVG generation for PDF export | `Apache-2.0` |
| `rehype-mathjax` | `^7.1.0` | Rehype plugin used to render math inside the React markdown pipeline | `MIT` |
| `remark-math-extended` | `^6.1.0` | Markdown parser plugin used to support both dollar delimiters and `\(...\)` / `\[...\]` delimiters | `MIT` |
| `svg-parser` | `^2.0.4` | Parses MathJax SVG output so it can be embedded into the existing `react-pdf` export path | `MIT` |

### Existing direct dependencies changed
- No existing direct dependency versions were upgraded or replaced
- The dependency delta is limited to the four new packages above

### Direct dependency license profile
- `Apache-2.0`: `mathjax-full`
- `MIT`: `rehype-mathjax`, `remark-math-extended`, `svg-parser`

### Practical takeaway
- No copyleft licenses were introduced in the direct dependency set for this change
- The added dependency footprint is small and uses permissive licenses only

### UI rendering
- The shared markdown renderer now supports math rendering when explicitly enabled
- Math rendering was turned on only for:
  - agent responses
  - final reports

### PDF export
- The existing PDF pipeline was preserved
- Math expressions are converted to MathJax SVG and then rendered through the current `react-pdf` flow
- This avoids introducing a separate HTML-to-PDF or TeX toolchain

## Validation
- `npm run type-check` passed
- `npm run lint` passed, with only pre-existing warnings elsewhere in the repo


exactlyallan:
Closing as stale - will try to reimplement as part of skills.

PR #149, opened 2026-03-13
Enhance title extraction logic in citation verification to prioritize titles closest to URLs

AjayThorve:
- Enhance title extraction logic in citation verification to prioritize titles closest to URLs
- Add tests to ensure correct title assignment for multiple URLs within the same text block and preference for preceding titles.


cdgamarose-nv (approved):
lgtm

PR #152, opened 2026-03-16
update super endpoints to build.nvidia

AjayThorve:


raykallen (approved):
lgtm. we'll patch update for super when ready.

PR #153, opened 2026-03-16
Improve loading spinner response 

exactlyallan:
**Summary**

- Fix research panel loading states so users see immediate feedback when a new deep research report starts (spinner)
- Backfill the final report on deep research completion when live report content has not yet reached the store, so an already-open report tab updates automatically.
- Load archived report content when users open the research panel directly to the Report tab for an already completed job, without requiring the View report button.


exactlyallan:
Stale - closing.

PR #162, opened 2026-03-31
added togle feature, and fix text

DinaLaptii:


exactlyallan:
Possibly useful for internal A/B testing, but not necessary for public blueprint. Closing.

PR #163, opened 2026-03-31
fix bug in uploading a file, and delete unnecessary func

DinaLaptii:
If a user uploads a file, deletes it, and then re-uploads the same file, it may not appear in the UI.
I also deleted a function that was only being used by tests.


frontends/ui/src/features/documents/store.ts
```diff
@@ -53,9 +53,10 @@ export const useDocumentsStore = create<DocumentsStore>()(
 
       addTrackedFile: (file) => {
         set(
-          (state) => ({
-            trackedFiles: [...state.trackedFiles, file],
-          }),
+          (state) => {
+            const next = [...state.trackedFiles, file]
+            return { trackedFiles: next }
+          },
```
AjayThorve, line 59: isn't this identical behavior? is this just updated for readability?

AjayThorve (approved):
LGTM

PR #165, opened 2026-03-31
default state of data sources side panel is open

DinaLaptii:
I need to add the feature toggle logic for the new UI from another pr


frontends/ui/src/features/layout/store.ts
```diff
@@ -119,15 +120,15 @@ export const useLayoutStore = create<LayoutStore>()(
         }
       },
 
-      disableNonWebSources: () =>
+      disableEciSources: () =>
```
AjayThorve: This was intentional, ECI terminology is internal only, and not generic for public blueprint

exactlyallan:
TBH I would wait on this PR since we are working on a more holistic design to fix this issue.

cdgamarose-nv (approved):

PR #169, opened 2026-04-02
Add register_token_fetcher plugin hook for auth extensibility

AjayThorve:
## Summary

- Add `register_token_fetcher(fetcher, priority)` to `aiq_agent.auth` — allows external plugins to register additional token sources without monkey-patching
- Registered fetchers are tried in priority order (highest first) before the default AIQContext cookie lookup
- `get_current_user_info()` benefits automatically since it delegates to `get_auth_token()`
- Add `clear_token_fetchers()` for test isolation

## Motivation

The current `get_auth_token()` only reads from AIQContext cookies set by the Next.js frontend. Enterprise deployments using custom SSO providers (Okta, Auth0, Azure AD, custom OIDC), CLI-based workflows, or API gateway token injection have no clean way to provide tokens — they must monkey-patch `get_auth_token()` at import time.

With this hook, auth extensions become:
```python
from aiq_agent.auth import register_token_fetcher

def my_sso_token_fetcher() -> str | None:
    # Check env var, cached token, custom header, etc.
    return os.getenv("MY_SSO_TOKEN")

register_token_fetcher(my_sso_token_fetcher, priority=10)
```

The fetcher chain is:
1. Registered fetchers (highest priority first)
2. Default AIQContext cookie lookup (existing behavior, unchanged)

First non-None token wins. Exceptions in fetchers are caught and logged, falling through to the next source.

## Test plan

- [x] `test_get_auth_token_default_returns_none` — no context, no fetchers → None
- [x] `test_register_token_fetcher_basic` — registered fetcher returns token
- [x] `test_register_token_fetcher_priority` — highest priority wins
- [x] `test_register_token_fetcher_skips_none` — None return falls through
- [x] `test_register_token_fetcher_skips_exceptions` — exceptions fall through
- [x] `test_clear_token_fetchers` — clear restores default behavior
- [x] `test_get_current_user_info_uses_registered_fetcher` — end-to-end with JWT

### DIFF
--- a/src/aiq_agent/auth/utils.py
+++ b/src/aiq_agent/auth/utils.py
@@ -22,10 +22,34 @@ Token source: Context cookies (idToken) - set by the frontend auth layer.
 import base64
 import json
 import logging
+from collections.abc import Callable
 
 from pydantic import BaseModel
 
 logger = logging.getLogger(__name__)
+
+_token_fetchers: list[tuple[int, Callable[[], str | None]]] = []
+
+
+def register_token_fetcher(fetcher: Callable[[], str | None], priority: int = 0) -> None:
+    """Register an additional token source.
+
+    Registered fetchers are tried in priority order (highest first) before
+    the default Context cookie lookup. The first fetcher that returns a
+    non-None token wins.
+
+    Args:
+        fetcher: Callable that returns a token string or None.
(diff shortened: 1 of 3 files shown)


cdgamarose-nv (approved):
nice, this helps

PR #171, opened 2026-04-05
deleted settings panel, move logical to use popover

DinaLaptii:
I think we need a feature toggle for these changes. I have updated the design according to the Figma:
https://www.figma.com/design/JHtiVv7hEmntIYISMcoQJP/AI-Q-3.0--Nvidia-Agent-Space-?node-id=139-7444&t=UdqFIqYeUipeSgRe-0


tanleach (approved):
Discussed offline, looks reasonable.

PR #172, opened 2026-04-06
feat: cost analysis tool with pricing configs, one-time report generation tools, and profiling config example

cdgamarose-nv:
This cost/profiling tool takes an NAT eval trace and generates a single HTML report. It  breaks down cost ($), tokens, and latency by model, workflow phase (orchestrator vs planner vs researcher), tools, and per-query behavior. 

The final charts are generated for cost, latency, cache use, and token patterns.

The `nat_adapter.py` is needed because NAT gives a raw event trace; it does not tell you “how much did the planner vs researchers cost?”. So this tool infers phase from task tool spans and subagent_type, applies pricing, and turns that into actionable rollups.


AjayThorve (approved):
This is great!

PR #179, opened 2026-04-13
fix: update allowed paths

cdgamarose-nv:


AjayThorve:
The middleware additions look right.
                                                                                                                                                                                                    
One thought — I think we can skip the `/v1/docs` remapping in plugin.py and just keep the docs at FastAPI's default `/docs`. Doc endpoints are API metadata rather than versioned business logic, so `/v1/` doesn't quite fit here. The remapping also adds some extra moving parts (stripping default routes, re-calling app.setup()) and creates a coupling between plugin.py and the middleware allowlist that could be easy to break later.                    
                                                                                                                                                                                                    
What do you think about reverting the plugin.py changes and just adding `/docs, /redoc, /openapi.json` to `EXTERNAL_ALLOWED_PATHS` and `AUTH_EXEMPT_PATHS`? That should be all we need.

cdgamarose-nv:
reverted `plugin.py`

frontends/aiq_api/src/aiq_api/auth/middleware.py
```diff
@@ -99,6 +99,11 @@ def get_current_user() -> dict[str, Any]:
     "/docs",
     "/redoc",
     "/openapi.json",
+    "/v1/docs",
+    "/v1/redoc",
+    "/v1/openapi.json",
```
AjayThorve, line 104: won't these be dead routes now that we reverted plugin.py changes

frontends/aiq_api/src/aiq_api/auth/middleware.py
```diff
@@ -109,7 +114,15 @@ def get_current_user() -> dict[str, Any]:
...
 # External paths that require no token (monitoring, etc.)
-AUTH_EXEMPT_PATHS: set[str] = {"/health", "/docs", "/redoc", "/openapi.json"}
+AUTH_EXEMPT_PATHS: set[str] = {
+    "/health",
+    "/docs",
+    "/redoc",
+    "/openapi.json",
+    "/v1/docs",
+    "/v1/redoc",
+    "/v1/openapi.json",
```
AjayThorve, line 124: same as above

AjayThorve:
Previous PR actually fixed the issue, this is not needed

PR #181, opened 2026-04-14
feat: add exa_web_search data source

maxwbuckley:
## Summary
- Adds `sources/exa_web_search`, a new NAT data source wrapping `langchain-exa`'s `ExaSearchResults`, mirroring the existing `tavily_web_search` package (config, stub-on-missing-key, retries, content truncation, XML-tagged output).
- Exposes `auto` / `deep` / `fast` `search_type` options (requires `langchain-exa>=1.1.0`). `search_type` is typed as `Literal` so invalid values fail at config parse time.
- Wires the new plugin into the workspace, `deploy/Dockerfile`, and `scripts/setup.sh` so it installs in dev, Docker, and container builds.
- Adds unit tests covering config, stub path, result formatting, query & content truncation, retries, empty results, and 401 handling (13 tests, all passing).
- Documentation updates: config reference (with `search_type` explanations), extending guides, installation, quick-start, deployment (docker-build, docker-compose, kubernetes), FAQ, and troubleshooting.

## Justification
The rationale here is similar to the NeMo Agent Toolkit request and contribution for an Exa source:
- Issue: https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues/1848
- PR: https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1846

AI-Q currently only ships a Tavily-backed web search source; adding Exa gives users a first-class alternative search provider with the same ergonomics and config surface.

Context: I am speaking to several NVIDIA folks tonight about further Exa contributions — happy to coordinate with Corey Nolet, Manas Singh, and Nathan Stephens.

## Known issue (inherited from `tavily_web_search`)
`max_retries` is not bounded at the `Field` level. If a user sets `max_retries: 0` in config, the retry loop in `_exa_web_search` is skipped and the function returns `None` implicitly. This PR intentionally mirrors the existing `tavily_web_search` behavior for consistency rather than diverging — the same bug exists in `sources/tavily_web_search/src/register.py`. Suggested follow-up: add `ge=1` (or similar) to both sources' `max_retries` `Field` and ensure a concrete fallback return after the loop. Happy to do this as a separate PR covering both sources.

## Test plan
- [x] `uv run pytest sources/exa_web_search/tests` — 13 passed
- [ ] Build Docker image and confirm `_type: exa_web_search` resolves
- [ ] End-to-end run with a real `EXA_API_KEY` against a config using `exa_web_search`


docs/source/customization/configuration-reference.md
```diff
@@ -153,6 +153,39 @@ functions:
...
+```yaml
+functions:
+  web_search_tool:
+    _type: exa_web_search
+    max_results: 5
+    max_content_length: 1000
+
+  deep_web_search_tool:
+    _type: exa_web_search
+    max_results: 2
```
shyamexa: We should use max_results = 5, not 2. Quality will be much higher

docs/source/customization/configuration-reference.md
```diff
@@ -153,6 +153,39 @@ functions:
...
+### `exa_web_search`
+
+Web search powered by the [Exa API](https://exa.ai/) via `langchain-exa`.
+
+```yaml
+functions:
+  web_search_tool:
+    _type: exa_web_search
+    max_results: 5
+    max_content_length: 1000
```
shyamexa: should make it clear that max content length only applies for text = true; and even then, 1000 is too low, i would make it 10,000

sources/exa_web_search/src/register.py
```diff
@@ -0,0 +1,196 @@
...
+_missing_key_warned = False
+
+
+class ExaWebSearchToolConfig(FunctionBaseConfig, name="exa_web_search"):
+    """
+    Tool that retrieves relevant contexts from web search (using Exa) for the given question.
+    Requires an EXA_API_KEY environment variable or api_key config.
+    """
+
+    max_results: int = Field(default=3, description="Maximum number of search results to return")
```
shyamexa: i'd say we ignore tavily's implementation; lets use exa's preferred implementation which is 5 or 10 results. let's go with 5.

sources/exa_web_search/src/register.py
```diff
@@ -0,0 +1,196 @@
...
+    )
+    highlights: bool = Field(
+        default=True,
+        description=(
+            "Whether to return highlighted snippets for each result. Highlights are token-efficient and "
+            "enabled by default; they are used as the result body when `full_text` is False."
+        ),
+    )
+    max_content_length: int | None = Field(
+        default=None,
```
louiswalsh: I think the default here should also be 10k in case full_text is passed in as true and max_content_length is not passed in, it would potentially be `none`

sources/exa_web_search/src/register.py
maxwbuckley: Done in d30007d — default bumped to 5.

docs/source/customization/configuration-reference.md
maxwbuckley: Done in d30007d — example and defaults table both use 5 now.

docs/source/customization/configuration-reference.md
maxwbuckley: Done in d30007d — the example now sets `full_text: true` alongside `max_content_length: 10000` so it's obvious the cap only applies in that mode. Default bumped to 10000 too (see reply on louiswalsh's thread).

sources/exa_web_search/src/register.py
maxwbuckley: Done in d30007d — default is now 10000. Users can still pass `None` explicitly to disable truncation.

maxwbuckley:
Addressed in d30007d:
- @shyamexa: `max_results` default bumped to 5 in both code and docs; deep example also uses 5.
- @shyamexa: example in the reference docs now pairs `full_text: true` with `max_content_length: 10000` to make it clear the cap only applies when full text is on.
- @louiswalsh: `max_content_length` default is now `10000` (was `None`); passing `None` still opts out of truncation.
- Greptile P2 cleanup: dropped `httpx` from `sources/exa_web_search/pyproject.toml` — it was only a transitive dep via `langchain-exa`.

Also verified end-to-end against the live Exa API with a standalone script (config defaults, stub path, `max_retries=0` guard, default search, `full_text` + `max_content_length` truncation, highlights-only, `search_type=fast`, long-query truncation, 401 handler) — 10/10 checks pass.

tanleach:
Changes look good. Looking into the langchain-core version bump to make sure we dont have any unintended impact in other areas.

tanleach:
@maxwbuckley please take a look at pipeline and update: https://github.com/NVIDIA-AI-Blueprints/aiq/actions/runs/24554615672/job/71834903034?pr=181#step:8:37

pastorsj (approved):
This looks great to me. Let's fix the CI, but otherwise, this is great. Thanks for the contribution here!

pastorsj:
Worth noting that if we add more support for more search engines like Tavily and Exa, it will be worth refactoring to clean up duplicate code across these tools.

maxwbuckley:
@tanleach pushed 4dc73ff — was a ruff I001 (import order) on `sources/exa_web_search/tests/test_register.py`; `exa_web_search` isn't classified as first-party so the local-import block needed merging into the third-party group. Run is queued in `action_required` state — could you approve it when you get a moment? Thanks!

tanleach:
@maxwbuckley please check pipeline

sources/exa_web_search/src/register.py
```diff
@@ -0,0 +1,150 @@
...
+        description="Exa search type: 'auto', 'deep', or 'fast'",
+    )
+    include_text: bool = Field(default=True, description="Whether to include full text contents in results")
+    max_content_length: int | None = Field(
+        default=None,
+        description="Max characters per result content. If set, truncates each result to reduce token usage.",
+    )
+
+
+@register_function(config_type=ExaWebSearchToolConfig)
```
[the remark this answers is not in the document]
tanleach: resolving as it is outdated

sources/exa_web_search/src/register.py
```diff
@@ -0,0 +1,150 @@
...
+                            "Error: Web search failed due to invalid API key (401 Unauthorized).\n"
+                            "Please check your EXA_API_KEY and ensure it is valid.\n"
+                        )
+                    return f"Error: Web search failed - {error_msg}"
+                await asyncio.sleep(2**attempt)
+
+    yield FunctionInfo.from_fn(
+        _exa_web_search,
+        description=_exa_web_search.__doc__,
+    )
```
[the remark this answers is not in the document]
tanleach: @maxwbuckley please take a look at greptile feedback, feel free to resolve if not needed

sources/exa_web_search/src/register.py
```diff
@@ -0,0 +1,150 @@
...
+
+        exa_search = ExaSearchResults()
+
+        def _truncate_content(content: str) -> str:
+            if tool_config.max_content_length and len(content) > tool_config.max_content_length:
+                return content[: tool_config.max_content_length - 3] + "..."
+            return content
+
+        for attempt in range(tool_config.max_retries):
+            try:
```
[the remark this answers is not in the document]
tanleach: @maxwbuckley please take a look at greptile feedback, feel free to resolve if not needed

sources/exa_web_search/src/register.py
```diff
@@ -0,0 +1,150 @@
...
+        )
+        return
+
+    async def _exa_web_search(question: str) -> str:
+        """Retrieves relevant contexts from web search (using Exa) for the given question.
+
+        Args:
+            question (str): The question to be answered. Will be truncated to 400 characters if longer.
+
+        Returns:
```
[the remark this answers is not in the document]
tanleach: @maxwbuckley please take a look at greptile feedback, feel free to resolve if not needed

tanleach (approved):
Looks good and follows tavilly implementation. Agreed that we should consider refactorign in the future if more search tools need to be implemented.

tanleach (approved):

PR #184, opened 2026-04-14
fix: surface unavailable tool details in user-facing error messages

KyleZheng1284:
## Summary
- When research fails with no sources captured, the error handler now reports which tools were unavailable (stubs) instead of showing a generic "temporary issue" message
- Enriches `EmptySourceRegistryError` with an optional `unavailable_tools` list populated via the existing `validate_tool_availability` check
- Preserves the generic fallback when no tool details are available
- Should help resolve the internal version to outline unset api keys in response since ECI might've been the cause 
## Test plan
- Start CLI with `TAVILY_API_KEY` unset
- Submit a query and confirm the error lists the specific unavailable tools
- Set valid keys and confirm normal research works

### DIFF
--- a/src/aiq_agent/agents/chat_researcher/agent.py
+++ b/src/aiq_agent/agents/chat_researcher/agent.py
@@ -205,11 +205,9 @@ class ChatResearcherAgent:
             except EmptySourceRegistryError as exc:
                 logger.warning("Shallow research produced no verifiable sources")
                 if exc.unavailable_tools:
-                    err_msg = (
-                        "Cannot start shallow research: No tools are available. "
-                        "At least one tool must be configured and available. "
-                        f"Unavailable tools: {', '.join(exc.unavailable_tools)}."
-                    )
+                    from aiq_agent.common.tool_validation import format_tool_unavailability_error
+
+                    err_msg = format_tool_unavailability_error("shallow research", exc.unavailable_tools)
                 else:
                     err_msg = (
                         "The search tools did not return any results for this question. "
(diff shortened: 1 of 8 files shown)


KyleZheng1284:
@greptileai

AjayThorve:
Nice improvement over the generic error message — surfacing tool details is much better for debugging.

  A couple of concerns:

  1. Security: tool/API key details in user-facing responses

  The formatted error message (e.g. "tavily_search is unavailable") gets returned as an AIMessage, which flows through the WebSocket directly to the UI. For internal/CLI use this is fine, but for production deployments it leaks infrastructure details to end users (which services are configured, which keys are missing).

  Suggestion: keep the detailed message in server logs only, and return a generic user-facing message like "Some search capabilities are currently unavailable. Contact your administrator." Or gate the detailed message behind a config flag.
  
  May be use [AIQ_DEV_ENV](https://github.com/NVIDIA-AI-Blueprints/aiq/blob/develop/deploy/.env.example#L12) as a check to see what mode we are in?

  2. Startup validation would be a better primary fix

  Right now the user has to submit a query, wait for research to run, and only then learns that tools were unavailable from the start. validate_tool_availability already exists — calling it at server startup (config-aware, checking only the tools the active config declares) would catch this immediately. That's probably a small follow-up.

  This PR is still useful as a runtime safety net (tools can fail mid-session), but the startup check should be the primary guard.

AjayThorve (changes requested):
requested a minor update, to keep this CLI only

KyleZheng1284:
@greptileai

AjayThorve (approved):
LGTM

PR #188, opened 2026-04-15
fix: inject fresh idToken on WebSocket upgrade for reliable auth

AjayThorve:
## Summary

- WebSocket upgrades bypass Next.js middleware (`proxy.ts`), so the `idToken` cookie may be missing or stale when the gateway proxies to the backend
- Downstream services that need the caller's identity (e.g. ECI enterprise search) fail with "No authentication token available" for some users — clearing browser cache is the only workaround
- Fix: decode the NextAuth session JWT (which always has the latest refreshed token) during WebSocket upgrade in the gateway and inject the fresh `idToken` cookie on the proxied request

## Root Cause

`proxy.ts` sets the `idToken` cookie on HTTP responses, but the gateway `server.js` handles WebSocket upgrades directly without going through Next.js. When the browser hasn't received a recent HTTP response (first load race, expired token, stale cache), the WebSocket connects without auth. Even when the cookie exists, it may be stale (from before a token refresh) while the NextAuth session JWT has the latest credential.

## What Changed

`frontends/ui/server.js`:
- Added `extractIdTokenFromSession()` — decodes the NextAuth session JWT using `next-auth/jwt` and extracts the `idToken`
- On every WebSocket upgrade to `/websocket`, the gateway now decodes the session JWT and injects the fresh `idToken` cookie, replacing any stale value
- Falls back gracefully if `NEXTAUTH_SECRET` is not set or decoding fails

## Test plan

- [ ] Sign in via Starfleet SSO, immediately send a chat query that triggers ECI search — should succeed without "No authentication token" error
- [ ] Wait for token refresh (~30 min), then send another ECI query via WebSocket — should still succeed with the refreshed token
- [ ] Clear browser cache and repeat — should work on first load
- [ ] Verify non-auth deployments (`REQUIRE_AUTH=false`) are unaffected

### DIFF
--- a/frontends/ui/server.js
+++ b/frontends/ui/server.js
@@ -24,6 +24,7 @@
 const http = require('http')
 const httpProxy = require('http-proxy')
 const { parse } = require('url')
+const { decode: decodeJwt } = require('next-auth/jwt')
 
 const dev = process.env.NODE_ENV !== 'production'
 const hostname = process.env.HOSTNAME || '0.0.0.0'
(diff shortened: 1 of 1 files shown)


exactlyallan (approved):
LGTM

PR #199, opened 2026-04-18
fix auth trust boundary and enforce async job ownership

cdgamarose-nv:
## Summary

  This PR fixes two auth/security issues in aiq:

  1. Unverified JWT payloads were being used as trusted user identity in request-backed flows.
  2. Async job endpoints lacked object-level authorization and only checked request auth + job existence.

  ## Changes

  ### 1. Trusted identity now comes only from verified middleware context

  - Added verified principal helpers in aiq_agent.auth:
      - get_current_principal()
      - get_verified_current_user()
  - Changed get_current_user_info() to read only from verified middleware context.
  - Renamed unsafe JWT helpers to make their trust boundary explicit:
      - decode_unverified_jwt_payload()
      - get_user_info_from_unverified_token()
  - Kept the raw decode behavior only as explicitly unverified helpers.
  - Updated request-backed chat_researcher flows to use the verified principal instead of deriving identity from raw token payloads.
  - Updated middleware internal caller labeling so raw internal JWT/cookie traffic is marked unverified_jwt, not a trusted auth type.

  ### 2. Async job object-level authorization

  - Added a separate AIQ-owned job_access table keyed by job_id.
  - Persist ownership at submission time using verified principal identity:
      - owner_auth_type
      - owner_subject
      - owner_email
  - Added shared authorization helpers to require:
      - job exists
      - job_access row exists
      - requesting principal matches stored owner
  - Applied ownership enforcement to async job endpoints:
      - status
      - stream
      - stream resume
      - cancel
      - state
      - report
  - Unauthorized or ownerless job access now returns 404.

  ### 3. WebSocket auth bridge for UI-backed chat (transitional)

  - Added a transitional WebSocket auth path in aiq_api so UI WebSocket chat can establish verified identity before the UI moves fully to HTTP/SSE.
  - Moved WebSocket auth to the actual socket route boundary rather than relying only on handler-level monkeypatching.
  - Reused the same validator chain and trusted principal shape as HTTP auth.

  ### 4. DB init and cleanup

  - Added idempotent job_access creation to both init SQL paths:
      - deploy/compose/init-db.sql
      - deploy/helm/helm-charts-k8s/aiq/files/init-db.sql
  - Added cleanup for stale job_access rows alongside job cleanup.
  - Ensured submit fails closed and rolls back partial job state if access metadata persistence fails.

  ### 5. Code quality / cleanup

  - Removed package-level lazy export workaround for aiq_api.jobs.
  - Switched callers to direct submodule imports.
  - Removed internal-only UI auth/Datadog changes from this repo so public aiq remains free of NVIDIA-internal UI overlays.

  ## Security impact

  This PR closes:

  - the request-identity trust-boundary issue where decoded JWT payloads could be mistaken for authenticated identity
  - the authenticated BOLA/IDOR-style gap on async job endpoints

  Trusted authorization decisions now use verified principal identity (type + sub), and async job access is enforced per object rather than per request only.

  ## Testing

  Validated in local/dev environments with:

  - auth helper behavior for verified principal resolution
  - async job ownership persistence in job_access
  - cross-principal access denial (404) for async jobs
  - idempotent schema creation for job_access
  - WebSocket/UI chat verified-principal propagation in internal testing
  - targeted compile/test checks for touched modules

  ## Notes

  - Deprecated unsafe JWT helper aliases remain in utils.py for compatibility, but they are no longer part of the public auth surface.
  - The WebSocket auth path is a transitional bridge until the UI fully migrates to authenticated HTTP/SSE.

### DIFF
--- a/frontends/aiq_api/src/aiq_api/jobs/submit.py
+++ b/frontends/aiq_api/src/aiq_api/jobs/submit.py
@@ -29,6 +29,7 @@ from aiq_agent.auth import Principal
 from aiq_agent.auth import get_current_principal
 
 from ..registry import get_agent_config
+from .access import _make_no_auth_principal
 from .access import create_job_access
 from .access import rollback_job_submission
 from .runner import run_agent_job
@@ -39,9 +40,9 @@ logger = logging.getLogger(__name__)
 def _resolve_submission_principal(owner: str) -> Principal | None:
     """Resolve the best available principal for async job ownership.
 
-    Verified middleware identity remains the preferred source. For deployments
-    with auth disabled, fall back to a compatibility principal so legacy
-    internal/anonymous flows can still submit async jobs.
+    Used only when submit_agent_job is called programmatically without an
+    explicit principal.  Verified middleware identity is preferred; falls back
+    to a compatibility principal when auth is disabled.
     """
     principal = get_current_principal()
     if principal is not None:
@@ -50,17 +51,7 @@ def _resolve_submission_principal(owner: str) -> Principal | None:
     if os.environ.get("REQUIRE_AUTH", "false").lower() == "true":
         return None
 
-    try:
-        from aiq_api.auth.middleware import get_current_user
-
-        current_user = get_current_user()
-    except Exception:
-        current_user = {}
-
-    principal_type = str(current_user.get("type") or "anonymous")
-    subject = owner if owner else principal_type
-    email = owner if "@" in owner else None
-    return Principal(type=principal_type, sub=subject, email=email)
+    return _make_no_auth_principal(owner)
 
 
 def _get_parent_trace_context() -> tuple[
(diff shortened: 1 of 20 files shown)


frontends/aiq_api/src/aiq_api/jobs/submit.py
```diff
@@ -177,35 +184,54 @@ async def submit_agent_job(
 
         auth_token = get_auth_token()
 
+    if principal is None:
+        principal = get_current_principal()
+    if principal is None:
+        raise RuntimeError("Verified current principal required for async job submission")
```
AjayThorve, line 209: This looks like a behavior regression for auth-disabled and internal deployments. `get_current_principal()` returns `None` for `anonymous/internal/unverified_jwt callers`, so this now hard-fails async submission even when `REQUIRE_AUTH=false`. That also appears to break `use_async_deep_research` in chat_researcher, which previously worked without verified auth.

AjayThorve (commented):
One issue, rest all looks good

AjayThorve (approved):
LGTM, works will w/ and w/o auth

PR #201, opened 2026-04-22
Add reranked_search tool

Jack-Yu-815:
# changes
1. added reranked_search as a new source package
    * the tool takes in sub-tools (other search tools), triggers them in parallel, and rerank based on returned content's relevance to query. 
    * this reranker tool's input schema is dynamically created based on the input schemas of its sub-tools. Therefore, it can use sub-tools with different number of parameters, so long as they return some kind of search results. Usage is quite flexible.  
    * sub-tool must utilize `air_agent.common.SOURCE_DELIMITER` to separate result strings.

2. small changes to other built-in tools in `sources/`
    * switched to using `air_agent.common.SOURCE_DELIMITER` to separate result strings.

3. updated dependency-related files and setup script. 

## Summary by CodeRabbit

* **New Features**
  * Added a reranked search layer that runs multiple search tools, then scores and reranks results.
  * Added a new workspace package with plugin registration and an editable install during setup.

* **Improvements**
  * Standardized search/knowledge/web output formatting using a shared delimiter across sources.
  * Web search answers now appear only for the supported modes (basic/advanced).

* **Documentation**
  * Added reranker documentation plus a complete CLI configuration example.

* **Chores**
  * Refreshed secret baseline metadata.


sources/reranker/src/cross_encoder.py
```diff
@@ -0,0 +1,67 @@
...
+
+    client = NVIDIARerank(model=model_name)
+
+    documents = [Document(page_content=r) for r in results]
+    reranked_docs = client.compress_documents(query=query, documents=documents)
+
+    # Map reranked documents back to our SearchResult objects by matching content.
+    content_to_idx: dict[str, int] = {}
+    for i, r in enumerate(results):
+        content_to_idx[r] = i
```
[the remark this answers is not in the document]
Jack-Yu-815: actually this mapping logic is no longer needed. Will remove this section.

sources/reranker/example_cli_config.yml
```diff
@@ -0,0 +1,170 @@
...
+  # Example reranked_search tool config. Useful when 2 or more sources are given to the agent.
+  # ========================================================================
+  reranked_search:
+    _type: reranked_search
+    llm: nemotron_nano_llm
+    cross_encoder_model: nv-rerank-qa-mistral-4b:1
+    search_tools:
+      - web_search_tool
+      - paper_search_tool
+    top_k: 5
```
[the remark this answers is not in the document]
Jack-Yu-815: greptile is so good

sources/reranker/src/register.py
```diff
@@ -0,0 +1,144 @@
...
+
+@register_function(config_type=RerankedSearchConfig)
+async def reranked_search(config: RerankedSearchConfig, builder: Builder):
+    """
+    A cross-encoder feeds the query–document pair together into a single model pass and outputs a
+    relevance score directly.  This is more accurate but slower per document.
+
+    Uses `langchain-nvidia-ai-endpoints` (`NVIDIARerank`).  Expects `NVIDIA_API_KEY` in the environment.
+    """
+    compressor = NVIDIARerank(model=config.cross_encoder_model)
```
peterychang: Is there a reason you're not passing in `top_n=config.top_k` and doing the top_k truncation manually? 

Also, `top_n` is 5 by default, so it seems like you'll only get up to 5 results regardless of the top_k?

sources/reranker/src/register.py
```diff
@@ -0,0 +1,144 @@
...
+                logger.warning("SOURCE_DELIMITER not found in tool '%s' output", name)
+            return results
+        except TimeoutError:
+            logger.warning("tool '%s' timed out after %.0fs", name, config.timeout_seconds)
+            return []
+        except Exception:
+            logger.exception("tool '%s' failed", name)
+            return []
+
+    RerankedSearchInput = create_model(
```
peterychang, line 113: Just commenting that this is gross, but probably the best way to handle the abomination of an input the reranker needs to take in..

I assume the input parameters are LLM-generated?

Thanks, I hate it.

sources/reranker/src/register.py
```diff
@@ -0,0 +1,144 @@
...
+        all_results = []
+        for results in per_tool_results:
+            all_results.extend(results)
+        logger.info(f"{len(all_results)} results found across all search tools.")
+
+        if not all_results:
+            return "No results found across any search tool."
+
+        # Rerank. Use the overall_query for reranking purposes.
+        documents = [Document(page_content=r) for r in all_results]
```
peterychang, line 146: Do we need to worry at all about things like malformed data or injection attacks?

sources/reranker/src/register.py
```diff
@@ -0,0 +1,144 @@
...
+        reranked_docs = await compressor.acompress_documents(query=input.overall_query, documents=documents)
+        ranked_contents: list[str] = [doc.page_content for doc in reranked_docs[: config.top_k]]
+
+        return f"Top {config.top_k} results ranked by relevance:\n" + SOURCE_DELIMITER.join(ranked_contents)
```
peterychang: This line appears to confuse the LLM. The string being returned is

`Top X results ranked by relevance:`

followed by what I assume is the page's source or text. The LLM's response is

```
We need to extract relevant sources for our milestones. The top result is a summary but not a specific source; it's a generic summary. However we need to cite specific sources: peer-reviewed papers, official product announcements/blog posts. The top result appears to be a generic summary maybe from some article; not sure of the source. It might be from an article but not given. It doesn't provide a URL. It just says "Top 5 results ranked by relevance:" and then a text. That might be the output of reranked_search, which includes snippet and maybe a link? It didn't show a URL for that top result. It just gave snippet text. Possibly the tool returns a formatted answer with snippet and maybe a link is omitted. But we need to capture URLs from search results. The subsequent items show Document href links: e.g., <websites returned by search>. Also maybe others.

Thus we have at least these URLs:

- <URL 1>

- <URL 2>

- <URL 3>

- <URL 4>

- Possibly also the top summary snippet, but no URL.
```

While it looks like the LLM can figure it out, it seems like returning the raw URLs instead of text would be better in this case

peterychang: Also, the response only lists out 4 extracted URLs rather than the 5 that should be there.

peterychang: It looks like this is caused by the `<Answer>` section that tavily returns

sources/reranker/src/register.py
Jack-Yu-815: yes this is the best hack for passing llm-generated param that I could come up with. Unfortunately hh

sources/reranker/src/register.py
Jack-Yu-815: that's a good catch. Default top_n should be set to something that makes more sense, even though users can configure later.

peterychang:
Once https://github.com/Jack-Yu-815/aiq-research-assistant/pull/2 in, I think this is good to merge

peterychang (approved):

PR #203, opened 2026-04-24
Add request trace classification and pseudonymous ids

AjayThorve:
## Summary
- add always-on request trace classification tags plus optional pseudonymous user and client ids in aiq_api auth middleware
- support explicit low-cardinality access channel tagging and have the public UI proxy routes send ui
- document the new env vars and observability contract and cover the middleware behavior with auth tests

## Testing
- uv run pytest frontends/aiq_api/tests/test_auth.py -k "tags_active or common_tags or missing_secret or explicit_access_channel or client_ip_mode or does_not_tag_active_span or none_mode"

### DIFF
--- a/deploy/.env.example
+++ b/deploy/.env.example
@@ -128,3 +128,10 @@ FILE_EXPIRATION_CHECK_INTERVAL_HOURS=0
 # - /app/configs/config_web_default_llamaindex.yml
 # - /app/configs/config_web_frag.yml
 # BACKEND_CONFIG=
+#
+# Optional trace enrichment for aiq_api auth middleware:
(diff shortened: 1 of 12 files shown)


cdgamarose-nv:
Do we need to add the placeholder variables to `.env.example`?

Also, nitpick but the `middleware.py` file is now very long with many helper functions. Would it make sense to move into a `utils.py`?

cdgamarose-nv:
But otherwise, looks good to me

AjayThorve:
@cdgamarose-nv good suggestions, update .env.example and refactored to add utils.py

cdgamarose-nv (approved):
LGTM

PR #206, opened 2026-04-27
Propagate AIQ request tags to NAT spans

AjayThorve:
## Summary
- propagate resolved AIQ request tags into NAT-exported spans for HTTP, WebSocket, and async job workflows
- carry request tags into async Dask jobs so child workflow spans retain caller and access-channel context
- keep public channel tagging generic: ui, api, headless, internal, anonymous, unknown
- move request trace span injection into the API auth layer and document nat.aiq.* / nat.enduser.id span attributes

## Validation
- ruff check <changed python files and tests>
- ruff format --check <changed python files and tests>
- python -m py_compile <changed python files and tests>
- uv run pytest tests/ -q (881 passed, 2 skipped)
- uv run pytest tests/aiq_agent/async_api/test_websocket_reconnect.py -q (22 passed)

## Local E2E
- verified curl /chat traces include nat.aiq.* and nat.enduser.id attributes
- verified UI websocket traces include NAT request tags after binding websocket request context

### DIFF
--- a/deploy/.env.example
+++ b/deploy/.env.example
@@ -128,10 +128,3 @@ FILE_EXPIRATION_CHECK_INTERVAL_HOURS=0
 # - /app/configs/config_web_default_llamaindex.yml
 # - /app/configs/config_web_frag.yml
 # BACKEND_CONFIG=
-#
-# Optional trace enrichment for aiq_api auth middleware:
-# AIQ_TRACE_USER_IDENTITY_MODE=none
-# AIQ_TRACE_USER_IDENTITY_HMAC_SECRET=
-# AIQ_TRACE_CLIENT_ID_MODE=none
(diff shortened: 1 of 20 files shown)


AjayThorve:
Addressed the review blockers in the latest branch update:

- Added and tested the `build_request_trace_tags` API helper used by WebSocket auth, so `websocket_reconnect.py` no longer has the import-time failure.
- Moved request trace injection into the API auth package and install it from middleware instead of an agent module.
- Added debug logging for span lookup misses during request tag injection.
- Added a WebSocket regression test covering request tag context binding.
- Kept public access-channel examples generic (`ui`, `api`, `headless`, `internal`, `anonymous`, `unknown`).

Validation: `uv run pytest tests/ -q` passes locally with 881 passed, 2 skipped.

cdgamarose-nv (approved):
Looks good!

PR #209, opened 2026-04-30
fix: keep commas inside URL paths in citation source extractor

AjayThorve:
## Summary

- Drop `,` from the `_GENERIC_URL_RE` exclusion class so URLs whose paths legitimately contain commas (e.g., FAA cam URLs with embedded lat/lon coordinates) are captured in full.
- The existing `rstrip(".,;)")` continues to handle trailing comma when it's actually sentence punctuation, so no behavior change for that case.
- `]` stays excluded so markdown link brackets still terminate the URL match.

## Repro

A `web_search_tool` result containing `<Document href="https://weathercams.faa.gov/map/-122.31167,47.22287,10/airport/SEA/details/weather">` was registered in the source registry as just `https://weathercams.faa.gov/map/-122.31167` (truncated at the first `,`). When the LLM later cited the full URL, the verifier saw `url_not_in_registry` and silently removed the citation, even though the source had been retrieved by the tool call.

## Test plan
- [x] 3 new regression tests in `TestGenericUrlExtractor`:
  - URL with commas in path (FAA-style coordinates) preserves full URL
  - Trailing comma followed by whitespace is still stripped via `rstrip`
  - Markdown link `[text](url)` still terminates at `]`
- [x] `uv run pytest tests/aiq_agent/common/test_citation_verification.py` → 103 passed
- [x] Full suite: 909 passed (3 new), 2 skipped, 1 pre-existing collection error unrelated

## Notes

This is the regex-level fix only. The longer-term improvement — moving the verifier from binary "valid or silently removed" to confidence-scored citations with a UI confidence badge — is being tracked separately for the 2.2 release.

### DIFF
--- a/src/aiq_agent/common/citation_verification.py
+++ b/src/aiq_agent/common/citation_verification.py
@@ -440,10 +440,17 @@ def extract_sources_from_tool_result(tool_name: str, content: str) -> list[Sourc
 # Generic URL extractor — works for any tool output format.
 # Commas are valid URL path characters (RFC 3986 sub-delim) and appear in real
 # URLs like https://weathercams.faa.gov/map/-122.31167,47.22287,10/...; we
-# include them in the match and rely on the trailing rstrip below to remove
-# any comma that's actually punctuation. ``]`` stays excluded because it
-# almost always terminates a markdown link rather than appearing in a path.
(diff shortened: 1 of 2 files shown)


AjayThorve:
@greptileai can you re-review

cdgamarose-nv (approved):
Interesting find, LGTM!

PR #211, opened 2026-05-03
feat: add support for skills and sandbox along with example

cdgamarose-nv:
- Job-scoped deep agents sandbox support with modal as the first provider.
- Built-in skills loading from the deep researcher skills directory.
- A quantitative data-table-analysis skill that uses sandboxed Python/pandas for normalized tables, growth rates, rankings, CSV/JSON, and markdown outputs.
- Example `config_skills.yml` showing how to enable skills and Modal sandbox execution.
 - Documentation for running AI-Q with skills/sandbox and adding new skills.

### DIFF
--- a/deploy/.env.example
+++ b/deploy/.env.example
@@ -48,11 +48,6 @@ DASK_DISTRIBUTED__LOGGING__DISTRIBUTED=warning
 # AIQ_SUMMARY_DB=
 
 # -----------------------------------------------------------------------------
-# NVIDIA API Key (required for skills sandbox)
-# -----------------------------------------------------------------------------
-NVIDIA_API_KEY=
-
-# -----------------------------------------------------------------------------
 # Environment Variables for Modal Sandbox for Skills Execution (optional)
 # -----------------------------------------------------------------------------
 # MODAL_TOKEN_ID=
--- a/deploy/.env.example
+++ b/deploy/.env.example
@@ -48,6 +48,17 @@ DASK_DISTRIBUTED__LOGGING__DISTRIBUTED=warning
 # AIQ_SUMMARY_DB=
 
 # -----------------------------------------------------------------------------
+# NVIDIA API Key (required for skills sandbox)
+# -----------------------------------------------------------------------------
+NVIDIA_API_KEY=
+
+# -----------------------------------------------------------------------------
+# Environment Variables for Modal Sandbox for Skills Execution (optional)
+# -----------------------------------------------------------------------------
+# MODAL_TOKEN_ID=
+# MODAL_TOKEN_SECRET=
+
+# -----------------------------------------------------------------------------
 # Observability / Tracing (optional)
 # See docs/source/deployment/observability.md for setup guides.
 # -----------------------------------------------------------------------------
--- a/docs/source/architecture/agents/sandbox.md
+++ b/docs/source/architecture/agents/sandbox.md
@@ -1,3 +1,8 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
 # Deep Research Sandbox Notes
 
 Deep research can optionally run DeepAgents `execute` calls in a Modal sandbox.
--- a/src/aiq_agent/agents/deep_researcher/deepagents_runtime.py
+++ b/src/aiq_agent/agents/deep_researcher/deepagents_runtime.py
@@ -1,5 +1,18 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 """DeepAgents skills and sandbox runtime support for deep research."""
 
 from __future__ import annotations
--- a/src/aiq_agent/agents/deep_researcher/deepagents_runtime.py
+++ b/src/aiq_agent/agents/deep_researcher/deepagents_runtime.py
@@ -32,6 +32,7 @@ from deepagents.backends.protocol import EditResult
 from deepagents.backends.protocol import ExecuteResponse
 from deepagents.backends.protocol import FileDownloadResponse
 from deepagents.backends.protocol import FileUploadResponse
+from deepagents.backends.protocol import ReadResult
 from deepagents.backends.protocol import WriteResult
 from deepagents.backends.sandbox import BaseSandbox
 from pydantic import BaseModel
@@ -86,6 +87,12 @@ class _PrefixedStateBackend(StateBackend):
         result = super().edit(file_path, old_string, new_string, replace_all=replace_all)
         if result.error and file_path in result.error:
             return EditResult(error=self._rewrite_error(result.error, file_path))
+        return result
+
+    def read(self, file_path: str, offset: int = 0, limit: int = 2000) -> ReadResult:
+        result = super().read(file_path, offset=offset, limit=limit)
+        if result.error and file_path in result.error:
+            return ReadResult(error=self._rewrite_error(result.error, file_path))
         return result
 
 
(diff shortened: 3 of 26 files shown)


docs/source/architecture/agents/sandbox.md
```diff
@@ -0,0 +1,42 @@
+# Deep Research Sandbox Notes
```
AjayThorve, line 6: Wondering if this goes in `docs/`, but since this feature itself is WIP, there is a fair argument in favor of current place as well. Curious to know your thoughts

src/aiq_agent/agents/deep_researcher/deepagents_runtime.py
```diff
@@ -0,0 +1,338 @@
...
+                },
+            )
+            return self._backend
+
+        self._backend = CompositeBackend(
+            default=StateBackend(),
+            routes={
+                BUILTIN_SKILL_SOURCE: StateBackend(),
+                "/shared/": StateBackend(),
+            },
```
AjayThorve, line 192: **CRITICAL — this is what causes the researcher loop in a sandbox-enabled run.**

Repro: researcher writes `/shared/0_weather_data.txt` once successfully, then on a follow-up write gets back

> `Cannot write to /0_weather_data.txt because it already exists. Read and then make an edit, or write to a new path.`

Note the **stripped path**. Why: `CompositeBackend.write` (deepagents 0.5.0, `composite.py:460`) strips the route prefix before calling `StateBackend.write`, then rewrites `result.path` back to the full path on success — but does **not** rewrite the path embedded in `result.error`. `StateBackend.write` (`state.py:254`) bakes the stripped key into the error string, which surfaces to the model.

Result: nemotron-nano spends 3+ minutes alternating `rm`/`unlink`/`stat`/`mv`/`find` on `/shared/...` (the real Modal sandbox path, empty) vs `read_file`/`glob` (which see the StateBackend file). The two views never reconcile and the researcher subagent never returns. This is the actual cause of the failed run, not the model's reasoning.

Minimal fix — wrap `StateBackend` so error messages echo the user-visible path. ~20 LOC, no upstream change required:

```python
from dataclasses import replace
from deepagents.backends.protocol import EditResult, WriteResult

class _PrefixedStateBackend(StateBackend):
    """StateBackend that re-prepends a route prefix to error messages.

    CompositeBackend.write/edit strip the route prefix before delegating but
    do not rewrite paths embedded in error strings. That confuses agents that
    try to act on the stripped path. Restore the user-visible path here.
    """
    def __init__(self, route_prefix: str) -> None:
        super().__init__()
        self._prefix = route_prefix.rstrip("/")

    def _restore(self, key: str) -> str:
        return f"{self._prefix}{key}" if key.startswith("/") else f"{self._prefix}/{key}"

    def write(self, file_path: str, content: str) -> WriteResult:
        res = super().write(file_path, content)
        if res.error:
            res = replace(res, error=res.error.replace(file_path, self._restore(file_path)))
        return res

    def edit(self, file_path, old_string, new_string, replace_all=False):
        res = super().edit(file_path, old_string, new_string, replace_all=replace_all)
        if res.error:
            res = replace(res, error=res.error.replace(file_path, self._restore(file_path)))
        return res
```

Then in `backend`:

```python
routes={
    BUILTIN_SKILL_SOURCE: _PrefixedStateBackend(BUILTIN_SKILL_SOURCE),
    "/shared/": _PrefixedStateBackend("/shared/"),
},
```

Worth filing upstream against deepagents too, but this unblocks merge.

src/aiq_agent/agents/deep_researcher/prompts/researcher.j2
```diff
@@ -62,8 +71,7 @@ List every source used, one per line, using ONLY URLs from search tool results:
 [2] Source Title: URL
 Do NOT add URLs from memory — only use URLs that appeared in tool call responses.
 
-Write this output using write_file to /shared/[query_topic_x].txt and return.
-Paths are VIRTUAL.
+Write this output using the `write_file` filesystem tool with `file_path="/shared/[query_topic_x].txt"` and return.
```
AjayThorve, line 74: Two pieces of context are missing here that the orchestrator prompt has but the researcher doesn't, and they together caused the loop in my run. The researcher uses `nemotron-nano-30b-a3b` (4K context) — it can't infer either of these from tool docstrings alone:

1. **`/shared/` is virtual.** Like `orchestrator.j2:23` says, `execute`/shell never sees it. When `write_file` errors, the researcher reaches for `rm`/`ls`/`stat` and ends up debugging a phantom filesystem.
2. **`write_file` is no-overwrite.** If the path exists, the call hard-fails with the "Cannot write … already exists" error. There is no `--force`. Use `edit_file`, or pick a new filename.

Suggested patch (4 lines):

```jinja
Write this output using the `write_file` filesystem tool with `file_path="/shared/[query_topic_x].txt"` and return.

Notes on the filesystem:
- `/shared/` is a virtual path. Do NOT inspect it via `execute`/shell — `ls /shared`, `rm`, `stat` will fail. Use `read_file`, `ls`, `glob` instead.
- `write_file` cannot overwrite. If you need to replace content, use `edit_file`. Do not write to the same path twice in one turn.
```

Same patch should land in the skills block of `orchestrator.j2:14`. The orchestrator (`gpt-oss-120b`) handles this correctly today; the researcher consistently doesn't.

src/aiq_agent/agents/deep_researcher/register.py
```diff
@@ -118,13 +127,15 @@ async def _run(state: DeepResearchAgentState) -> DeepResearchAgentState:
...
-            if data_sources is not None and selected_tools != tools:
+            if config.sandbox is not None or (data_sources is not None and selected_tools != tools):
                 active_agent = DeepResearcherAgent(
                     llm_provider=provider,
                     tools=selected_tools,
                     max_loops=config.max_loops,
                     verbose=verbose,
                     callbacks=callbacks,
+                    skills=config.skills,
+                    sandbox=config.sandbox,
```
AjayThorve, line 148: `SANDBOX.md` says "the Modal sandbox name is the resolved job ID … prevents unrelated jobs from sharing sandbox filesystem state". For async jobs that doesn't currently hold:

- The outer `agent` (line 114) is constructed once at registration with no `job_id`, so `DeepAgentsRuntime` falls back to `str(uuid4())`.
- `_run` rebuilds `DeepResearcherAgent` per request whenever `config.sandbox is not None` (line 130) and again doesn't pass `job_id`. Each request gets a fresh uuid sandbox name even though the async job has a real id available via `nat.builder.context.Context.get().workflow_run_id` (set in `runner.py:365`).

Net effect: every async deep-research call cold-starts a brand-new Modal sandbox; the per-job scoping promise in the doc is just not wired up.

Minimal fix in `_run`:

```python
from nat.builder.context import Context
job_id = Context.get().workflow_run_id  # may be None for sync/eval paths

if config.sandbox is not None or (data_sources is not None and selected_tools != tools):
    active_agent = DeepResearcherAgent(
        llm_provider=provider,
        tools=selected_tools,
        max_loops=config.max_loops,
        verbose=verbose,
        callbacks=callbacks,
        skills=config.skills,
        sandbox=config.sandbox,
        job_id=job_id,  # falls back to uuid in DeepAgentsRuntime when None
    )
```

src/aiq_agent/agents/deep_researcher/deepagents_runtime.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+def _strip_builtin_skill_source(file_path: str) -> str:
+    if file_path.startswith(BUILTIN_SKILL_SOURCE):
+        return "/" + file_path[len(BUILTIN_SKILL_SOURCE) :].lstrip("/")
+    return file_path
+
+
+def _validate_modal_sandbox_name(job_id: str) -> str:
+    if len(job_id) > 64 or re.match(r"^[a-zA-Z0-9-_.]+$", job_id) is None or re.match(r"^ap-[a-zA-Z0-9]{22}$", job_id):
```
AjayThorve, line 245:
```python
if len(job_id) > 64 or re.match(r"^[a-zA-Z0-9-_.]+$", job_id) is None or re.match(r"^ap-[a-zA-Z0-9]{22}$", job_id):
    raise ValueError(...)
```

The third clause rejects strings that *do* match `^ap-[A-Za-z0-9]{22}$`. I assume the intent is "don't let callers pass raw Modal object IDs as sandbox names" — if so, fine, but a one-line comment explaining that would save a future reader from thinking it's inverted. Otherwise this looks like a bug.

AjayThorve (commented):
Tested this branch with a deep-research run that exercises both skills and the Modal sandbox. The researcher subagent gets stuck in a multi-minute loop on the first `task()` call and never returns. Trajectory analysis points to four concrete things worth addressing — the first one is the actual root cause and is small enough to ship now. Inline.

docs/source/architecture/agents/sandbox.md
cdgamarose-nv: Yeah that was my thinking as well. 
I don't think this documentation will stay forever in its current form. But there will be docs for sandbox in some form. So there are arguments in favour of both. I think I'll move it under `docs/`

deploy/.env.example
```diff
@@ -47,6 +47,17 @@ DASK_DISTRIBUTED__LOGGING__DISTRIBUTED=warning
 # -----------------------------------------------------------------------------
 # AIQ_SUMMARY_DB=
 
+# -----------------------------------------------------------------------------
+# NVIDIA API Key (required for skills sandbox)
+# -----------------------------------------------------------------------------
+NVIDIA_API_KEY=
```
AjayThorve: Redundant key, it already exists in this file, right?

AjayThorve (approved):
Lgtm, just one comment about redundant key

deploy/.env.example
cdgamarose-nv: that was a mistake

PR #224, opened 2026-05-08
Issue 214 data sources job submit

tanleach:
## Summary
- Adds issue 214 data sources job submit handling.

## Note
- REST `data_sources: []` intentionally means no tools at all. Tools not listed in any data source entry (e.g., utility tools like "think") are always included regardless of filtering
- WebSocket semantics are not being aligned in this PR; alignment will be handled in a follow-up.

### DIFF
--- a/docs/source/architecture/agents/clarifier.md
+++ b/docs/source/architecture/agents/clarifier.md
@@ -77,7 +77,7 @@ graph TD
 | Field | Type | Default | Description |
 | ----- | ---- | ------- | ----------- |
 | `messages` | `Annotated[list[AnyMessage], add_messages]` | required | Conversation history with [LangGraph](https://docs.langchain.com/oss/python/langgraph/overview) message reducer |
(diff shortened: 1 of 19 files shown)


cdgamarose-nv:
When a tool is excluded from an agent, I'm still able to submit a job with that data source.

For example, in the config for deep research agent:
```
    exclude_tools: 
      - web_search_tool
      - paper_search_tool
```
image

But then when the agent starts, it cant find any valid tools to call because all are filtered out which causes it to spiral. I think we need to handle this edge case.

tanleach:
> When a tool is excluded from an agent, I'm still able to submit a job with that data source.
> 
> For example, in the config for deep research agent:
> 
> ```
>     exclude_tools: 
>       - web_search_tool
>       - paper_search_tool
> ```
> 
> image
> But then when the agent starts, it cant find any valid tools to call because all are filtered out which causes it to spiral. I think we need to handle this edge case.

Thanks for this, ill look into it.

cdgamarose-nv (approved):
Works well now across all cases. Thanks!

PR #225, opened 2026-05-10
fix problem in websocket with expired access token

DinaLaptii:
I turned an "exploitable long-lived auth" into silent self-healing rotation with two independent layers of defense:

Frontend knows its own exp — proactive rotation before the backend has a chance to complain.
Backend re-validates exp per message — if the frontend somehow misses (sleep, race, any bug), the backend still won't accept work under a dead token; it asks for a reconnect instead.
Under normal operation the user sees nothing (no banners, no lost messages). The session expired banner now fires only in one case — when the refresh token itself genuinely dies and a real re-login is required. The security hole is closed on both sides of the wire, while the UX of long Deep Research sessions is preserved.

Additional cleanup:
- Sessions panel now checks persisted Deep Research job IDs against backend status when opened.
- If a current-user session points at a job removed by backend retention, the frontend prunes that session instead of leaving a dead report link in history.

Validation:
- vitest run src/features/chat/store.spec.ts src/features/layout/components/SessionsPanel.spec.tsx
- eslint src/features/chat/store.ts src/features/chat/types.ts src/features/chat/store.spec.ts src/features/layout/components/SessionsPanel.tsx src/features/layout/components/SessionsPanel.spec.tsx
- tsc --noEmit

### DIFF
--- a/frontends/ui/src/features/chat/hooks/use-websocket-chat.ts
+++ b/frontends/ui/src/features/chat/hooks/use-websocket-chat.ts
@@ -777,7 +777,10 @@ export const useWebSocketChat = (options: UseWebSocketChatOptions = {}): UseWebS
             // `lastSentOutgoingRef` and silently drop the user's payload
             // (the auth_expired handler couldn't re-buffer it).
             lastSentOutgoingRef.current = pending
-            setLoading(false)
+            // Match each send path's loading contract: chat sends clear the
+            // composer spinner after putting the message on the wire, while
+            // HITL answers keep it visible until the backend processes them.
+            setLoading(pending.kind === 'interaction')
           }
           return
         }
--- a/frontends/ui/src/features/chat/hooks/use-websocket-chat.ts
+++ b/frontends/ui/src/features/chat/hooks/use-websocket-chat.ts
@@ -61,6 +61,17 @@ const EMPTY_MESSAGES: ChatMessage[] = []
 const EMPTY_CONVERSATIONS: Conversation[] = []
 
 /**
+ * Buffer entry for an outgoing payload that has to bridge a socket
+ * rotation. Discriminated by `kind` because both chat messages and HITL
+ * interaction responses can hit `auth_expired` at the backend, and the
+ * drain on the freshly-handshaken socket has to dispatch back to the
+ * right `NATWebSocketClient` method.
+ */
+type PendingOutgoing =
+  | { kind: 'message'; content: string; dataSources: string[] }
+  | { kind: 'interaction'; interactionId: string; parentId: string; response: string }
+
+/**
  * Seconds before token expiry to proactively rotate the WebSocket. The
  * backend only validates the JWT at the WS upgrade, so an open socket
  * keeps trusting an expired token until we close + reopen it.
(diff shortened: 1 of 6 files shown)


frontends/ui/src/features/chat/hooks/use-websocket-chat.ts
```diff
@@ -158,8 +173,73 @@ export const useWebSocketChat = (options: UseWebSocketChatOptions = {}): UseWebS
...
+
+  /**
+   * Single rotation primitive. `connect()` runs `onBeforeReconnect`, which
+   * is the only call site for `getSession()` -- routing all refreshes
+   * through one path avoids racing the SessionProvider polling (which
+   * would cause `invalid_grant` with rotating refresh tokens).
+   */
+  const rotateSocket = useCallback((reason: string): void => {
+    const client = wsClientRef.current
+    if (!client) return
```
exactlyallan, line 265: This rotation is vulnerable to an old-socket `onclose` race. `disconnect()` sets the client's shared `isIntentionallyClosed` flag, closes the old socket, and then `connect()` immediately flips that same flag back to `false` before the old socket's async `onclose` is guaranteed to fire. If the stale close arrives after the reconnect starts, `NATWebSocketClient.onclose` can classify it as an unintentional disconnect, drive `onConnectionChange('disconnected' | 'error')`, clear streaming/loading state, and potentially schedule another reconnect while the fresh socket is still connecting.

One way to fix this is to make rotation a first-class client operation that detaches or ignores stale socket handlers before closing:

```ts
rotate = async (): Promise<void> => {
  const oldSocket = this.ws
  if (oldSocket) {
    oldSocket.onclose = null
    oldSocket.onerror = null
    oldSocket.onmessage = null
    oldSocket.onopen = null
    oldSocket.close()
    if (this.ws === oldSocket) this.ws = null
  }
  this.isIntentionallyClosed = false
  this.reconnectCount = 0
  await this.connect()
}
```

Another robust pattern is to capture the socket instance in `setupEventHandlers()` and early-return from every event handler unless `this.ws === socket`. Please add a regression test with two mock sockets where socket A closes after socket B has started connecting; that stale close should not reset streaming state or enqueue another reconnect.

frontends/ui/src/features/chat/hooks/use-websocket-chat.ts
```diff
@@ -701,10 +859,12 @@ export const useWebSocketChat = (options: UseWebSocketChatOptions = {}): UseWebS
         wsClientRef.current = createNATWebSocketClient({
           conversationId,
           callbacks,
+          onBeforeReconnect: refreshAuthBeforeReconnect,
         })
         wsClientRef.current.connect()
       } else {
-        // No conversation ID - shouldn't happen but handle gracefully
```
exactlyallan, line 707: The backend now applies the same expiry gate to `WebSocketUserInteractionResponseMessage`, but this frontend replay logic only protects normal chat messages. `respondToInteraction()` sends directly through `sendInteractionResponse()` and never populates `lastSentMessageRef` / `pendingOutgoingRef`. If a user answers a HITL prompt after the handshake token expires, the backend can return `auth_expired`; the current handler will either replay the previous chat message or rotate with no buffered interaction response, silently losing the user's answer.

A practical fix is to generalize the pending buffer to include both message kinds and drain by kind:

```ts
type PendingOutgoing =
  | { kind: 'message'; content: string; dataSources: string[] }
  | { kind: 'interaction'; interactionId: string; parentId: string; response: string }

const pendingOutgoingRef = useRef<PendingOutgoing | null>(null)
const lastSentOutgoingRef = useRef<PendingOutgoing | null>(null)
```

Then `respondToInteraction()` should run the same stale-token preflight used by `sendMessage`; if stale, buffer `{ kind: 'interaction', ... }` and rotate, otherwise send and record it as `lastSentOutgoingRef`. The `auth_expired` handler can re-buffer that last outgoing item, and the `connected` drain can dispatch either `sendMessage()` or `sendInteractionResponse()` based on `kind`. Please add a regression test for `respondToInteraction()` receiving `auth_expired` and then reconnecting without losing or replacing the HITL response.

exactlyallan (commented):
Two focused comments on the token-rotation state machine. I am not requesting changes formally here, but I think both are worth addressing before merge.

frontends/ui/src/features/chat/hooks/use-websocket-chat.ts
DinaLaptii: Resolved. Implemented both patterns suggested:

NATWebSocketClient.rotate() as a first-class atomic primitive that detaches handlers from the old socket before closing it.
Socket-instance capture in setupEventHandlers with if (this.ws !== socket) return early-returns in every handler (defense-in-depth).
Regression test rotate() ignores late onclose from the rotated-out socket in websocket-client.spec.ts covers the exact scenario you described (A closes after B starts connecting; no onConnectionChange or onError fires).

Bonus: also added rotationInFlight promise coalescing for concurrent rotate() calls (closes the parallel-rotate variant of the same race class), with rotate() coalesces concurrent calls -- only one new socket is created test.

frontends/ui/src/features/chat/hooks/use-websocket-chat.ts
DinaLaptii: Resolved. PendingOutgoing discriminated union exactly as suggested; pendingOutgoingRef and lastSentOutgoingRef both typed against it. respondToInteraction() now mirrors sendMessage()'s stale-token preflight (buffer + rotate when stale, otherwise send and record as lastSentOutgoingRef). The auth_expired handler re-buffers lastSentOutgoingRef regardless of kind, and the connected drain dispatches via kind to either sendMessage or sendInteractionResponse.

Three regression tests cover the surface:

respondToInteraction + auth_expired re-issues the HITL response (not lost, not replaced) — the exact case you asked for.
HITL response after a chat send replays the HITL on auth_expired (no cross-payload poisoning) — guards against the failure mode where the previous chat message was replayed instead of the HITL answer.
respondToInteraction with stale token rotates and drains the HITL response — symmetric stale-token preflight coverage for HITL.

exactlyallan:
@greptile ?

tanleach (approved):
Good to go

PR #226, opened 2026-05-11
fix: add headless mode header for chat endpoint

cdgamarose-nv:


AjayThorve (approved):
Good catch, this was indeed missing

PR #227, opened 2026-05-12
fix(citation): register MCP tool results as sources when no URLs present

tanleach:


AjayThorve:
@tanleach we should probably add 

 "nvidia-nat[langchain,async_endpoints,phoenix,mcp]==1.6.0",
 
 nvidia-nat-mcp dependency in pyprojects

src/aiq_agent/common/citation_verification.py
```diff
@@ -747,9 +811,16 @@ def verify_citations(report_text: str, registry: SourceRegistry) -> CitationVeri
...
 
-    # Remove orphaned inline citations from body
+    # Body fixups:
+    #  * Duplicate citations get rewritten to the canonical number — the
+    #    cited source is real, only the [N] label is wrong.
+    #  * Genuinely invalid citations get stripped — the source is fabricated
+    #    or unverifiable.
     cleaned_body = body
-    for num in removed_numbers:
+    for old_num, canonical_num in duplicate_rewrites.items():
```
AjayThorve, line 820: This only rewrites exact standalone citations like `[2]`. If the model emits a grouped inline citation such as `[1, 2]`, the duplicate `[2]` reference line is removed but the body still contains `[1, 2]`, leaving
  `2` dangling with no reference entry. Can we handle citation numbers inside grouped citation brackets before removing the duplicate ref line?

AjayThorve (changes requested):

AjayThorve (approved):
LGTM

PR #228, opened 2026-05-12
feat(knowledge): add OpenSearch as a first-class knowledge backend

rhossi:
## Summary

Adds OpenSearch as a built-in knowledge backend in AIQ 2.0, alongside the
existing LlamaIndex (ChromaDB) and Foundational RAG (Milvus) options.
Selectable through the standard `backend: opensearch` field in workflow YAML;
no fork, no custom base image.

OpenSearch is broadly available — self-hosted, and as a managed offering on
every major cloud (Amazon OpenSearch Service, Amazon OpenSearch Serverless,
others). Until now, customers running OpenSearch had to write their own
`BaseRetriever`/`BaseIngestor` adapter, rebuild base images to include it,
and handle their environment's authentication themselves — including inside
the Dask workers during ingestion. This PR removes that burden and
establishes a plugin pattern that other native backends can follow.

## What is included

- **`OpenSearchRetriever` + `OpenSearchIngestor`** extending `BaseRetriever` /
  `BaseIngestor`, registered through the knowledge-layer factory, exposed as
  a NAT function (`backend: opensearch` in YAML).
- **Three auth modes**: `none` (development clusters), `basic` (self-hosted
  with username/password), `sigv4` (Amazon OpenSearch Service via
  `aws_service: es`, Amazon OpenSearch Serverless via `aws_service: aoss`).
  Pluggable enough for additional auth schemes to slot in as new Literal
  values without adapter rewrites.
- **Authentication inside the Dask worker process**: the ingestion task
  constructs its own OpenSearch client inside each worker, so credentials
  resolve from the worker's environment — works with EKS Pod Identity,
  IRSA, SSO, env profiles, and instance profiles. No signer state is
  serialized across the cluster.
- **Session isolation**: AIQ collections map to per-session OpenSearch
  indexes (`<prefix>-s_<uuid>`) inside a single persistent OpenSearch
  cluster, with 24-hour TTL cleanup driven by `_meta.updated_at`. Lets
  multiple sessions share one OpenSearch resource without cross-session
  leakage.
- **Helm chart override surface**: example values at
  `deploy/helm/examples/aws-opensearch-serverless-values.yaml` shows the
  `image`, `imagePullSecrets`, and `secretEnv` shape for a managed-serverless
  EKS deployment. The chart's existing schema accepts the override directly —
  no chart fork required to point at a custom image or environment.
- **Reference workflow YAML**: `configs/config_web_opensearch.yml`
  (env-substitution driven; the same file works for self-hosted, managed,
  and serverless OpenSearch backends).
- **Reference deployment guide**:
  `docs/source/deployment/aws-opensearch-serverless.md` — end-to-end
  walkthrough for one concrete cloud (Amazon OpenSearch Serverless on EKS
  with Pod Identity), covering architecture, prerequisites, OpenSearch
  collection setup, IAM trust policy, data access policy, Pod Identity
  association, image pull secret, embedding endpoint, verification, and
  cleanup. The same pattern adapts to other cloud or self-hosted OpenSearch
  deployments.

## Test plan

All paths validated against a real Amazon OpenSearch Serverless collection
(SigV4 via assumed role) prior to PR submission:

- [x] Adapter unit tests (`tests/knowledge_layer_tests/test_opensearch_adapter.py`) — 19 passed.
- [x] AOSS live test suite (`tests/knowledge_layer_tests/test_opensearch_serverless_live.py`) —
      2 passed in 160s. Covers SigV4 health, collection lifecycle, vector
      ingest, k-NN retrieval, filtered k-NN, and AOSS-aware delete batching.
- [x] Reference YAML loads cleanly under `nat serve` against a live
      OpenSearch endpoint.
- [x] Local-mode ingestion: `POST /v1/collections/<name>/documents` → embed
      → bulk index → status `completed`. Verified via direct doc-count
      query.
- [x] Dask-mode ingestion: separate scheduler + worker; `_bulk` issued from
      the worker process (verified absent from backend log). Authentication
      resolved inside the worker.
- [x] File deletion: search-then-bulk-delete pattern with eventual visibility
      wait. Doc count drops to 0 in the target index; other indexes
      unaffected.
- [x] Real PDF ingestion: 30-page paper → 30 chunks via `pypdf` → multi-batch
      embedding → bulk index in 5.3s. Page-level citations preserved through
      retrieval.
- [x] UI flow with session-bound collection: `<prefix>-s_<uuid>` index
      appears in OpenSearch alongside API-driven collections.
- [x] Sphinx docs build under strict mode (`SPHINXOPTS="-W --keep-going -n"`)
      — zero warnings.
- [x] Reference YAML and helm example values both parse cleanly.

## CONTRIBUTING.md compliance

- All 26 commits carry `Signed-off-by:` trailers per the DCO requirement
  (`git rebase --signoff` across the branch).
- Manual testing complete per §4 (no CI/CD in this repo today).

## Release Notes

- **New Features**
  - Added an OpenSearch-backed knowledge backend for ingestion and retrieval (self-hosted and Amazon OpenSearch Serverless), including async/Dask ingestion support and retrieval filters.
  - Introduced a “Web mode” configuration combining real-time web search with OpenSearch-backed knowledge retrieval and async deep-research workflows.

- **Documentation**
  - Added end-to-end Amazon OpenSearch Serverless deployment guides, Helm examples/overrides, and Knowledge Layer setup/routing troubleshooting updates.

- **Tests**
  - Added comprehensive OpenSearch adapter tests plus opt-in live OpenSearch/AOSS integration tests.

- **Chores**
  - Added pytest markers for focused test selection.


AjayThorve:
Thank you @rhossi , we will review this soon

AjayThorve:
@KyleZheng1284 can you review this PR

rhossi:
@coderabbitai resume

configs/config_web_opensearch.yml
```diff
@@ -0,0 +1,207 @@
...
+    llm: nemotron_nano_llm  # replace with nemotron_super_llm if available
+    planner_llm: nemotron_nano_llm  # replace with nemotron_super_llm if available
+    # tools: omitted -> inherits all from data_source_registry
+    # exclude_tools: []
+    max_turns: 3
+    enable_plan_approval: true
+    log_response_max_chars: 2000
+    verbose: true
+
+  shallow_research_agent:
```
KyleZheng1284, line 159: can you configure the models to be the latest super 120b like what we have on latest of develop https://github.com/NVIDIA-AI-Blueprints/aiq/blob/develop/configs/config_web_frag.yml

sources/knowledge_layer/src/opensearch/adapter.py
```diff
@@ -0,0 +1,1618 @@
...
+        thread = threading.Thread(
+            target=self._monitor_dask_ingestion,
+            args=(job_id, future, client, file_paths, job_config),
+            daemon=True,
+        )
+        thread.start()
+
+    def _worker_config(self, job_config: dict[str, Any]) -> dict[str, Any]:
+        """Return a sanitized copy of job_config safe to serialize and send to Dask workers."""
+        worker_config = dict(job_config)
```
KyleZheng1284, line 769: Keep OpenSearch credentials out of the Dask task payload

`job_config` contains the basic-auth password populated by `_setup_backend()`, and this shallow copy is passed to `client.submit()`. That serializes the password through the Dask scheduler and workers, where task arguments may be retained or exposed through diagnostics.

can we ensure the worker configuration never contains the password. Workers should resolve `OPENSEARCH_PASSWORD` or an injected secret locally, as they already do for AWS credentials. The public config field should also use `SecretStr` so normal model representations and serialization redact it. Please add a test asserting the submitted worker config contains no credential material.

sources/knowledge_layer/src/opensearch/adapter.py
```diff
@@ -0,0 +1,1618 @@
...
+        self.bulk_refresh = self.config.get("bulk_refresh")
+        if self.bulk_refresh is None:
+            self.bulk_refresh = False if self.auth_type == "sigv4" and self.aws_service == "aoss" else True
+
+        self._client = None
+        self._client_lock = threading.RLock()
+
+    def _index_name_for_collection(self, collection_name: str) -> str:
+        """Return the OpenSearch index name for the given collection."""
+        collection_part = _sanitize_index_part(collection_name, "default")
```
KyleZheng1284, line 344: Make logical-to-physical collection mapping collision-safe

This normalization is not injective: `Tenant A`, `tenant-a`, `tenant/a`, and `TENANT+A` all map to `aiq-tenant-a`. `_ensure_index()` then reuses the existing index without checking its stored `collection_name`, so two logical collections can read from and delete the same physical index.

Can you add a stable disambiguator derived from the original name and reject reuse whenever `_meta.collection_name` does not exactly match the requested logical name. Tests should cover normalization collisions and verify that deleting one collection cannot delete another.

sources/knowledge_layer/src/opensearch/adapter.py
```diff
@@ -0,0 +1,1618 @@
...
+        self,
+        name: str,
+        description: str | None = None,
+        metadata: dict[str, Any] | None = None,
+    ) -> CollectionInfo:
+        """Create a new OpenSearch collection (index) and return its info."""
+        index_name = self._ensure_index(name, description)
+        meta = self._get_index_meta(index_name)
+        if metadata:
+            meta.update(metadata)
```
KyleZheng1284, line 934: Prevent caller metadata from overwriting reserved index metadata

`metadata` comes directly from the public collection API, so this update allows callers to replace `backend`, `collection_name`, embedding configuration, and lifecycle timestamps. For example, changing `backend` makes the index disappear from `list_collections()`, while changing `collection_name` breaks identity checks.

Can you keep user metadata under a dedicated nested key, or explicitly reject reserved keys, and add coverage showing that caller metadata cannot modify adapter-owned fields.

sources/knowledge_layer/src/opensearch/adapter.py
```diff
@@ -0,0 +1,1618 @@
...
+            body = []
+            for hit in hits:
+                hit_id = hit.get("_id")
+                if hit_id and hit_id not in seen_ids:
+                    seen_ids.add(hit_id)
+                    body.append({"delete": {"_index": index_name, "_id": hit_id}})
+
+            if not body:
+                stale_batches += 1
+                if stale_batches >= 2:
```
KyleZheng1284: Are we treating a stale first search page as completed deletion

This seems to repeatedly queries only the first `bulk_batch_size` hits. After deleting that page, AOSS can continue returning the same IDs until its search view refreshes; because those IDs are already in `seen_ids`, two empty action batches return a partial count. The caller then treats any positive count as successful deletion, potentially leaving later chunks behind.

Can you enumerate the complete matching ID set using stable pagination before deleting, or wait for a confirmed-empty result with a refresh-aware timeout and fail rather than return partial success. Add a test with more than one page where the first deleted page remains visible.

KyleZheng1284 (changes requested):
The OpenSearch integration is well-scoped and the core ingestion/retrieval path is promising. I’m requesting changes for four issues that affect security and data correctness: basic-auth credentials entering Dask task arguments, collision-prone logical collection naming, caller metadata overwriting adapter-owned fields, and AOSS deletion reporting success after a partial delete. These should be addressable without redesigning the integration.

sources/knowledge_layer/src/opensearch/adapter.py
rhossi: Fixed in 4adec45. `_worker_config` now strips `username`/`password` before the task is submitted; workers resolve `OPENSEARCH_USERNAME`/`OPENSEARCH_PASSWORD` from their own environment (as SigV4 already does), and distributed basic-auth ingestion fails fast if they are absent. `opensearch_password` is now a `SecretStr`, so it redacts in model reprs/dumps. Added `test_worker_config_excludes_credentials` asserting the submitted worker config contains no credential material, plus fail-fast and SecretStr-redaction tests. The worker-env requirement is documented in KNOWLEDGE-LAYER-SETUP.md.

sources/knowledge_layer/src/opensearch/adapter.py
rhossi: Fixed in 4adec45. `_index_name_for_collection` now appends a stable uuid5 disambiguator derived from the original collection name, so distinct logical names map to distinct physical indexes at every call site (create/read/delete). `_ensure_index` rejects reuse when the stored `_meta.collection_name` does not exactly match the requested name. Tests cover normalization collisions and verify that deleting one collection cannot delete another.

sources/knowledge_layer/src/opensearch/adapter.py
rhossi: Fixed in 4adec45. `create_collection` now rejects reserved keys (`backend`, `collection_name`, `embedding_model`, `embedding_dim`, `created_at`, `updated_at`) with a `ValueError` instead of merging them into `_meta`. Tests confirm caller metadata cannot modify adapter-owned fields and that non-reserved metadata is still stored.

sources/knowledge_layer/src/opensearch/adapter.py
rhossi: Fixed in 4adec45. `_delete_file_documents_for_aoss` now enumerates the complete matching ID set via stable `search_after` pagination (sorted by `chunk_id`) before deleting anything, so a stale search view can no longer mask later pages as complete. If the set cannot be fully enumerated within the batch cap, it raises rather than returning a partial count. Added a multi-page deletion test and a test asserting that a persistently stale view fails instead of reporting partial success.

configs/config_web_opensearch.yml
rhossi: Done in 4adec45. `config_web_opensearch.yml` now uses `nvidia/nemotron-3-super-120b-a12b` — activated `nemotron_super_llm` and pointed the intent, clarifier, shallow-research, and deep-research agents at it, matching `configs/config_web_frag.yml` on develop.

configs/config_web_opensearch.yml
```diff
@@ -0,0 +1,183 @@
...
+
+  deep_research_agent:
+    _type: deep_research_agent
+    orchestrator_llm: nemotron_super_llm
+    researcher_llm: nemotron_super_llm
+    planner_llm: nemotron_super_llm
+    # tools: omitted -> inherits all from data_source_registry
+    exclude_tools:
+      - web_search_tool
+    max_loops: 2
```
KyleZheng1284: The shipped OpenSearch workflow does not currently validate on this head. Running:

```text
uv run nat validate --config_file configs/config_web_opensearch.yml
```

fails with `Invalid configuration: functions: Extra inputs are not permitted`. Inspecting the underlying Pydantic error identifies `functions.deep_research_agent.deep_research_agent.max_loops` as the forbidden field. `config_web_frag.yml` validates successfully, and removing only `max_loops` from this file also makes the OpenSearch config validate.

Can you rebuild this agent block from the current canonical web/FRAG config and add `nat validate --config_file configs/config_web_opensearch.yml` to the validation evidence?

KyleZheng1284 (commented):
I revalidated the latest head after the review fixes. The focused OpenSearch checks pass (`36 passed`), and the full knowledge-layer suite reports `87 passed, 6 skipped`. Four skips are live OpenSearch coverage: two self-hosted tests and two AOSS tests. Because `4adec45` changes the physical index naming and AOSS deletion algorithm, please rerun `AIQ_OPENSEARCH_SERVERLESS_LIVE_TESTS=1 uv run pytest tests/knowledge_layer_tests/test_opensearch_serverless_live.py -q` against the latest head and add the result to the PR before approval.

configs/config_web_opensearch.yml
rhossi: Fixed in be31fe9. Rebuilt the `deep_research_agent` block from the canonical `configs/config_web_frag.yml` on develop: removed the unsupported `max_loops` and added `enable_citation_verification: true`, `source_router_llm`, and `writer_llm` (all `nemotron_super_llm`).

Validation evidence:

```text
$ uv run nat validate --config_file configs/config_web_opensearch.yml
Validating configuration file: configs/config_web_opensearch.yml
✓ Configuration file is valid!
```

rhossi:
@KyleZheng1284 Re: the AOSS serverless live rerun on the latest head (be31fe9).

Ran the serverless live suite against a live Amazon OpenSearch Serverless (SigV4/`aoss`) VECTORSEARCH collection in `us-east-1`, which exercises the changed physical index naming and the new enumerate-before-delete AOSS path end-to-end:

```text
AIQ_OPENSEARCH_SERVERLESS_LIVE_TESTS=1 \
OPENSEARCH_URL=https://<collection>.us-east-1.aoss.amazonaws.com \
AWS_REGION=us-east-1 \
uv run pytest tests/knowledge_layer_tests/test_opensearch_serverless_live.py -q

..                                                                       [100%]
2 passed in 167.14s (0:02:47)
```

Covered: `test_aoss_sigv4_health_and_collection_lifecycle` and `test_aoss_vector_ingest_retrieve_filter_and_delete` (health check, collection lifecycle, vector ingest/retrieve/metadata-filter, and file deletion via the search_after enumeration path).

Config validation evidence (per the inline comment):

```text
uv run nat validate --config_file configs/config_web_opensearch.yml
✓ Configuration file is valid!
```

KyleZheng1284 (commented):
Thanks for addressing the earlier findings—the core fixes and updated AOSS live results look good. We’re evaluating this for AI-Q 2.2, so I did one final developer-facing pass of the AWS setup docs. I found three copy/paste issues below that should be fixed so the guide works on a fresh cluster.

docs/source/deployment/aws-opensearch-serverless.md
```diff
@@ -0,0 +1,478 @@
...
+        repository: <registry>/<aiq-agent-image>
+        tag: <tag>
+```
+
+## Verify the deployment
+
+### 1. Pod is running and Pod Identity is attached
+
+```bash
+kubectl -n ns-aiq get pods -l app.kubernetes.io/name=aiq-agent
```
KyleZheng1284: I matched these commands against the Helm chart, and neither `svc/aiq-agent` nor the `aiq-agent` label exists. The chart creates `svc/aiq-backend` with `app=aiq-backend`. Can we update the pod, log, and port-forward commands so they work when copied?

docs/source/deployment/aws-opensearch-serverless.md
```diff
@@ -0,0 +1,478 @@
...
+**AOSS visibility delay.** AOSS is eventually consistent for search after writes. A `_count` immediately
+after a successful upload may report `0` for ~5–30 seconds before catching up. If the AIQ status says
+`completed` but the AOSS console index browser shows zero docs, wait 30s and refresh — the index will
+populate. This is also why the live-test suite includes a polling visibility wait.
+```
+
+### 5. Run a knowledge query
+
+```bash
+curl -sf -X POST http://localhost:8000/v1/chat/completions \
```
KyleZheng1284, line 415: This uploads into `smoke`, but the chat request doesn’t send `conversation-id: smoke`, so it won’t reliably query that collection. The polling states should also use the lowercase API values (`completed`/`processing`), and the physical index is now `aiq-smoke-<hash>`, not `aiq-smoke`. Can we update the smoke test end to end?

deploy/helm/examples/aws-opensearch-serverless-values.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+aiq:
+  apps:
+    backend:
+      image:
+        # Override this when testing a custom image that includes unreleased changes.
+        repository: nvcr.io/nvidia/blueprint/aiq-agent
+        tag: "2.0.0"
+        pullPolicy: IfNotPresent
+      imagePullSecrets:
```
KyleZheng1284, line 17: The NGC pull secret is only applied to the backend, but the frontend image also comes from `nvcr.io` and defaults to no pull secret. On a fresh cluster this can leave the frontend in `ImagePullBackOff`. Can we apply the same secret to the frontend or configure it globally?

docs/source/deployment/aws-opensearch-serverless.md
rhossi: Fixed in f3ccdab. The chart creates `svc/aiq-backend` with selector `app=aiq-backend` (helper `aiq.appName` = `<project.name>-<appName>`), so the pod, describe, port-forward, and logs commands now use `-l app=aiq-backend` / `svc/aiq-backend`.

docs/source/deployment/aws-opensearch-serverless.md
rhossi: Fixed in f3ccdab, end to end:
- The chat request now sends `-H 'conversation-id: smoke'`. The backend maps `conversation-id` to the collection name (`chat_researcher/register.py`: `collection_name = Context.get().conversation_id`), so this is required to query the `smoke` collection.
- Poll states now use the lowercase API values `completed`/`processing` (matching the `JobState` enum), replacing `SUCCESS`/`INGESTING`.
- The expected physical index is now documented as `aiq-smoke-<hash>` (the naming disambiguator appends a stable suffix to the logical `smoke` collection).

deploy/helm/examples/aws-opensearch-serverless-values.yaml
rhossi: Fixed in f3ccdab. Added `apps.frontend.imagePullSecrets` with the same `ngc-image-pull-secret` to the example values. The frontend image is also `nvcr.io`-hosted and previously fell back to the empty default, so on a fresh cluster it would land in ImagePullBackOff.

sources/knowledge_layer/src/opensearch/adapter.py
```diff
@@ -0,0 +1,1665 @@
...
+
+        if self.auth_type == "basic" and not (
+            os.environ.get("OPENSEARCH_USERNAME") and os.environ.get("OPENSEARCH_PASSWORD")
+        ):
+            raise RuntimeError(
+                "Distributed OpenSearch ingestion with basic auth requires OPENSEARCH_USERNAME and "
+                "OPENSEARCH_PASSWORD to be set in the worker environment; credentials are never sent "
+                "through the Dask scheduler as task arguments."
+            )
+        return worker_config
```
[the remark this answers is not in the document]
rhossi: Addressed in f3ccdab. `_worker_config` now logs a warning when a job requests `generate_summary` under distributed (Dask) ingestion, since the summary LLM is not worker-serializable and summaries are only produced in local mode. Added a matching note in KNOWLEDGE-LAYER-SETUP.md (summaries section) and a test asserting the warning fires.

docs/source/deployment/aws-opensearch-serverless.md
```diff
@@ -0,0 +1,483 @@
...
+curl -sf -X POST http://localhost:8000/v1/collections/smoke/documents \
+  -F 'files=@README.md'
+```
+
+Expected: a `job_id` is returned. Poll `GET /v1/documents/{job_id}/status` until `status` is
+`completed`. If it stalls in `processing`, check the Dask worker logs for SigV4 errors:
+
+```bash
+kubectl -n ns-aiq logs -l app=aiq-backend --tail=200 | grep -i opensearch
+```
```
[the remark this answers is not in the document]
rhossi: Fixed in fa2c3ab. Note: there are no separate Dask worker pods here — `deploy/entrypoint.py` starts the Dask scheduler and workers embedded in the backend pod (`NAT_DASK_SCHEDULER_ADDRESS=tcp://localhost:...`), so `-l app=aiq-backend` is the right selector. Dropped the `| grep -i opensearch` filter so a SigV4/credential error whose message does not contain "opensearch" is no longer hidden; the command now reads the raw `--tail=200`.

sources/knowledge_layer/KNOWLEDGE-LAYER-SETUP.md
```diff
@@ -317,6 +501,10 @@ Summaries are generated for the following file types:
 
 Other file types are ingested normally but do not receive summaries.
 
+> **Note:** Summaries are only generated during **local** ingestion. In distributed (Dask) OpenSearch
+> ingestion the summary LLM is not worker-serializable, so `generate_summary` is forced off and a
+> warning is logged; use `opensearch_ingestion_mode: local` if you require summaries.
+
```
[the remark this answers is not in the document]
rhossi: Fixed in fa2c3ab. Converted the distributed-summary note from a blockquote to a plain paragraph so it is no longer adjacent to the following blockquote; markdownlint MD028 no longer fires for this note.

docs/superpowers/plans/2026-05-03-opensearch-eks-reference-deployment.md
```diff
@@ -0,0 +1,908 @@
...
+```
+
+The backend pod and every Dask worker assume the same IAM role through the EKS Pod Identity
+association on the `aiq-backend` service account. Each Dask worker constructs its own OpenSearch
+client, so SigV4 signing happens in the worker's process — no signer state is serialized across
+the cluster.
+
+```text
+
+(The triple-backtick `{mermaid}` fence is the MyST/Sphinx mermaid directive.)
```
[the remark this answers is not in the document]
rhossi: Resolved in 7c5a91e by removing the file. `docs/superpowers/plans/` was internal development planning scratch (not wired into the docs toctree/build) and carried hardcoded personal paths, so it should not have shipped in this PR — removed rather than polished.

docs/superpowers/plans/2026-05-03-opensearch-eks-reference-deployment.md
```diff
@@ -0,0 +1,908 @@
...
+
+```
+
+```bash
+curl -sf "http://localhost:8000/v1/collections" | jq
+
+```
+
+Expected: `aiq-smoke` index visible in the AOSS console under the collection's index browser,
+and the `smoke` collection listed by the AIQ API.
```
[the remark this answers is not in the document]
rhossi: Resolved in 7c5a91e by removing the file. `docs/superpowers/plans/` was internal development planning scratch (not wired into the docs toctree/build) and carried hardcoded personal paths, so it should not have shipped in this PR — removed rather than polished.

docs/superpowers/plans/2026-05-05-opensearch-gaps-and-risks.md
```diff
@@ -0,0 +1,504 @@
...
+        "embed_base_url": "http://nim-embed.ns-nim.svc.cluster.local:8000/v1",
+        "start_ttl_cleanup": False,
+    })
+    # Patch the OpenAI client so we don't hit the network; the test just
+    # asserts no early-exit RuntimeError fired.
+    class _FakeOpenAI:
+        def __init__(self, base_url, api_key):
+            self.embeddings = type("E", (), {"create": staticmethod(lambda **kw: type("R", (), {"data": [type("D", (), {"embedding": [0.0] * 4})()]})())})()
+    monkeypatch.setattr("openai.OpenAI", _FakeOpenAI)
+    result = ingestor._embed_texts(["hello"])
```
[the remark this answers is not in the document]
rhossi: Resolved in 7c5a91e by removing the file. `docs/superpowers/plans/` was internal development planning scratch (not wired into the docs toctree/build) and carried hardcoded personal paths, so it should not have shipped in this PR — removed rather than polished.

KyleZheng1284 (approved):

PR #230, opened 2026-05-12
feat: enhance citation handling with confidence and match kind

rkarmaka:
- Updated the citation emission process to include confidence scores and match kinds when a SourceRegistry is attached, allowing the UI to display verification badges.
- Modified the API interfaces and internal state management to accommodate the new citation attributes.
- Added tests to ensure proper handling of confidence and match kind in citation updates.
- Implemented a confidence threshold in the DeepResearcherAgent to filter citations based on their strength before inclusion in reports.

This change improves the user experience by providing clearer insights into citation validity and enhances the overall citation verification process.


AjayThorve:
Thank you @rkarmaka, we will review this soon

src/aiq_agent/common/citation_verification.py
```diff
@@ -792,15 +1048,47 @@ def verify_citations(report_text: str, registry: SourceRegistry) -> CitationVeri
...
-    if not removed_citations:
-        logger.debug("[CitationVerify] Result: all %d citation(s) valid — no changes", len(valid_citations))
-        verified = body + ref_section if url_replacements else report_text
+    # The report keeps every citation that resolved, even with low confidence.
+    # Only genuinely unresolved (unmatched / ambiguous / unverifiable) entries
+    # get stripped — those are likely fabrications. ``verified`` is reported
+    # separately on each record so the UI can render strength.
+    valid_citations = [_citation_to_dict(v) for v in verifications if v.resolved]
+    removed_records = [v for v in verifications if not v.resolved]
+    removed_citations = [_citation_to_dict(v, include_reason=True) for v in removed_records]
```
torkian: Move the initial build of `valid_citations`/`removed_records`/`removed_citations` (currently at lines 1055-1057) above the dedup block. Fixes the NameError and prevents the dedup's output from being silently overwritten.

```suggestion
    # The report keeps every citation that resolved, even with low confidence.
    # Only genuinely unresolved (unmatched / ambiguous / unverifiable) entries
    # get stripped — those are likely fabrications. ``verified`` is reported
    # separately on each record so the UI can render strength.
    valid_citations = [_citation_to_dict(v) for v in verifications if v.resolved]
    removed_records = [v for v in verifications if not v.resolved]
    removed_citations = [_citation_to_dict(v, include_reason=True) for v in removed_records]

    # Dedup: collapse multiple [N] reference lines that resolve to the same
    # registry source. The model often makes the same tool call twice (e.g.
    # ``mcp_time__get_current_time`` for two timezones) and emits a separate
    # ``[N] tool_name`` line for each call; without this pass both lines
    # survive verification because each is independently valid. We keep the
    # lowest-numbered occurrence and rewrite later inline citations to that
    # number so the prose still cites the source.
    seen_keys: dict[str, int] = {}  # canonical_key -> kept citation number
    duplicate_rewrites: dict[int, int] = {}  # duplicate_num -> canonical_num
    deduped_valid: list[dict] = []
    for c in valid_citations:
        key = c["url"] or c["citation_key"]
        if key is None:
            # Defensive: a valid citation must have one of url/citation_key.
            # If neither is set we cannot dedup, so keep the entry.
            deduped_valid.append(c)
            continue
        canonical_num = seen_keys.get(key)
        if canonical_num is None:
            seen_keys[key] = c["number"]
            deduped_valid.append(c)
            continue
        duplicate_rewrites[c["number"]] = canonical_num
        removed_citations.append(
            {
                "number": c["number"],
                "line": c["line"],
                "reason": f"duplicate_of_citation_{canonical_num}",
            }
        )
        logger.debug(
            "[CitationVerify]   [%d] REMOVE — duplicate of [%d]: %s",
            c["number"],
            canonical_num,
            key,
        )
    valid_citations = deduped_valid

    # Apply URL replacements (garbled -> canonical) in the references section
    if url_replacements:
        for garbled, canonical in url_replacements.items():
            ref_section = ref_section.replace(garbled, canonical)
```

torkian (commented):
Friendly follow-up on the P1 Greptile flagged — I went deeper and there are actually **two related bugs** in this block that have the same root cause and the same fix:

1. **NameError on every report with a references section** (already flagged): the dedup loop at line 1018 reads `valid_citations` before it's defined.
2. **Silent no-op dedup**: even after fixing the NameError, line 1055 (`valid_citations = [_citation_to_dict(v) for v in verifications if v.resolved]`) reassigns `valid_citations`, **discarding** the deduped result from line 1044. Line 1057 similarly clobbers any duplicate entries appended to `removed_citations` inside the loop. So the dedup pass would run, find duplicates, and then have its work thrown away.

Both go away if we initialize `valid_citations` / `removed_citations` / `removed_records` from `verifications` **before** the dedup block instead of after. The dedup then mutates the already-built lists, and the URL-replacement block at the end is unaffected. Suggested change attached on the relevant line range — happy to put it up as a PR against your fork if that's easier.

src/aiq_agent/common/citation_verification.py
rkarmaka: Thanks — merged develop and applied your reorder. Worth flagging: your fix cleared the NameError, but my refactor's if not removed_records early-return was also discarding the deduped result (and skipping duplicate-line stripping), since removed_records is unresolved-only.

torkian: Nice catch on the `removed_records` vs `removed_citations` distinction — that's the subtle one. Keying the early-return off `removed_citations` (the superset of unresolved + dedup entries) is the right call: a report with duplicate citations but no unresolved ones now correctly strips/rewrites the dup lines instead of bailing early. Pulled `80f24de0` and traced it — the reorder + your early-return fix read correctly together, and `invalid_numbers = removed_numbers - set(duplicate_rewrites)` cleanly separates the strip-vs-rewrite paths. LGTM 👍

cdgamarose-nv:
This PR mostly targets observability and UX - not sure if this is an issue that has been reported or surfaced. Citation handling is still heuristic-based, so I'm not sure if this affects correctness at all. @AjayThorve, what do you think?

cdgamarose-nv:
Thank you for your contribution @rkarmaka
At this time, we're going to keep the current implementation of the citation verification and instead make it optional for workflows.

PR #231, opened 2026-05-12
fix helm bootstrap resources

AjayThorve:
## Summary
- Create bootstrap Helm resources for the default source chart install so `aiq-postgres-init` and `aiq-credentials` exist without out-of-band kubectl commands.
- Add app-level `hostAliases` passthrough.
- Repackage the `aiq-0.0.4` dependency used by `deploy/helm/deployment-k8s` and document credential bootstrap behavior.
- Add Helm render regression tests for referenced ConfigMaps/Secrets and `hostAliases`.

## Root Cause
The parent `deployment-k8s` chart referenced a packaged `aiq` dependency whose rendered workloads mounted `aiq-postgres-init` and consumed `aiq-credentials`, but the default install path did not render those resources. Fresh clusters therefore hit missing ConfigMap/Secret pod errors.

## Validation
- `uv run pytest tests/deploy/test_helm_deployment_k8s.py`
- `uv run ruff check tests/deploy/test_helm_deployment_k8s.py`
- `helm lint deploy/helm/helm-charts-k8s/aiq`
- `helm lint deploy/helm/deployment-k8s`
- Fresh Kind repro: `kind create cluster --name aiq-qa-6164580`, `kubectl create namespace ns-aiq`, `helm install aiq deploy/helm/deployment-k8s -n ns-aiq`; verified `configmap/aiq-postgres-init` and `secret/aiq-credentials` were created, backend/postgres reached `1/1 Running`, and no missing ConfigMap/Secret events appeared. Frontend remained blocked only by an arm64 image manifest pull issue for `nvcr.io/nvidia/blueprint/aiq-frontend:2.0.0`, which is separate from NVBug 6164580.


AjayThorve:
Updated the PR to keep the Secret behavior explicit instead of creating placeholder credentials by default.

Current scope:
- The chart now packages/renders the chart-owned `aiq-postgres-init` ConfigMap.
- The source-chart docs now explicitly tell users to create `aiq-credentials` before install.
- `aiq-credentials` must include `DB_USER_NAME`, `DB_USER_PASSWORD`, `NVIDIA_API_KEY`, and `TAVILY_API_KEY` for the default deployment.
- The chart continues to reference `aiq-credentials` as a user-supplied Secret rather than creating empty API key defaults.
- Regression tests assert ConfigMap references are chart-backed and the credential Secret remains user-supplied.

cdgamarose-nv:
This makes more sense to me, looks good.

cdgamarose-nv (approved):

Issue #235, opened 2026-05-13
Allow continuation of deep research from a checkpoint if citation verification module doesn't capture sources


cdgamarose-nv:
If the citation verficiation module fails to capture sources, it crashes the whole deep research job:

```
2026-04-10 18:30:58 - INFO     - aiq_agent.common.callbacks:204 - Here is the corrected report with the required **"## Sources"** section added. The content from the previous version has been retained, and the missing sources list has been inserted at the end.

---  

# Deep Technical Research Report  
**Nemoclaw, OpenShell, and OpenClaw**  
*(All information current as of April 2026)*  

---  

## 1. Executive Summary  

| Component | Domain | Primary Goal | License | Development Status |
|-----------|--------|--------------|---------|--------------------|
| **OpenClaw** | Autonomous AI‑agent framework (the "operating system" for AI agents) | Provide a **self‑hosted, privacy‑first runtime** for AI agents with orchestration, tool sandboxing, memory, and channel adapters. | Apache 2.0 | Production‑ready core, fast‑growing open‑source project (≈247 k ★). |
| **OpenShell** | Secure sandbox runtime for agents | Offer **kernel‑level isolation, policy‑based network & data‑privacy enforcement**, and a lightweight messaging channel for agents. | Apache 2.0 | Early‑preview (alpha); GA targeted Q3 2026. |

---  
l... [truncated, 15699 total chars]
2026-04-10 18:30:58 - INFO     - aiq_agent.common.callbacks:221 - [Tokens] prompt=17543, completion=4059, model=openai/gpt-oss-120b
2026-04-10 18:30:58 - INFO     - aiq_agent.common.callbacks:183 - ------------------------------

2026-04-10 18:30:58 - INFO     - aiq_agent.agents.deep_researcher.agent:377 - Report completed successfully. Reason: complete_via_heuristic
2026-04-10 18:30:58 - ERROR    - aiq_agent.agents.deep_researcher.agent:490 - Deep Research Subagent failed: Research failed: no sources were captured during deep research. All tool calls may have failed or returned no results. Please try again.
Traceback (most recent call last):
  File "/Users/<>/projects/prompt_registry/src/aiq_agent/agents/deep_researcher/agent.py", line 463, in run
    raise EmptySourceRegistryError("deep research")
aiq_agent.common.citation_verification.EmptySourceRegistryError: Research failed: no sources were captured during deep research. All tool calls may have failed or returned no results. Please try again.
[...]

PR #237, opened 2026-05-13
refactor: update citation verification handling in DeepResearcherAgent

rkarmaka:
- Removed the raising of EmptySourceRegistryError when no sources are available during deep research.
- Added logging for cases where reports are generated without captured sources, indicating whether tools were unavailable or if the model answered without using search results.
- Introduced a new citation_verification_status field in DeepResearchAgentState to track unverified reports, including the reason and available tool count.
- Updated tests to ensure correct behavior when the source registry is empty and when sources are captured.

## Summary by CodeRabbit

* **New Features**
  * Deep-research job report endpoints now expose `citation_verification_status` and consistently surface the related warning in returned report text.
  * Report-context, report-edit, and report-rewrite flows now propagate citation-verification disposition through to the final output.
* **Bug Fixes**
  * Deep research no longer fails when citation verification is enabled but no sources or tools are available; reports are returned with an `unverified` disposition.
  * Inline chat-researcher reporting now applies the same warning and disposition handling.
* **Tests**
  * Expanded unit tests for warning prefixing, API field exposure, status sanitization, and verified/unverified/disabled cases.


rkarmaka:
@cdgamarose-nv Solving issue #235

src/aiq_agent/agents/deep_researcher/models/state.py
```diff
@@ -66,3 +72,4 @@ class DeepResearchAgentState(BaseModel):
     subagents: list[dict[str, Any]] = Field(default_factory=list)
     clarifier_result: str | None = None
     available_documents: list[AvailableDocument] | None = None
+    citation_verification_status: dict[str, Any] | None = None
```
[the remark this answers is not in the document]
rkarmaka: Not adopting here — `DeepResearchAgentState` uses untyped `dict[str, Any]` consistently for its structured payload fields (`user_info`, `tools_info`, `files`, `subagents`, `todos`). Adding a typed model just for `citation_verification_status` would break that pattern for one field while the schema is still settling (downstream consumers in `chat_researcher` / API not yet wired).

For this repo: prefer matching the existing `dict[str, Any]` convention on `DeepResearchAgentState` and similar state bags. If we want type safety later, the right move is to convert all of them together, not per-field.

cdgamarose-nv:
Thank you for your contribution @rkarmaka. We will review soon

AjayThorve:
@rkarmaka can you resolve conflicts?

AjayThorve:
@rkarmaka if you can fix the conflicts, we can make sure this goes into 2.2 release

rkarmaka:
@AjayThorve Sorry I missed your earlier comment. I will go ahead and resolve the conflict. Thanks!

src/aiq_agent/agents/deep_researcher/agent.py
```diff
@@ -333,11 +332,27 @@ async def run(self, state: DeepResearchAgentState) -> DeepResearchAgentState:
...
+                    )
+                else:
+                    logger.warning(
+                        "Deep research produced a report but no sources were captured "
+                        "despite %d available tool(s). The model may have answered "
+                        "without using search results. Returning the report unverified.",
+                        available_count,
+                    )
+                if result is not None:
+                    result["citation_verification_status"] = {
```
AjayThorve: **[P1] Do not silently publish the unverified report**

This status is retained only in the internal `DeepResearchAgentState`. Both external response boundaries discard it: the async job runner extracts only `report` and marks the job successful, while inline chat returns only `messages` and `last_report_markdown`. As a result, a report produced with zero captured sources is indistinguishable to users and downstream consumers from a citation-verified report.

Please propagate a typed verification disposition through both job and chat responses and surface an explicit warning for `unverified`. Keep tool-level diagnostics in logs or protected metadata, and add boundary tests for both paths. The existing propagation thread is marked resolved, but this remains reproducible at the current head.

AjayThorve: @rkarmaka can you resolve this

AjayThorve (commented):
Remaining citation-verification contract findings on the current head.

src/aiq_agent/agents/deep_researcher/models/state.py
```diff
@@ -55,6 +116,12 @@ class DeepResearchAgentState(BaseModel):
         rubric: DeepAgents rubric used by RubricMiddleware when available.
         clarifier_result: Log from clarifier agent dialog.
         available_documents: User-uploaded documents with summaries for context.
+        citation_verification_status: Set when citation verification was skipped
+            because the source registry was empty. ``None`` means the report was
```
AjayThorve: **[P1] Do not equate missing status with verified**

The new contract says `None` means the report was verified, but the control flow also leaves it `None` when `enable_citation_verification=False` and when verification runs with a nonempty registry but finds zero valid citations. In the latter case, a report with no citation markers is returned unchanged and the API/UI omit the warning even though no claims were verified.

Make the verification engine return an explicit closed outcome for every run—at minimum `verified`, `unverified`, or `disabled`—with machine-readable reasons such as `no_sources` and `no_valid_citations`. Propagate that outcome instead of inferring success from absence.

frontends/aiq_api/src/aiq_api/jobs/runner.py
```diff
@@ -824,6 +824,9 @@ async def run_agent_job(
                     # Apply caller metadata first, then set the canonical report last so a
                     # stray "report" key in output_metadata can never overwrite the real report.
                     output = {**(output_metadata or {}), "report": report}
+                    citation_verification_status = _extract_citation_verification_status(result)
```
AjayThorve: **[P1] Preserve unverified disposition across derived reports**

This extracts the disposition only from the current agent result. `ReportRewriterAgentState` has no citation status, and `report_output_metadata(parent_job_id, "edit")` does not carry the parent disposition, so an edit of an unverified parent is persisted and returned with `citation_verification_status=null`. The warning is also flattened into report text before persistence, so the rewrite model can alter or drop the only remaining trust signal.

Treat `unverified` as monotonic through report edits and parent-seeded research unless the complete derived report is reverified. Carry the disposition through `ReportContext` and output metadata, and keep canonical report text separate from presentation warnings.

frontends/aiq_api/src/aiq_api/routes/jobs.py
```diff
@@ -397,12 +398,24 @@ class JobStateResponse(BaseModel):
     artifacts: dict | None = Field(None, description="Tool calls, outputs, and sources collected during execution")
 
 
+class CitationVerificationStatusResponse(BaseModel):
+    """Public citation-verification disposition for a generated report."""
+
+    status: str = Field(..., description="Citation verification status, for example 'unverified'")
```
AjayThorve: **[P2] Make the public disposition a closed domain type**

This new API exposes unrestricted `str` fields while the agent state uses `dict[str, Any]` and the TypeScript client uses plain `string`. Invalid status/reason combinations are representable, and `public_citation_verification_status()` silently discards any future status other than `unverified`.

Define the authoritative `CitationVerificationOutcome` alongside `CitationVerificationResult` in `common.citation_verification`, use `Literal` or enum values for status and reason, and map that type into the API DTO. User-facing warning copy should be rendered at the response/UI boundary rather than defining the domain contract.

cdgamarose-nv:
@rkarmaka Following up - would you be able to address the PR comments and resolve merge conflicts?

cdgamarose-nv:
Thanks again for working through this and for keeping the branch updated. Since this PR was opened, #397/#405 changed the empty-source behavior: AI-Q now preserves the sanitized report, returns an actionable typed failure, and makes the report available through the report endpoint. That addresses the original report-loss problem.

After the latest merges, this PR also no longer removes the empty-source exception—the agent still raises before the new verification status is set. The remaining changes are now fairly broad, and there are a couple of unresolved concerns, including empty-source reports not receiving the new status and verification-disabled reports potentially being promoted to “verified” during follow-up.

Given how the underlying behavior has evolved, I’m going to close this PR as no longer needed. If you can still reproduce a user-facing gap on the current develop branch, please reopen the issue (or open a focused new one) with the current reproduction steps. We’d be happy to look at a smaller change targeted specifically at that remaining behavior.

PR #242, opened 2026-05-15
AIQ skill deployment

exactlyallan:
## Summary

- Restructures `aiq-deploy` into a reference-based deployment router for external/local AI-Q users.
- Adds backend-only Agent Skill startup via `scripts/start_as_skill.sh`, with debug UI disabled by default.
- Documents the install/use flow across `aiq-deploy`, `aiq-research`, and minimal evals. 

## Why

External users need a clear path for using AI-Q through Agent Skills without depending on pre-hosted AIQ endpoints. The deploy skill now separates backend-only skill usage, CLI usage, browser UI usage, Docker Compose, Helm, FRAG/RAG integration, shutdown, troubleshooting, and validation.

## Notes

- The default Agent Skill backend path uses `configs/config_web_default_llamaindex.yml`.
- Docker Compose skill-backend guidance starts `aiq-agent` only and uses `BUILD_TARGET=release`.
- `aiq-research` now explicitly hands off to `aiq-deploy` when the backend is unavailable.

## Validation

- `git diff --check`
- `git diff --cached --check`
- Skill frontmatter sanity check
- Claude skill symlink check
- `bash -n scripts/start_as_skill.sh scripts/start_e2e.sh scripts/start_server_in_debug_mode.sh scripts/setup.sh`
- Python AST parse for `frontends/aiq_api/src/aiq_api/plugin.py`


.agents/skills/aiq-deploy/references/env-and-secrets.md
```diff
@@ -0,0 +1,57 @@
...
+from pathlib import Path
+
+env = Path("deploy/.env")
+values = {}
+for line in env.read_text().splitlines():
+    line = line.strip()
+    if not line or line.startswith("#") or "=" not in line:
+        continue
+    key, value = line.split("=", 1)
+    values[key.strip()] = value.strip()
```
peterychang: If we're only checking for the presence of keys, we should probably avoid storing the values. Storing them just feels like an unnecessary risk

.claude/skills/aiq-configure
```diff
@@ -0,0 +1 @@
+../../.agents/skills/aiq-configure
```
peterychang: nit: do these files need a newline? Not sure what our linter is expecting

frontends/aiq_api/src/aiq_api/plugin.py
```diff
@@ -253,13 +253,17 @@ async def shutdown_sse_connections():
...
-            pass
+        enable_debug = os.environ.get("AIQ_ENABLE_DEBUG", "true").lower() not in {"0", "false", "no", "off"}
+        if enable_debug:
+            try:
+                from aiq_debug import register_debug_routes
+
+                await register_debug_routes(app)
+                logger.info("Debug console registered at /debug")
+            except ImportError:
+                pass
```
peterychang, line 264: maybe add a log line stating debugging was not enabled

.agents/skills/aiq-deploy/references/env-and-secrets.md
exactlyallan: Good catch. Fixed in commit 3edc38a: the check now stores only boolean presence flags in `presence` / `runtime_presence` and never retains raw `deploy/.env` values.

.claude/skills/aiq-configure
exactlyallan: This is no longer present. We removed the experimental `aiq-configure` skill/symlink when config guidance was folded back into `aiq-deploy`. The remaining `.claude/skills/*` entries are symlinks, so there is no meaningful newline requirement for those link targets; pre-commit is passing.

frontends/aiq_api/src/aiq_api/plugin.py
exactlyallan: Added in commit 3edc38a: when `AIQ_ENABLE_DEBUG` disables debug route registration, the API now logs `Debug console disabled by AIQ_ENABLE_DEBUG`.

exactlyallan:
Addressed the Greptile findings in `145ee5c`:

- `scripts/start_as_skill.sh` now derives the displayed `AIQ_SERVER_URL` from `--host` / `--port` while keeping wildcard binds (`0.0.0.0`, `::`) user-facing as `localhost`.
- The `front_end` config check now ignores commented YAML lines instead of matching `# front_end:`.
- `deploy/Dockerfile` now resolves `uv python find 3.13` once and reuses that path for both Python symlinks.

Validation: `bash -n scripts/start_as_skill.sh`, a focused `front_end` detection sample check, `git diff --check`, and targeted `uv run pre-commit run --files deploy/Dockerfile scripts/start_as_skill.sh`.

frontends/aiq_api/src/aiq_api/plugin.py
peterychang: I meant specifically in the case where the import fails. In that case, there wouldn't be anything explicitly stating that debugging is not enabled even though the `AIQ_ENABLE_DEBUG` flag was on

peterychang (approved):

tanleach (approved):

PR #245, opened 2026-05-19
fix: clarifier searches before asking clarification questions

### LINKED ISSUE #234: Force clarifier agent to trigger searches before asking questions
If the clarifier LLM doesn't know about something from its internal knowledge (training data), it immediately asks questions. Since we have a clarifier agent with a tool-calling loop, we need to trigger searches to gather more context before asking questions. 

### Example:

Image

### Expected

This should have triggered a web search to find out context and then ask more questions if needed. This should not be overly prompt engineered for one model but instead should work with most frontier and open models.

torkian:
Closes #234.

## Summary

The clarifier agent already had tool-calling bound to a search loop, but weak or non-frontier models routinely skipped the search and asked the user a clarification question immediately. The expected behavior — and what the issue asks for — is to gather context with the available tools first, and only ask the user once a genuine ambiguity remains.

This fixes the bug at the agent level (the issue explicitly says "this should not be overly prompt engineered for one model"), with a small prompt update as a supporting hint.

## Changes

- **Inline search-before-clarify retry in `agent_node`.** On the first turn, if the model returns a clarification request without using its bound search tools, the node nudges it once with `FORCE_SEARCH_GUIDANCE` and re-invokes the LLM inline, returning both the original response and the retry. No extra graph node or state field is needed.
- **One-shot by construction.** The retry fires only when `iteration == 0` (so it never fires after the user has replied) **and** `not _has_tool_invocations(state.messages)` (so once any tool call is in history — e.g. after a successful forced search, even while iteration is still 0 — it never re-nudges). A model that stubbornly refuses to search after the single nudge simply falls through to asking the user; no loop.
- **Guidance stays ephemeral.** `FORCE_SEARCH_GUIDANCE` is sent only in the local `retry_messages` and is never returned to state, so it cannot leak into `get_latest_user_query` or other history lookups (which would otherwise surface internal scaffolding in fallback text).
- **`prompts/research_clarification.j2`** Tool Usage section rewritten as "Search first, ask second" so frontier models follow this voluntarily and skip the nudge entirely.

> Design note: an earlier revision used a dedicated `force_search` graph node plus a `ClarifierAgentState.force_search_used` flag. Per review (thanks @cdgamarose-nv), this was simplified to the inline retry above — same behavior, ~70 fewer source lines, no added graph state.

## How I verified (end-to-end, against a real LLM + web search)

I exercised the full clarifier graph against a real NVIDIA NIM-hosted LLM and a real Tavily web search tool, with model `nvidia/nemotron-3-nano-30b-a3b` (the repo default in `configs/config_cli_default.yml`) and the query *"Research the latest NVIDIA AI announcements from 2026"*:

| Branch | tool_calls emitted | user prompted? |
| --- | --- | --- |
| `develop` | **0** (bug — asked the user immediately) | yes |
| this PR | **1** (searched first) | no |

Deterministic across three runs. (The throwaway reproducer script is not included in the PR, per review — these are its results.)

## Tests

150 clarifier tests pass. Key additions in `tests/aiq_agent/agents/clarifier/test_agent.py`:

- **`TestHasToolInvocations`** — the helper that detects prior tool calls in the message history.
- **`TestClarifierForceSearch`**:
  - `test_force_search_fires_when_llm_skips_tools` — the bug path: model clarifies → nudge + inline retry → model emits a tool call → completes; user never prompted.
  - `test_force_search_skipped_when_no_tools` — no behavior change when no tools are configured.
  - `test_force_search_fires_at_most_once` — no loop if the model stubbornly refuses tools after the nudge.
  - `test_force_search_not_triggered_when_llm_searches_first` — no extra detour for models that already search.
  - `test_force_search_guidance_not_in_state_messages` — regression test: the internal guidance must never appear in user-facing fallback text.
  - `test_force_search_guidance_not_injected_after_user_reply` — regression test: the nudge does not re-fire after the user has answered.
  - `test_forced_retry_does_not_emit_consecutive_assistant_messages` — regression test for the message-ordering fix below.
- **`TestClarifierSkipMessageOrdering`** — regression tests for the skip-path ordering fix (persisted history + planner sequence under `enable_plan_approval`).

## Message-ordering correctness (from review)

Per @cdgamarose-nv's review the implementation was simplified to an inline retry; follow-up review feedback then flagged that returning both the skipped and retried responses produced two consecutive assistant-role messages once the retry carried a tool call, which the OpenAI/Anthropic APIs reject with a 400 (NVIDIA NIM tolerates it). Fixed by returning only the retry response. Auditing the same invariant across the graph surfaced a separate pre-existing instance on the skip-clarification path (skip emitted a completion without persisting the user reply, and graph re-entry appended a duplicate completion) — also fixed, with regression tests for both.

## Test plan

- [x] `pytest tests/aiq_agent/agents/clarifier/` — 150/150 passing
- [x] `ruff check` and `ruff format` clean
- [x] Reproduces the bug on `develop` against the repo's default model
- [x] Verifies the fix on this branch against the same model + query
- [ ] Maintainer smoke test via the CLI / Web UI workflows


cdgamarose-nv:
@torkian Thank you for your contribution. We will review soon.

torkian:
Thanks for the review! Addressed both findings in 8f18f80:

- **P1 (agent.py)**: added `state.iteration == 0` to the agent_node ephemeral-injection condition so the nudge can't be re-applied after the user has already replied. Added `test_force_search_guidance_not_injected_after_user_reply` as a regression test.
- **P2 (e2e script)**: PASS verdict now requires both `tool_calls >= 1` AND `len(user_prompts) == 0`, with distinct FAIL messages for the two failure modes.

Also cleaned up an unused mock side_effect in `test_force_search_guidance_not_in_state_messages` that the review flagged. 149 clarifier tests pass.

scripts/e2e_clarifier_search_test.py
```diff
@@ -0,0 +1,135 @@
+"""End-to-end test for issue #234 fix.
```
cdgamarose-nv: Don't include this script in the PR. You can add test and results in PR description.

src/aiq_agent/agents/clarifier/agent.py
```diff
@@ -610,8 +658,37 @@ def decide_route(state: ClarifierAgentState | dict):
...
+                and iteration == 0
+                and not force_search_used
+                and not self._has_tool_invocations(messages[:-1])
+            ):
+                logger.info("Clarifier: forcing search before first clarification question")
+                return "force_search"
+
             return "ask_for_clarification"
 
+        async def force_search_node(state: ClarifierAgentState):
```
cdgamarose-nv: Wondering if we need to add a node or a variable in state. I was thinking something like this would be sufficient:

```
response = await bound_llm.ainvoke(messages)

if (
    self.tools
    and state.iteration == 0
    and not self._has_tool_invocations(state.messages)
    and not getattr(response, "tool_calls", None)
    and self._is_needed(response.content)
):
    logger.info("Clarifier: model skipped search; injecting guidance and retrying once")
    retry_messages = messages + [response, HumanMessage(content=FORCE_SEARCH_GUIDANCE)]
    retry_response = await bound_llm.ainvoke(retry_messages)
    return {"messages": [response, retry_response]}

return {"messages": [response]}
```

cdgamarose-nv (commented):
Thanks for your PR! Left some comments

scripts/e2e_clarifier_search_test.py
torkian: Done — removed `scripts/e2e_clarifier_search_test.py` in 1377209. Moved the end-to-end test approach and before/after results into the PR description.

src/aiq_agent/agents/clarifier/agent.py
torkian: Good call — adopted your inline approach in 1377209. Dropped the `force_search` node, its routing branch/edges, and the `ClarifierAgentState.force_search_used` field; the nudge-and-retry now lives directly in `agent_node` exactly as you sketched.

The one-shot guarantee holds without the flag from two conditions alone:
- `iteration == 0` — only the first turn; once the user replies, iteration advances and it never fires again.
- `not _has_tool_invocations(state.messages)` — after a successful forced search the tool-call message is in history (even while iteration is still 0), so it won't re-nudge.

`FORCE_SEARCH_GUIDANCE` is passed only in the local `retry_messages` and never returned to state, so it can't leak into `get_latest_user_query`. Net −70 source lines, behavior unchanged (same LLM call sequence + final message history). All 147 clarifier tests pass, including the force-search and ephemeral-guidance regression tests.

src/aiq_agent/agents/clarifier/agent.py
```diff
@@ -548,6 +580,34 @@ async def agent_node(state: ClarifierAgentState):
...
+                self.tools
+                and state.iteration == 0
+                and not self._has_tool_invocations(state.messages)
+                and not getattr(response, "tool_calls", None)
+                and self._is_needed(response.content)
+            ):
+                logger.info("Clarifier: model skipped search before clarifying; injecting guidance and retrying once")
+                retry_messages = messages + [response, HumanMessage(content=FORCE_SEARCH_GUIDANCE)]
+                retry_response = await bound_llm.ainvoke(retry_messages)
+                return {"messages": [response, retry_response]}
```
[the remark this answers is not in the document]
torkian: Good catch — fixed in 2eccd5b. `agent_node` now returns only `{"messages": [retry_response]}`; the skipped first attempt was already shown to the model in the local `retry_messages` (ephemeral), so dropping it from state preserves a valid sequence whether the retry is a tool call or another clarification. Added a regression test asserting no two consecutive AIMessages are ever sent to the LLM across a forced-search run (it fails against the old `[response, retry_response]` return).

Auditing the same invariant across the rest of the graph turned up a separate, pre-existing instance of the same bug class on the **skip-clarification** path: the skip branch emitted `AIMessage(complete)` without persisting the user's reply, and the unconditional re-entry into `agent_node` (budget exhausted) appended a second completion — three adjacent assistant turns that 400 the planner's `ainvoke` when `enable_plan_approval=True`. Fixed in 832a1fa (persist the skip reply as a HumanMessage + suppress the duplicate completion on re-entry), with regression tests covering both the persisted history and the planner sequence. 150 clarifier tests pass; `ruff` clean.

src/aiq_agent/agents/clarifier/agent.py
```diff
@@ -576,8 +656,17 @@ async def ask_clarification(state: ClarifierAgentState):
...
+                # AIMessage(clarification) in history; without an interleaving
+                # human message the two assistant turns would be adjacent, which
+                # the OpenAI/Anthropic APIs reject. (The duplicate completion on
+                # graph re-entry is suppressed by the guard in agent_node.)
                 return {
-                    "messages": [AIMessage(content=complete_response.model_dump_json())],
+                    "messages": [
+                        HumanMessage(content=user_reply),
+                        AIMessage(content=complete_response.model_dump_json()),
+                    ],
```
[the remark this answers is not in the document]
torkian: Fixed in 8dda5c0 — blank/whitespace skip replies now substitute a non-empty `[skipped clarification]` sentinel, and a regression test asserts the planner never receives an empty HumanMessage.

tests/aiq_agent/agents/clarifier/test_agent.py
```diff
@@ -872,3 +875,472 @@ def test_init_with_planner_llm(self, mock_llm_provider):
...
+        assert not offenders, f"persisted history has consecutive assistant messages at {offenders}: {msgs}"
+        # The skip reply must be persisted as a HumanMessage.
+        assert any(isinstance(m, HumanMessage) and m.content == "skip" for m in msgs), (
+            "skip reply was not persisted as a HumanMessage"
+        )
+        # Exactly one completion AIMessage should terminate the dialog (no duplicate).
+        completion_ais = [
+            m for m in msgs if isinstance(m, AIMessage) and ClarifierAgent._has_tool_invocations([m]) is False
+        ]
+        assert len(completion_ais) >= 1
```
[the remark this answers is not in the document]
torkian: Fixed in 8dda5c0 — assertion now matches only the terminal completion AIMessage (via `_is_complete`) and requires exactly one.

torkian:
Did a deep pre-merge review pass to clear remaining edge cases in one go. Pushed in `6fef9e3`:

**Fixed (in scope of this change):**
- **Search guard scoping** — `_searched_since_last_user_turn` replaces the whole-history `_has_tool_invocations` check, so a tool call from an *earlier* conversation turn no longer suppresses the search-before-clarify nudge for a fresh query. Unit + behavioral regression tests added.
- **Exhausted-completion hardening** — `agent_node`'s budget-exhausted branch never emits two consecutive assistant messages now; when the last turn is a non-complete `AIMessage` at exhaustion it interleaves a sentinel user turn before the completion (previously it could append an adjacent completion that 400s the planner under plan approval). Regression test added.
- **Sentinel constant** — promoted the blank-skip placeholder to `SKIPPED_CLARIFICATION_SENTINEL` (fixes the casing drift with the log literal), reused by the interleave above.
- **Test polish** — tightened two planner `call_count` asserts to `== 1`, made the fallback-query assertion require both tokens (`and`), removed an unused dict, and added docstrings to the graph-node functions and new fixtures/helpers (ruff + format clean, 157 tests pass).

**Intentionally out of scope** (pre-existing, not touched by this PR — flagging rather than expanding scope):
- `ClarifierAgent.max_turns` vs `state.max_turns` precedence in `run()`.
- `_get_fallback_clarification` using the latest `HumanMessage` rather than the original query on multi-turn fallbacks.
- Blank plan-approval replies counting as empty feedback in `plan_preview_node`.

Happy to file follow-ups for those if you'd like them addressed.

src/aiq_agent/agents/clarifier/agent.py
```diff
@@ -548,9 +629,53 @@ async def agent_node(state: ClarifierAgentState):
                 messages.append(HumanMessage(content=JSON_REMINDER_AFTER_TOOLS))
 
             response = await bound_llm.ainvoke(messages)
+
```
cdgamarose-nv, line 643: This is an unnecessarily long comment, even longer than the code additions themselves. I recommend reducing them to ~3 lines.
 The same rationale is duplicated in the agent_node docstring (~L585) and the skip-path comment (~L704)

src/aiq_agent/agents/clarifier/agent.py
```diff
@@ -548,9 +629,53 @@ async def agent_node(state: ClarifierAgentState):
...
+            # regardless of whether it is a tool call or another clarification.
+            if (
+                self.tools
+                and state.iteration == 0
+                and not self._searched_since_last_user_turn(state.messages)
+                and not getattr(response, "tool_calls", None)
+                and self._is_needed(response.content)
+            ):
+                logger.info("Clarifier: model skipped search before clarifying; injecting guidance and retrying once")
+                retry_messages = messages + [response, HumanMessage(content=FORCE_SEARCH_GUIDANCE)]
```
cdgamarose-nv: The nudge is injected as a trailing HumanMessage. Some models weight the last human turn as "the real query"; a SystemMessage reads more naturally as scaffolding and sidesteps that. Was HumanMessage deliberate?

tests/aiq_agent/agents/clarifier/test_agent.py
```diff
@@ -872,3 +875,670 @@ def test_init_with_planner_llm(self, mock_llm_provider):
...
+
+    @pytest.fixture
+    def mock_llm_provider(self, mock_llm):
+        """Create a mock LLM provider returning the mock LLM."""
+        provider = MagicMock(spec=LLMProvider)
+        provider.get = MagicMock(return_value=mock_llm)
+        return provider
+
+    @staticmethod
+    def _adjacent_assistant_pairs(messages: list[BaseMessage]) -> list[tuple[int, int]]:
```
cdgamarose-nv: _adjacent_assistant_pairs is defined three times — inline in test_forced_retry_does_not_emit_consecutive_assistant_messages (~L1236), and as a staticmethod in TestClarifierSkipMessageOrdering (~L1291) and TestClarifierReviewRegressions (~L1456). Hoist it to one module-level helper and reuse.

cdgamarose-nv (changes requested):

src/aiq_agent/agents/clarifier/agent.py
torkian: Trimmed to ~6 lines in 82fdab9 — dropped the duplicated rationale.

src/aiq_agent/agents/clarifier/agent.py
torkian: Good call — switched the nudge to a trailing SystemMessage in 82fdab9. HumanMessage wasn't deliberate; SystemMessage reads as scaffolding and avoids the model treating it as the real query. Still ephemeral (never persisted to state).

tests/aiq_agent/agents/clarifier/test_agent.py
torkian: Done in 82fdab9 — hoisted _adjacent_assistant_pairs to one module-level helper and removed the three duplicates.

src/aiq_agent/agents/clarifier/agent.py
```diff
@@ -525,9 +582,33 @@ def _build_graph(self) -> CompiledStateGraph:
...
+                    content=ClarificationResponse(
+                        needs_clarification=False, clarification_question=None
+                    ).model_dump_json()
+                )
+                if isinstance(last_message, AIMessage):
+                    # The last turn is a non-complete assistant message (e.g. an
+                    # unanswered clarification at exhaustion). Interleave a
+                    # sentinel user turn so the completion is not adjacent to it.
+                    return {"messages": [HumanMessage(content=SKIPPED_CLARIFICATION_SENTINEL), complete]}
+                return {"messages": [complete]}
```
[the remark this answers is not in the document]
torkian: Good catch — fixed in 62151d8. If the last message at exhaustion is an AIMessage with pending tool_calls, agent_node now returns no new message so decide_route routes the call to the tools node first; the tool result then re-enters and completion proceeds validly. Added a regression test.

torkian:
Thanks for the review @cdgamarose-nv — addressed all of it:

- **Nudge as SystemMessage** (`82fdab9`): switched the search-before-clarify nudge from a trailing HumanMessage to a SystemMessage so the model reads it as scaffolding, not the real query. It stays ephemeral (never persisted to state). HumanMessage wasn't deliberate — good call.
- **Trimmed the oversized comment** (`82fdab9`): the retry-block rationale is down from ~26 lines to ~6.
- **Deduped the test helper** (`82fdab9`): hoisted `_adjacent_assistant_pairs` to a single module-level function.
- **Pending tool call at exhaustion** (`62151d8`): the exhausted-completion branch no longer stacks a completion on top of an AIMessage with pending tool_calls (which would leave the call unresolved); it defers so the tools node resolves it first. Added a regression test.

158 clarifier tests pass; ruff check + format clean.

src/aiq_agent/agents/clarifier/agent.py
```diff
@@ -548,9 +629,34 @@ async def agent_node(state: ClarifierAgentState):
...
+            if (
+                self.tools
+                and state.iteration == 0
+                and not self._searched_since_last_user_turn(state.messages)
+                and not getattr(response, "tool_calls", None)
+                and self._is_needed(response.content)
+            ):
+                logger.info("Clarifier: model skipped search before clarifying; injecting guidance and retrying once")
+                retry_messages = messages + [response, SystemMessage(content=FORCE_SEARCH_GUIDANCE)]
+                retry_response = await bound_llm.ainvoke(retry_messages)
```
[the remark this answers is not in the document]
torkian: Fixed in 1f6ae72 — the guidance is now folded into the leading system prompt for the retry instead of appended as a trailing SystemMessage, so the message list stays valid for wrappers that only accept a leading system message. Tightened the test to require exactly one system message, in the leading position, carrying the guidance.

torkian:
@cdgamarose-nv branch is up to date with `develop` and all review feedback is addressed (per-thread replies above) — ready for re-review whenever convenient. 158 clarifier tests pass; lint clean.

src/aiq_agent/agents/clarifier/agent.py
cdgamarose-nv: you don't need to mention the issue number.

cdgamarose-nv (approved):
lgtm

PR #261, opened 2026-06-01
feat: add nimble_web_search data source

wildcard:
## Summary

- Adds `sources/nimble_web_search`, a NAT data source that wraps `langchain-nimble`'s `NimbleSearchRetriever`, mirroring the existing `exa_web_search` and `tavily_web_search` packages (typed config, stub-on-missing-key, retries, content truncation, XML-tagged output).
- Exposes `lite` / `fast` / `deep` `search_depth`, typed as a `Literal` so invalid values fail at config-parse time. `lite` is the default — metadata-only, token-cheap, works on any account. `fast` is enterprise-tier and surfaces a clear 403 entitlement message on non-enterprise keys.
- Adds a typed `focus` mode (default `general`, validated `Literal`), `country` / `locale` regional controls, and an optional `max_content_length` per-result cap.
- Wires the plugin into the workspace, `deploy/Dockerfile`, and `scripts/setup.sh` so it installs in dev, Docker, and container builds. The Docker layer also installs Nimble's runtime deps (`langchain-nimble`, `nimble-python`, lockfile-pinned) so `_type: nimble_web_search` resolves in built images, where the `--no-dev` sync would otherwise omit them.
- Documentation across the config reference, extending guides, installation, quick-start, deployment (docker-build, docker-compose, kubernetes), FAQ, and troubleshooting (with 401 and 403-enterprise rows).

#### DCO sign-off for the squash commit

Signed-off-by: Kobi Kadosh <kobi.kadosh@gmail.com>

## Motivation

AI-Q ships Tavily- and Exa-backed web search today. [Nimble](https://nimbleway.com/) provides web search and content extraction for AI agents; this adds it as a first-class alternative with the same ergonomics and config surface — handy for users who already have a Nimble subscription, prefer its regional coverage, or want to test across multiple search backends. The default provider is unchanged (Tavily stays the documented default).

It wraps the official `langchain-nimble` package (maintained by Nimble) rather than calling the HTTP API directly, so retry, auth, and response normalization come from the upstream integration — the same rationale as the Exa source (#181).

## Configuration

```yaml
functions:
  web_search_tool:
    _type: nimble_web_search
    max_results: 5
    search_depth: lite      # lite (default) | fast (enterprise) | deep
    focus: general          # general (default) | news | location | shopping | geo | social
    country: US
    locale: en
```

```bash
NIMBLE_API_KEY=...   # or set api_key: in the YAML
```

## How it works

A real `lite` query with `NIMBLE_API_KEY` set, trimmed:

[...]

PR #262, opened 2026-06-02
feat(nimble_web_search): optional domain/recency/date filters + output_format + result cap

wildcard:
## Summary

Adds optional search-refinement controls to the `nimble_web_search` data source:
domain include/exclude lists, recency and date-range filters, and output-format
selection, plus a hard upper bound on result count. Every new field is optional
and defaults off, so existing workflow configs are unaffected.

> **Stacked on #261.** This builds on the base `nimble_web_search` data source in
> #261. Its net-new change is the final commit (`feat(nimble_web_search): add
> optional domain/recency/date filters, output_format, and result cap`). Until
> #261 merges, the diff here also shows #261's commits; it collapses to just the
> enrichment once #261 lands. **Please review #261 first.** Marked draft until then.

## What changed

New optional `NimbleWebSearchToolConfig` fields (forwarded to
`NimbleSearchRetriever` only when set):

- `include_domains` / `exclude_domains` — restrict or exclude result domains.
  Both reject an empty list (`min_length=1`); use `None` (the default) to disable.
- `time_range` — recency window (`hour`/`day`/`week`/`month`/`year`).
- `start_date` / `end_date` — published-date range (`YYYY-MM-DD` or `YYYY`).
- `output_format` — `plain_text` / `markdown` (default) / `simplified_html`.

Behavior:

- `max_results` is now enforced as a hard upper bound. The API may return more
  than requested, so results are sliced to `max_results` for predictable breadth
  consistent with the other web-search data sources. The API may still return
  fewer (a property of any SERP backend), which is surfaced as-is.
- `include_answer` remains intentionally unexposed.

Validation: `time_range` and `output_format` are constrained `Literal[...]`
enums, so invalid values are rejected at config-load. `start_date` / `end_date`
are free-form `str` (accepting `YYYY-MM-DD` or `YYYY`); they are forwarded to the
SDK as-is rather than format-validated in the provider, matching the
`langchain-nimble` contract.

## Tests

- `uv run pytest sources/nimble_web_search -q` → 44 passed (12 new: passthrough
  per field, enum-rejection, empty-domain-list rejection, the upper-bound-cap
  behavior, and a filter-active-with-empty-results path).
- Co-run with a sibling provider → 58 passed (no test-name collision).
- `ruff check` / `ruff format --check` / `uv lock --check` clean.

## Compatibility

All new fields default to `None` (or the prior implicit default for
`output_format`), so configs that set none of them parse unchanged and the data
source is a strict superset of the base.

One intentional behavior change applies even when no new field is set: because
`max_results` is now a hard upper bound, a workflow that previously received
*more* than `max_results` results (the API soft-caps above the requested count)
will now receive exactly `max_results`. Consumers that already sliced downstream
are unaffected; the rendered `<Document>` block format is otherwise unchanged.


AjayThorve:
see https://github.com/NVIDIA-AI-Blueprints/aiq/pull/261#issuecomment-4869756060

AjayThorve:
feel free to re-open if it's a possibility to comply with the above requirements

wildcard:
Hi @AjayThorve — with the base provider merged (#261, thank you!), this follow-up is refreshed and ready: I've rebuilt the branch on `release/2.2`, so the diff reduces to exactly the optional search filters (`include_domains` / `exclude_domains` / `time_range` / `start_date` / `end_date`), `output_format`, and the `max_results` hard cap — re-applied on top of the merged provider (the cap now applies after the resolvable-URL filter, so the result budget is spent on usable results).

Verified on the refreshed branch:

- `uv run pytest sources/nimble_web_search/tests` → 54 passed, 1 skipped (the 12 filter/cap tests join the merged suite; the key-gated live test passes against the real API in ~2s).
- Live spot-check of the new surface with the CI credential: `include_domains: ["docs.nvidia.com"]` returned results exclusively from that domain.
- `ruff check` / `ruff format --check` / `uv lock --check` / pre-commit hooks all clean; every commit signed off.

Since the PR is closed I can't reopen it or push the refresh from my side — could you reopen it (or let me know your preferred route)? Once it's open I'll retarget the base to `release/2.2` and push. One note for transparency: the refresh re-creates the original commits with DCO sign-offs (content-identical trees) before layering the release/2.2 merge, so the DCO check comes up green from the start.

PR #265, opened 2026-06-08
Add AI-Q product-level skill evals (push-triggered, complementary to NVSkills CI)

sarathm-nvidia:
## Summary
Adds a product-level skill-eval harness: runs AI-Q Agent Skills through Harbor
trials against a live AI-Q server the job builds and deploys, scoring
deterministic checks. Complements the existing NVSkills CI with product-side
coverage of skill behavior across the profiles we control.

## Included
- `.github/skill-eval/` — eval coordinator, `aiq-research` adapter, deterministic
  verifier, docs.
- `.github/workflows/skills-eval.yml` — `generate-datasets` (spec validation, no
  credentials) and `harbor-eval` (full trial) on the self-hosted `aiq-eval`
  runner. Push-triggered to match the repo CI model (`ci.yml`/`ui.yml`): runs on
  `develop`, `release/**`, and `pull-request/<N>` mirror branches, filtered to
  skill/eval/deploy paths. `workflow_dispatch` retained for manual runs.
- `skills/aiq-research/evals/basic-product.json` — first product spec (smoke
  profile: health + agent listing; no model-credit `/chat` calls).
- A one-line `aiq-research` connectivity adjustment so the in-container eval can
  reach the deployed server (kept as its own commit; happy to discuss in review).

## Behavior
- Incremental on PRs (only the skills changed vs `develop`); full sweep on
  `develop`/`release`/manual dispatch.
- `harbor-eval` is single-flight so the shared runner isn't double-booked.
- Distinct trigger from NVSkills CI — the two pipelines don't collide.

## Validation
Full Harbor trial on the self-hosted runner: build -> deploy -> health -> trials
-> teardown — reward 1.0, 0 exceptions.

### DIFF
--- a/.github/skill-eval/AGENTS.md
+++ b/.github/skill-eval/AGENTS.md
@@ -1,0 +1,24 @@
+# AI-Q Skills Eval Agent
+
+You are the AI-Q skills-eval coordinator. Your job is to run product-level evals for Agent Skills under `skills/`.
+
+## Workflow
+
+1. Find changed skill directories or, in manual mode, all product-eval specs under `skills/<skill>/evals/*-product.json`.
+2. Require each spec to define `skills`, `resources.platforms`, `env`, and `expects`.
+3. Require an adapter at `.github/skill-eval/adapters/<skill>/generate.py`.
+4. Generate datasets with the adapter. Do not edit skill source during an eval run.
+5. If Harbor execution is enabled, run one trial for each generated platform/mode dataset.
+6. Report deterministic verifier results and trace paths.
+
+## Hard Rules
+
+- Do not print API keys or copied environment values.
+- Do not invent deployment assumptions. If a live AI-Q server is required, use `AIQ_SERVER_URL` and the `aiq-deploy` skill.
+- Do not bake VSS profile names, VSS ports, Brev secure-link rules, or video-service checks into AI-Q adapters.
+- If an adapter or spec is missing, fail clearly with the missing path.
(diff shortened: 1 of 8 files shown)


sarathm-nvidia:
re-post /ok to test 02dbdae57a8736f83657c676e11f1202a962b226

AjayThorve:
@sarathm-nvidia quick question: should we add an explicit post-Harbor pass/fail gate here?

I noticed `aiq_checks.py` writes a fractional reward but exits `0`, and `skills_eval_agent.py` only propagates `harbor run`'s process exit. My understanding is that Harbor records and prints reward metrics, but does not fail the CLI just because reward is below `1.0`. If that is right, a future run with reward `0.5` and no verifier crash could still produce a successful GitHub check.

If `Skills Eval` being green is meant to mean the product smoke eval passed, should this workflow fail when any trial has an exception or any reward is below a configured threshold, defaulting to `1.0`? Or is there already a Harbor threshold mechanism we should be relying on here?

AjayThorve (approved):
approved with one caveat asked above, about the threshold, not sure if that's the standard.

sarathm-nvidia:
@AjayThorve - your comment is valid, earlier it was only presenting the result for dev to decide. Now I added an explicit gate to fail if the reward falls short of threshold.  (default 1.0, env-overridable).
Reference run (had set 1.1 to create failure scenario) : https://github.com/NVIDIA-AI-Blueprints/aiq/actions/runs/27225896407/attempts/2
Also resolved greptile suggestions

AjayThorve (approved):
LGTM!

PR #267, opened 2026-06-10
enhancement: make deep researcher customizable, flexible and less token expensive

cdgamarose-nv:
## Summary

  Adds structured-output deep research with advisory source routing, batched researcher workers, writer-only final synthesis, and new news / prediction-market sources.

  ## Changes

  - Refactors deep researcher into factory-built source-router, planner, researcher-batch, and writer stages.
  - Adds strict structured contracts for `SourceRoutingPlan`, `ResearchPlan`, `ResearchNotes`, answer strategy, research queries, and evidence judgments.
  - Adds `run_research_batch`, source-tool batching/throttling, compact verified-source lists from ResearchNotes, and stricter `/shared/output.md` final-answer extraction.
  - Adds source routing config via renamed `config_domain_routing_and_skills.yml`, new `deep_research_domain_catalog.yml`, `source_router_llm`, `writer_llm`, routing enablement, and
  concurrency limits.
  - Adds DuckDuckGo News and Polymarket prediction-market NAT source packages, workspace dependencies, tests, docs, and lockfile updates.
  - Adds per-agent DeepAgents skill routing plus nested research/synthesis skills for table analysis, forecast analysis, lightweight calculation, long-form reports, and prediction
  reports.
  - Updates prompts for routed planning, structured researcher notes, batch research, writer synthesis, and stricter citation rules.
  - Improves citation verification for footnote/source-location citation syntax and missing Sources sections backed by registry sources.
  - Updates async job construction for the new deep researcher constructor and writer skill/sandbox scoping.
  - Makes data-source filtering exact-case by source ID.
  - Updates docs, benchmark/default configs, AIQ research script headers, tokenomics/API/test license headers, and Exa headers.
  - Adds/updates tests for contracts, factory wiring, source routing, batching/concurrency, runtime skill routing, citation normalization, source plugins, async runner behavior, and
  config handling.

## Summary by CodeRabbit

* **New Features**
  * Added DuckDuckGo news search and Polymarket prediction market search tools.
* **Improvements**
  * Switched Nemotron usage from nano to super across configurations.
  * Upgraded deep research with source routing, specialized subagents, and improved role-based LLM wiring.
  * Enhanced citation normalization and more robust source tracking/verification.
  * Improved deep research concurrency and source-tool batching; removed the deep-research loop limit.
* **Documentation**
  * Updated deep research configuration references/examples for new LLM roles and parameters.


configs/config_domain_routing_and_skills.yml
cdgamarose-nv, line 1: Made this a deep research agent only config because the current planner messes up the ability to make different outputs since it comes up with section headers and so on

AjayThorve:
@coderabbitai can you review

src/aiq_agent/agents/deep_researcher/agent.py
```diff
@@ -411,103 +207,16 @@ async def run(self, state: DeepResearchAgentState) -> DeepResearchAgentState:
...
-
-            if result is None and last_error is not None:
-                raise last_error
-
-            final_message = "Research failed to produce a report."
-            if result and result.get("messages"):
-                final_message = self._extract_report_content(result["messages"])
+            final_message = self._extract_final_markdown(result)
+            if final_message is None:
+                raise ValueError("writer-agent did not produce a final Markdown answer")
```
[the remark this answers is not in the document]
cdgamarose-nv: will be handled in a separate PR, created an [issue](https://github.com/NVIDIA-AI-Blueprints/aiq/issues/270) for it.

src/aiq_agent/common/citation_verification.py
```diff
@@ -700,8 +722,29 @@ def verify_citations(report_text: str, registry: SourceRegistry) -> CitationVeri
...
+        if not _INLINE_CITATION_RE.search(report_text):
+            logger.warning("[CitationVerify] No references section found in report; skipping")
+            return CitationVerificationResult(verified_report=report_text)
+
+        cited_numbers = sorted({int(match.group(1)) for match in _INLINE_CITATION_RE.finditer(report_text)})
+        reference_lines = [
+            line
+            for i in cited_numbers
+            if 1 <= i <= len(all_sources)
+            if (line := _format_registry_reference(i, all_sources[i - 1]))
```
AjayThorve: **[Critical] Citations can be attributed to the wrong source.**

This missing-`## Sources`-section fallback maps each inline `[N]` to `all_sources()[N - 1]` — i.e. it assumes `[N]` is the N-th source in *registration* order. But the writer numbers citations from the **compact, renumbered** verified-source subset (default `mode="compact"` in `tools/source_registry.py`, renumbered via `loop.index`), which is a *filtered* subset of `all_sources()`.

So when the writer omits the Sources section and its compact numbering differs from registration order, the synthesized references attach the **wrong URL** to a claim — and they then pass as `valid_citations`, silently laundering an incorrect citation.

_Note: the new `test_missing_references_section_*` fixtures use a case where compact order == registration order, so they validate the buggy assumption rather than catch it._

src/aiq_agent/agents/deep_researcher/agent.py
```diff
@@ -1,399 +1,195 @@
...
         self.verbose = verbose
         self.callbacks = callbacks or []
 
-        if self.verbose:
-            logger.info("Tools configured: %d", len(self.tools))
         if config is not None:
             skills = skills or getattr(config, "skills", None)
             sandbox = sandbox if sandbox is not None else getattr(config, "sandbox", None)
-        self.deepagents_runtime = DeepAgentsRuntime(skills=skills, sandbox=sandbox, job_id=job_id)
+            enable_source_router = getattr(config, "enable_source_router", enable_source_router)
```
AjayThorve, line 104: **[High] New LLM roles + routing/concurrency config are silently ignored on the async (Dask) job path** — the path the server API actually serves `deep_researcher` on.

When constructed with `config=` (how `aiq_api/jobs/runner.py::_create_agent_instance` builds the agent — it passes only `llm_provider`, `tools`, `verbose`, `callbacks`, `config=fn_config`, `job_id`), this block reads **only** `skills`/`sandbox`/`enable_source_router` from config. `domain_catalog_path` (L108) and `max_research_concurrency`/`max_concurrent_source_tool_calls`/`max_source_tool_batch_size` come from constructor params the runner never passes, so they default.

And `runner.py` configures only `LLMRole.ORCHESTRATOR`/`PLANNER`/`RESEARCHER` — it never configures `LLMRole.ROUTER` (`source_router_llm`) nor `LLMRole.REPORT_WRITER` (`writer_llm`). Both new roles this PR adds are wired in `register.py` (the sync/CLI path) but not on the async path.

Net effect for async jobs: `source_router_llm`, `writer_llm`, `domain_catalog_path`, and the concurrency limits are **all inert** — e.g. `configs/config_domain_routing_and_skills.yml` sets all of these, but async jobs fall back to the default router/writer LLMs and the trivial single-domain default catalog.

src/aiq_agent/agents/deep_researcher/tools/research.py
```diff
@@ -0,0 +1,218 @@
...
+            )
+        notes, errors = await _run_research_queries(
+            queries=queries,
+            researcher_runnable=researcher_runnable,
+            runtime=runtime,
+            callbacks=callbacks,
+            max_concurrency=max_research_concurrency,
+        )
+        if errors:
+            raise RuntimeError(
```
AjayThorve, line 222: **[High] One failed worker discards all successful research notes.**

On any worker error this raises **before** `register_research_note_sources(notes)` (L209) and `_persist_research_notes(...)` (L210) run, so the successful `ResearchNotes` are neither persisted to `/shared/` nor registered into the compact source set — the orchestrator has to redo the whole batch. The orchestrator prompt says to 'resubmit only the invalid queries,' but nothing is kept for it to build on.

tests/aiq_agent/agents/deep_researcher/test_agent.py
```diff
@@ -349,14 +964,18 @@ def test_modal_backend_creates_sandbox_lazily(self):
...
         from aiq_agent.agents.deep_researcher.deepagents_runtime import _create_sandbox_backend
 
+        class NotFoundError(Exception):
+            pass
+
+        NotFoundError.__module__ = "modal.exception"
+
         first_modal_backend = MagicMock()
-        first_modal_backend.execute.side_effect = modal.exception.NotFoundError("gone")
+        first_modal_backend.execute.side_effect = NotFoundError("gone")
```
AjayThorve: **[High] This test fails wherever `modal` is installed — i.e. CI and any dev machine after `uv sync`.**

This PR changed the test from raising a real `modal.exception.NotFoundError` (as on `develop`) to a fake class with `__module__` spoofed to `"modal.exception"`. But `_is_modal_not_found_error` (`deepagents_runtime.py:509`) does `isinstance(exc, modal.exception.NotFoundError)` whenever `modal` is importable — and `modal` is a hard dep via `langchain-modal==0.0.5`. The fake isn't an instance of the real class, so detection returns `False`, the retry never fires, and the assertions below fail.

Verified locally: the changed suites give **422 passed, 1 failed** (this test); a direct repro shows the fake -> `False`, a real `NotFoundError` -> `True`.

configs/deep_research_domain_catalog.yml
```diff
@@ -0,0 +1,63 @@
...
+      - web_search
+    fallback_source_ids:
+      - web_search
+    is_default: true
+
+  - domain_id: scholarly_technical
+    domain_name: Scholarly and Technical Research
+    description: Scientific, engineering, medical, benchmark, method, paper, and standards-heavy questions.
+    preferred_source_ids:
+      - academic_search
```
AjayThorve: **[Medium] `academic_search` is not a registered source id.**

It matches no `data_source_registry` entry, so `source_routing._filter_available` silently drops it into `unavailable_source_ids` and the `scholarly_technical` domain loses that route — even though this file's header states 'Source IDs must match entries in the data_source_registry.'

AjayThorve (commented):
Reviewed this PR against the head commit and ran the changed test suites locally (**422 passed, 1 failed**). The refactor is solid overall — clean decomposition into source-router/planner/researcher-batch/writer, good unit coverage, and `max_loops` removed cleanly. Flagging the critical/high issues I'm most confident in as inline comments.

One more, not inline (the stale lines aren't in this diff): the `config_skills.yml` → `config_domain_routing_and_skills.yml` rename left runnable `--config_file configs/config_skills.yml` commands in `docs/source/examples/skills-sandbox/index.md` and a reference in `docs/source/examples/index.md` that now 404.

configs/domain_catalogs/deep_research_domain_catalog.yml
```diff
@@ -0,0 +1,63 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Domain catalog for deep research source routing.
+#
+# This file is intentionally data-only. Add, remove, or revise domains here
```
AjayThorve, line 18: **[Design / non-blocking] `configs/` mixes runnable workflow configs with a data asset.**

`configs/` otherwise holds *runnable workflow entry points* (`config_*.yml`, passed to `nat serve --config_file`). This file is a different kind of thing — a data asset consumed *by* a config via `domain_catalog_path:` (the header here even says 'intentionally data-only'). Sitting flat next to the runnable configs invites `nat serve --config_file configs/deep_research_domain_catalog.yml`, which fails cryptically (no `general:`/`workflow:`). Conceptually it's a sibling of `prompts/`/`skills/` (which live with the agent), not of the workflow configs.

Two related rough edges: `domain_catalog_path` is resolved CWD-relative (`DomainCatalogRegistry.from_path` -> `Path(path).expanduser()`), so it only resolves when launched from the repo root; and an unset path yields only the trivial single-domain default (`source_routing._default_general_research_config`), so multi-domain routing is effectively opt-in-by-path.

src/aiq_agent/agents/deep_researcher/skills/research-sandbox/data-table-analysis/SKILL.md
```diff
@@ -1,17 +1,12 @@
 ---
 name: data-table-analysis
```
AjayThorve, line 2: **[Design / non-blocking] Two different "skills" systems now share the same name.**

This directory (`src/aiq_agent/agents/deep_researcher/skills/`, new in this PR) holds **DeepAgents runtime skills** — markdown loaded *into the deep researcher's own sub-agents* via `SkillsConfig`/`skill_sources_for`, addressed by the virtual `/skills/...` mount.

The repo also has a top-level `skills/` (`aiq-research`, `aiq-deploy`) which are **Anthropic Agent Skills** — harness-facing packages for *external* callers (Claude Code / Codex / OpenCode), with `permissions`/`license`/`evals`/signature, and symlinked into `.agents/skills` and `.claude/skills`.

Same word, same `SKILL.md` filename, same `name`/`description` frontmatter — but **opposite audiences** (internal sub-agents vs external harness). Anyone grepping `SKILL.md` or searching 'skills' will conflate them, and this PR is what introduces the second concept into the repo.

configs/config_domain_routing_and_skills.yml
```diff
@@ -186,46 +191,28 @@ functions:
...
+    writer_llm: gpt_oss_llm
     # tools: omitted -> inherits all from data_source_registry
     exclude_tools:                     # Remove basic variant; deep uses advanced_web_search_tool
       - web_search_tool
-    max_loops: 2
     verbose: true
+    domain_catalog_path: configs/deep_research_domain_catalog.yml
     skills:
       enabled: true
+      agent_sources:
```
AjayThorve, line 198: **[Design / non-blocking] Reference skills by group name, not virtual `/skills/...` paths.**

The `/skills/research-sandbox/` and `/skills/synthesis/` values read like absolute OS paths, but they're actually a **virtual mount** (`BUILTIN_SKILL_SOURCE = "/skills/"`) resolved to the on-disk `src/aiq_agent/agents/deep_researcher/skills/`. That's confusing on three counts:

- It isn't a real filesystem path.
- It is **not** the repo-root `skills/` dir (the Anthropic Agent Skills — `aiq-research`/`aiq-deploy`), a different system that happens to share the name and the `SKILL.md` convention.
- The same path is repeated in `sandbox_required_sources`, so the guardrail has to be kept in sync with `agent_sources` by hand — and a path typo silently disables skills (`agent_sources.get(name)` -> `None`, no error).

Since `SkillsConfig` is AIQ's own Pydantic model (it can expose any schema and translate to DeepAgents' path-based API internally), referencing **named groups** removes all three problems while **keeping the guardrail explicit in config**:

```yaml
skills:
  enabled: true
  researcher: [research-sandbox]               # group names, not /skills/... paths
  writer-agent: [synthesis]
  sandbox_required_sources: [research-sandbox] # guardrail stays in config, by name
sandbox:
  provider: modal
```

- Group names are validated against the known set, so a typo errors loudly instead of silently desyncing.
- No virtual-path / Agent-Skills ambiguity.
- The sandbox guardrail stays an explicit, operator-owned line — just keyed by a stable group name instead of a duplicated path (optionally rename it to `sandbox_required` since the values are now names, not sources).

_Internals stay path-based — the `/skills/` mount + `SKILL.md` discovery are DeepAgents conventions; this is only about the config surface AIQ controls._

sources/duckduckgo_news_search/src/register.py
```diff
@@ -0,0 +1,155 @@
...
+            with DDGS() as ddgs:
+                return list(ddgs.news(query, **search_kwargs))
+
+        for attempt in range(tool_config.max_retries):
+            try:
+                results = await asyncio.wait_for(asyncio.to_thread(_search), timeout=tool_config.timeout)
+                if results:
+                    return "\n\n---\n\n".join(_format_news_result(result) for result in results)
+                return "News search returned no results"
+            except Exception as exc:  # noqa: BLE001 - source APIs can raise transport-specific exceptions
```
AjayThorve, line 146: **[Bug — confirmed at runtime] DuckDuckGo source treats "no results" as a hard error, so it fails on ~every call.**

`ddgs.news()` (L132) raises `ddgs.exceptions.DDGSException("No results found.")` when a query simply has no news hits — a benign/empty condition, not a transport error. This broad `except` catches it and returns `"Error: News search failed"` (L147) for those queries.

In a live server run this produced **18 identical `DDGSException: No results found.` tracebacks** and surfaced `Error: News search failed` on essentially every `duckduckgo_news_search_tool` call. Secondary effect: the CitationRegistry then captures the failed call as a source (`[CitationRegistry] Captured 1 source(s) from duckduckgo_news_search_tool`), polluting the source set with a non-result.

src/aiq_agent/agents/deep_researcher/prompts/researcher.j2
```diff
@@ -40,42 +36,47 @@ Your output will be used to write a long-form report. Produce **in-depth, detail
...
+- When the primary source tool supports batch input and subqueries are still useful after assessing the main query, pass the remaining related concrete search angles as one ordered list.
+- Do not run every possible source angle; stop once findings are adequately grounded.
+- Start broad, then narrow based on gaps identified.
+- Stop when you have comprehensive coverage for the assigned query and target components.
+- Do not answer from model memory. Every ResearchFinding must be grounded in results returned by preferred source tools, fallback source tools, or user-provided/internal documents surfaced by available tools.
+- Use `fallback_tools` only for corroboration or gaps after using the preferred tools. If the listed source tools do not provide support for a claim, record a ResearchGap instead of filling the missing fact from prior knowledge.
 
 ## Handling Failures
-- Do NOT get stuck retrying - proceed with available information and write the evidence file
+- Do NOT get stuck retrying - proceed with available information and return the structured ResearchNotes
```
AjayThorve, line 62: **[Bug — observed at runtime] The researcher runs tool names and a non-existent `return_json` as shell commands in the sandbox.**

With the research-sandbox skill's `execute` (bash) tool available alongside the function tools, the researcher (nemotron-3-nano) repeatedly calls `execute` with `{command: 'knowledge_search'}` and `{command: 'return_json'}` -> `bash: ... command not found [exit 127]`. It conflates (a) function/tool calls, (b) sandbox bash execution, and (c) *returning* its `ResearchNotes` — inventing a `return_json` command to emit the structured response this prompt asks for.

Observed repeatedly in a live run — it even self-diagnoses ('we attempted to run it as a shell command') and then repeats the mistake — burning turns and feeding exactly the 'stuck retrying' loop this line warns against. Smaller models seem especially prone to it when an `execute` tool and function tools are both in scope.

configs/domain_catalogs/deep_research_domain_catalog.yml
cdgamarose-nv: What is your recommendation here? To move to prompts/skills directory?

src/aiq_agent/agents/deep_researcher/skills/research-sandbox/data-table-analysis/SKILL.md
cdgamarose-nv: Created new issue: https://github.com/NVIDIA-AI-Blueprints/aiq/issues/275

configs/config_domain_routing_and_skills.yml
cdgamarose-nv: Created issue: https://github.com/NVIDIA-AI-Blueprints/aiq/issues/275

src/aiq_agent/common/citation_verification.py
```diff
@@ -586,6 +586,10 @@ def _parse_knowledge_layer(content: str, tool_name: str) -> list[SourceEntry]:
 )
 
 _CITATION_LINE_RE = re.compile(r"^\s*[-*]?\s*\[(\d+)\]\s*(.+)$", re.MULTILINE)
```
AjayThorve, line 602: **[Bug] Citation verifier only parses `[N]` reference lines — a `N.` numbered-list Sources section yields zero citations.**

After `_REFERENCE_SECTION_RE` matches a `## Sources` / `## References` heading, `_CITATION_LINE_RE` (`^\s*[-*]?\s*\[(\d+)\]\s*(.+)$`) is the only parser for the reference lines, and it requires the `[N]` bracket form (optionally bulleted). If the writer renders its Sources section as a **markdown ordered list** — `1. ...`, `2. ...` (a very common, natural LLM output) — `_CITATION_LINE_RE` matches **none** of those lines, so the verifier parses zero references and ends with an empty `valid_citations` set — *even when the section is present and well-formed and the inline `[N]` citations are correct and map to the right sources*.

Because the heading matched, this takes the strict-parse path rather than the missing-section synthesis fallback, so there's no recovery. `_normalize_citation_syntax` normalizes bracket *variants* (`【】`, `[^n]` footnotes, `[n †...]` source-locations) but does not convert `N.` numbered references into `[N]`. Net effect: a complete, correctly-cited report is scored as having no citations purely due to reference-line formatting (`N.` vs `[N]`).

src/aiq_agent/agents/deep_researcher/agent.py
```diff
@@ -522,10 +231,7 @@ async def run(self, state: DeepResearchAgentState) -> DeepResearchAgentState:
                     )
                 final_message = verification.verified_report
                 if not verification.valid_citations:
-                    logger.warning(
-                        "Deep researcher produced no valid citations after verification; "
-                        "returning sanitized report without fabricating references."
-                    )
+                    raise ValueError("writer-agent output contains no valid citations")
```
AjayThorve: **[Bug] Empty `valid_citations` is now a hard failure that discards the finished report (behavior change vs `develop`).**

When `verify_citations` returns an empty `valid_citations` set, this raises `ValueError("writer-agent output contains no valid citations")`, failing the whole job — the already-generated final report is thrown away.

On `develop` this same condition was a **logged warning** that still returned the report; this PR escalates it to a hard `raise`. Combined with the verifier's strict reference-line parsing (see the companion comment on `_CITATION_LINE_RE`), any report whose citation/Sources formatting the verifier can't parse — including complete, correctly-cited ones — now produces a job failure instead of a degraded-but-returned result. Given the repo's documented 'citation verification can be overly aggressive' caveat, making this a hard gate converts verifier false-negatives directly into lost reports.

configs/domain_catalogs/deep_research_domain_catalog.yml
AjayThorve: Not `prompts/skills` — that dir is for code-shipped assets, but this catalog is meant to be operator-editable (the header says "edit here without changing Python source"), so burying it in the package fights that intent. It's also not a runnable workflow config, so flat in configs/ is misleading too.

What I'd suggest: move it to a subdir like `configs/catalogs/` so top-level configs/*.yml stays "things you can nat serve"

src/aiq_agent/agents/deep_researcher/prompts/researcher.j2
cdgamarose-nv: Will add instructions to prompt for now as well as switch to nemotron super. But if this is observed going forward, we should probably remove `execute` as an allowed tool call from subagents that have source tools and create a separate `coding-subagent`.

configs/domain_catalogs/deep_research_domain_catalog.yml
cdgamarose-nv: Okay, the previous comment said prompts/skills. So it was misleading.

AjayThorve (dismissed):
GTG, great job on the PR

docs/source/architecture/agents/deep-researcher.md
```diff
@@ -149,9 +150,10 @@ functions:
   deep_research_agent:
     _type: deep_research_agent
     orchestrator_llm: nemotron_llm
+    source_router_llm: nemotron_nano_llm
```
AjayThorve: this is super now right?

docs/notebooks/0_Getting_Started_with_AIQ.ipynb
```diff
@@ -374,7 +374,6 @@
     "    orchestrator_llm: gpt_oss_llm\n",
     "    planner_llm: gpt_oss_llm\n",
     "    researcher_llm: nemotron_nano_llm  # replace with nemotron_super_llm if available\n",
```
AjayThorve: can you do a global rename to update nanos to supers

AjayThorve (approved):
LGTM

salimassili62-afk:
This is a big improvement structurally, especially separating source routing, researcher batches, and writer-only synthesis.

One thing I’d watch closely is the runtime envelope now that deep research has more moving parts and the loop limit was removed. Source routing + batched researchers + nested skills can create a lot of model/tool calls unless there is a central pre-call guard.

A pattern like this would make the pipeline safer:

```ts id="p8t4w2"
const decision = canRunResearchStep({
  jobId,
  stage, // router | planner | researcher | writer
  agentId,
  model,
  sourceId,
  batchSize,
  concurrency,
  estimatedTokens,
  remainingBudget,
});
```

Before each LLM/tool batch, I’d check:

* max researcher workers per job
* max source-tool calls per source and per job
* max nested skill depth
* known model pricing for `source_router_llm` and `writer_llm`
* per-stage token/cost budget
* repeated equivalent research queries
* fail-closed behavior when source routing config is missing or ambiguous

The structured contracts and citation verification are strong additions. I’d just make sure the new router/researcher/writer split also has a deterministic budget/concurrency boundary before calls execute, not only observability after the job has already expanded.

PR #271, opened 2026-06-14
Add report follow-up jobs and chat routing

AjayThorve:
#### Overview

Adds **report-aware follow-up** to the deep-research blueprint. After a report job completes, a user can ask about it, make cosmetic edits, or start delta research that reuses the prior report — without forcing every follow-up through a dedicated "report mode." A chat **router** picks a semantic route from the message plus an optional `active_report_job_id`; the active report is treated as *context*, never as a command.

**Modes**
- **report ask** — inline, bounded LLM answer from the parent report only. No tools, no live research, no child artifact.
- **report cosmetic edit** — an internal `report_rewriter` async child job for mechanical/aesthetic edits that do not need new evidence. The parent report stays immutable.
- **report delta research** — the existing deep researcher, seeded with parent-report context, for fresh evidence, deeper analysis, or a new analytical perspective on the same report topic.
- **standalone research** — normal shallow/deep research when the request is unrelated to the active report or asks for a separate report.

**Intent routing examples**
- `what are the risks in this report?` → `report_ask`
- `make this shorter` → `report_cosmetic_edit`
- `format the key takeaways as bullets` → `report_cosmetic_edit`
- `rewrite this report from a player-performance POV` → `report_delta_research`
- `redo this with newer evidence on 2026 logistics` → `report_delta_research`
- `write a separate report on player performance trends across 2014, 2018, and 2022` → `standalone_research`

**Key pieces**
- One new internal agent — `report_rewriter` — registered with `public=False` (hidden from `GET /agents`; direct `/submit` of internal-only agents is rejected, and the gate is also enforced at the submission boundary).
- A durable parent-report **context resolver** that reconstructs the report + sources from job output/events, **authorizes the caller before any read**, and seeds `/shared/*` files into child runs.
- New endpoint `POST /v1/jobs/async/job/{job_id}/report/edit` (per-job ownership auth); `GET .../report` extended with `parent_job_id` / `interaction_action` / `result_kind`.
- Chat (`POST /chat` / WebSocket) accepts `active_report_job_id`; the UI forwards the latest completed report job id and renders the report-edit child job through the existing report-streaming path.
- DeepAgents runtime normalizes seeded `/shared/*` parent-context files so delta research can read `/shared/original_report.md` and `/shared/source_summary.md` reliably.

> Rebased onto `develop` (which now includes #267); this branch is a standalone PR. The most recent commits harden the feature for production: the chat path works under the default `REQUIRE_AUTH=false`, a caller-supplied `job_id` can no longer delete another job's state, submission failures roll back cleanly, the UI consumes the report-edit response, and report-delta research can reuse seeded parent context without creating `/shared/shared` or malformed file records.

#### Validation

All commands run from the repo root unless noted.

**Lint / format / unit tests (backend)**
```bash
uv run ruff check .
uv run ruff format --check .
uv run pytest                                  # full suite
# or the suites this PR touches:
uv run pytest frontends/aiq_api/tests/ \
              tests/aiq_agent/agents/chat_researcher/ \
              tests/aiq_agent/agents/report_rewriter/ \
              tests/aiq_agent/agents/deep_researcher/test_deepagents_runtime.py \
              tests/aiq_agent/jobs/test_runner.py
```

**UI**
```bash
cd frontends/ui
npm run lint && npm run type-check && npm run test:ci
# focused: npx vitest run src/features/chat/hooks/use-websocket-chat.spec.ts \
#                         src/adapters/api/websocket-client.spec.ts
```

**End-to-end (manual)** — report follow-up needs async jobs, so bring up the full stack (PostgreSQL job store + embedded Dask scheduler/worker + web), e.g. via `deploy/compose` or `scripts/start_server_in_debug_mode.sh` with `deploy/.env`, then exercise the surfaces below.

**What to test**
1. **Internal-agent gating** — `GET /v1/jobs/async/agents` does *not* list `report_rewriter`; `POST /v1/jobs/async/submit` with `agent_type=report_rewriter` returns `400 Agent type is internal-only: report_rewriter`.
2. **HTTP report edit** — `POST /v1/jobs/async/job/{id}/report/edit` on a completed report → a `report_rewriter` child job; `GET .../report` on the child returns a revised report plus `parent_job_id`, `interaction_action="edit"`, `result_kind="report"`. The parent report is unchanged.
3. **Chat routing with `active_report_job_id`** — verify each semantic route:
   - **report ask**: `What are the top three takeaways from this report?`; `Where does the report say the evidence is weak or incomplete?`
   - **report cosmetic edit**: `Make this report shorter while preserving the sources.`; `Format the key takeaways as bullets.`; `Remove the one-table comparison section and keep the rest unchanged.`
   - **report delta research**: `Rewrite this report from a player-performance POV.`; `Redo this report with newer evidence on 2026 logistics and host-city operations.`; `Add a section on fan travel emissions for 2026.`
   - **standalone research**: `Write a separate report on player performance trends across the 2014, 2018, and 2022 World Cups.`; `Research the economics of Olympic host cities since 2000.`
4. **Parent-context seeding for delta research** — delta research should read `/shared/original_report.md` and `/shared/source_summary.md` successfully, not see `/shared/shared/`, and not fail with `string indices must be integers` or `'str' object has no attribute 'get'` from filesystem tools.
5. **Authorization** — works for anonymous callers under `REQUIRE_AUTH=false`; under `REQUIRE_AUTH=true`, a non-owner is rejected with `404` before any report content is read.
6. **Robustness / edge cases** — a colliding caller-supplied `job_id` returns `409` and does **not** delete the existing job; whitespace-only `input` returns `422`; report ask/edit degrade to a chat message (not an opaque workflow error) if context resolution fails.

**Evidence** — backend suites (`frontends/aiq_api/tests` + `tests/aiq_agent/...`) and UI specs pass; `ruff` and `tsc` clean. Verified live against a Postgres + Dask stack with `deploy/.env`: internal agent hidden + `/submit` rejected (400); HTTP and chat report-edit produce a revised child report with correct lineage; chat report-ask answers from the report only; `job_id` collision → 409 with the victim preserved; blank input → 422; ownership-mismatch → 404 under `REQUIRE_AUTH=true`; report-delta routing kicks off deep research with seeded parent context.

Read in this order — security-sensitive paths first:

1. **`frontends/aiq_api/src/aiq_api/jobs/report_context.py`** — durable report/source reconstruction and `resolve_authorized_report_context()`, which **authorizes the caller before any read** and seeds `/shared/*` for child runs. The core security boundary.
2. **`frontends/aiq_api/src/aiq_api/routes/jobs.py`** — the `report/edit` endpoint, the `public` agent filter on `/agents` + `/submit`, the new `JobReportResponse` fields, and the request validators (blank-input → 422, `job_id` collision → 409).
3. **`frontends/aiq_api/src/aiq_api/jobs/submit.py`** — `submit_agent_job()` ownership recording, the `JobIdConflictError`/`InternalAgentError` gates, and the rollback that only deletes state this submission created.
4. **`src/aiq_agent/agents/chat_researcher/nodes/intent_classifier.py`** + **`agent.py`** + **`register.py`** — the semantic route classifier (`report_ask` / `report_cosmetic_edit` / `report_delta_research` / `standalone_research`), the bounded tool-free report-ask path, and how each entry point resolves the principal (same contract as the HTTP routes).
5. **`src/aiq_agent/agents/deep_researcher/deepagents_runtime.py`** — route-aware `/shared/*` file seeding for delta research.
6. **`src/aiq_agent/agents/report_rewriter/`** — the single new internal agent (a bounded, tool-less single-LLM rewrite).
7. **`frontends/ui/src/features/chat/hooks/use-websocket-chat.ts`** + `adapters/api/websocket-client.ts` — forwards `active_report_job_id` and routes the report-edit child job through the existing report-streaming path.

Tests mirror these: `test_report_context.py`, `test_report_edit.py`, `test_submit_collision.py`, `test_submit_internal_agent.py`, `test_job_access.py`, `test_agent_registry_visibility.py`, `test_intent_classifier.py`, `test_deepagents_runtime.py` (backend), and `use-websocket-chat.spec.ts` (UI).

#### Related Issues

- Relates to #267 (deep-researcher structured output / customization — merged into `develop`).

## Release Notes

* **New Features**
  * Report editing and follow-up workflows to revise completed reports
  * Report Q&A capability to ask questions about existing reports
  * Conversation-scoped job tracking for multi-turn interactions
  * Agent visibility controls restricting certain agents from public submission

* **Improvements**
  * Enhanced input validation and error responses for job submissions
  * Expanded REST API documentation for report operations
  * Better job ID conflict detection and error handling

### DIFF
--- a/frontends/aiq_api/src/aiq_api/jobs/report_context.py
+++ b/frontends/aiq_api/src/aiq_api/jobs/report_context.py
@@ -203,17 +203,24 @@ async def resolve_report_context(job: Any, db_url: str, parent_job_id: str) -> R
     if not report:
         raise HTTPException(409, f"Parent job has no durable report: {parent_job_id}")
 
-    if events is None:
-        # The report is already in hand (from job output); the event log is only
-        # needed to enrich sources. A transient fetch failure here must not abort
-        # report follow-up — fall back to report-only source extraction.
-        try:
-            events = await EventStore.get_events_async(db_url, parent_job_id, 0, _EVENT_SCAN_LIMIT)
-        except Exception:
-            events = []
-    event_sources = _sources_from_events(events)
     report_sources = _extract_sources_from_report_markdown(report)
-    sources = _dedupe_sources([*event_sources, *report_sources])
+
+    if events is None:
+        # The report came from job output. Its own ## Sources section is authoritative for
+        # follow-up context, so only pay the (potentially large) event scan when the report
+        # carries no inline sources. A transient fetch failure must not abort follow-up.
+        if report_sources:
+            sources = report_sources
+        else:
+            try:
+                events = await EventStore.get_events_async(db_url, parent_job_id, 0, _EVENT_SCAN_LIMIT)
+            except Exception:
+                events = []
+            sources = _dedupe_sources([*_sources_from_events(events), *report_sources])
+    else:
+        # The event log was already fetched to reconstruct the report; reuse it for sources.
+        sources = _dedupe_sources([*_sources_from_events(events), *report_sources])
+
     return ReportContext(
         parent_job_id=parent_job_id,
         report_markdown=report,
@@ -238,18 +245,35 @@ def report_context_from_markdown(report_markdown: str, parent_job_id: str = "in-
     )
 
 
+# Cache JobStore instances by (scheduler_address, db_url). JobStore eagerly builds a SQLAlchemy
+# AsyncEngine with its own connection pool, so constructing one per request both wastes work and
+# leaks pooled connections. Engines are designed to be long-lived and shared, and JobStore's
+# sessions are task-scoped (async_scoped_session), so a process-wide cache is safe.
+_JOB_STORE_CACHE: dict[tuple[str, str], Any] = {}
+
+
+def _get_job_store(scheduler_address: str, db_url: str) -> Any:
+    from nat.front_ends.fastapi.async_jobs.job_store import JobStore
+
+    key = (scheduler_address, db_url)
+    store = _JOB_STORE_CACHE.get(key)
+    if store is None:
+        store = JobStore(scheduler_address=scheduler_address, db_url=db_url)
+        _JOB_STORE_CACHE[key] = store
+    return store
+
+
 async def resolve_authorized_report_context(parent_job_id: str, principal: Principal) -> ReportContext:
     """Authorize and reconstruct a parent report context using configured async job storage."""
 
     from nat.front_ends.fastapi.async_jobs.job_store import JobStatus
-    from nat.front_ends.fastapi.async_jobs.job_store import JobStore
 
     scheduler_address = os.environ.get("NAT_DASK_SCHEDULER_ADDRESS")
     if not scheduler_address:
         raise HTTPException(503, "Async job storage is not configured")
 
     db_url = os.environ.get("NAT_JOB_STORE_DB_URL", "sqlite:///./data/jobs.db")
-    job_store = JobStore(scheduler_address=scheduler_address, db_url=db_url)
+    job_store = _get_job_store(scheduler_address, db_url)
     job = await authorize_job_access(job_store, db_url, parent_job_id, principal)
     if getattr(job, "status", None) != JobStatus.SUCCESS.value:
         raise HTTPException(409, f"Parent job is not complete: {parent_job_id}")
@@ -257,7 +281,7 @@ async def resolve_authorized_report_context(parent_job_id: str, principal: Princ
 
 
 def to_initial_files(context: ReportContext, instruction: str | None = None) -> dict[str, str]:
-    """Convert report context into PR 267 DeepAgents virtual filesystem seed files."""
+    """Convert report context into DeepAgents virtual filesystem seed files."""
 
     # Exclude report_markdown from the JSON context: the full report is already seeded
     # verbatim into /shared/original_report.md, so embedding it again here would roughly
--- a/frontends/aiq_api/src/aiq_api/jobs/report_context.py
+++ b/frontends/aiq_api/src/aiq_api/jobs/report_context.py
@@ -87,11 +87,6 @@ def _report_from_events(events: list[dict[str, Any]]) -> str | None:
     return final_report or fallback
 
 
-async def _extract_report_from_events(db_url: str, job_id: str) -> str | None:
-    events = await EventStore.get_events_async(db_url, job_id, 0, _EVENT_SCAN_LIMIT)
-    return _report_from_events(events)
-
-
 def _is_url(value: str | None) -> bool:
     return bool(value and value.lower().startswith(("http://", "https://")))
 
@@ -141,12 +136,7 @@ def _sources_from_events(events: list[dict[str, Any]]) -> list[ReportContextSour
                 tool_name=str(data.get("tool_name") or "parent_report"),
             )
         )
-    return _dedupe_sources(sources)
-
-
-async def _extract_sources_from_events(db_url: str, job_id: str) -> list[ReportContextSource]:
-    events = await EventStore.get_events_async(db_url, job_id, 0, _EVENT_SCAN_LIMIT)
-    return _sources_from_events(events)
+    return sources
 
 
 def _sources_section(report_markdown: str) -> str:
@@ -176,7 +166,7 @@ def _extract_sources_from_report_markdown(report_markdown: str) -> list[ReportCo
             sources.append(ReportContextSource(url=url_match.group(0).rstrip(_URL_TRIM_CHARS), title=ref_text))
         else:
             sources.append(ReportContextSource(citation_key=ref_text, title=ref_text))
-    return _dedupe_sources(sources)
+    return sources
 
 
 def _source_summary_markdown(sources: list[ReportContextSource]) -> str:
@@ -210,7 +200,7 @@ async def resolve_report_context(job: Any, db_url: str, parent_job_id: str) -> R
         # follow-up context, so only pay the (potentially large) event scan when the report
         # carries no inline sources. A transient fetch failure must not abort follow-up.
         if report_sources:
-            sources = report_sources
+            sources = _dedupe_sources(report_sources)
         else:
             try:
                 events = await EventStore.get_events_async(db_url, parent_job_id, 0, _EVENT_SCAN_LIMIT)
@@ -236,7 +226,7 @@ def report_context_from_markdown(report_markdown: str, parent_job_id: str = "in-
     current session rather than as a durable async job. Sources are reconstructed from the report's
     own ``## Sources`` section via the same parser used for job-backed reports.
     """
-    sources = _extract_sources_from_report_markdown(report_markdown)
+    sources = _dedupe_sources(_extract_sources_from_report_markdown(report_markdown))
     return ReportContext(
         parent_job_id=parent_job_id,
         report_markdown=report_markdown,
--- a/frontends/aiq_api/src/aiq_api/jobs/report_context.py
+++ b/frontends/aiq_api/src/aiq_api/jobs/report_context.py
@@ -204,7 +204,13 @@ async def resolve_report_context(job: Any, db_url: str, parent_job_id: str) -> R
         raise HTTPException(409, f"Parent job has no durable report: {parent_job_id}")
 
     if events is None:
-        events = await EventStore.get_events_async(db_url, parent_job_id, 0, _EVENT_SCAN_LIMIT)
+        # The report is already in hand (from job output); the event log is only
+        # needed to enrich sources. A transient fetch failure here must not abort
+        # report follow-up — fall back to report-only source extraction.
+        try:
+            events = await EventStore.get_events_async(db_url, parent_job_id, 0, _EVENT_SCAN_LIMIT)
+        except Exception:
+            events = []
     event_sources = _sources_from_events(events)
     report_sources = _extract_sources_from_report_markdown(report)
     sources = _dedupe_sources([*event_sources, *report_sources])
--- a/frontends/aiq_api/src/aiq_api/jobs/report_context.py
+++ b/frontends/aiq_api/src/aiq_api/jobs/report_context.py
@@ -237,9 +237,12 @@ async def resolve_authorized_report_context(parent_job_id: str, principal: Princ
 def to_initial_files(context: ReportContext, instruction: str | None = None) -> dict[str, str]:
     """Convert report context into PR 267 DeepAgents virtual filesystem seed files."""
 
+    # Exclude report_markdown from the JSON context: the full report is already seeded
+    # verbatim into /shared/original_report.md, so embedding it again here would roughly
+    # double the report tokens carried into the rewrite/research prompt for no benefit.
     files = {
         "/shared/original_report.md": context.report_markdown,
-        "/shared/parent_report_context.json": context.model_dump_json(indent=2),
+        "/shared/parent_report_context.json": context.model_dump_json(indent=2, exclude={"report_markdown"}),
         "/shared/source_summary.md": context.source_summary_markdown,
     }
     if instruction is not None:
--- a/frontends/aiq_api/src/aiq_api/jobs/report_context.py
+++ b/frontends/aiq_api/src/aiq_api/jobs/report_context.py
@@ -222,6 +222,22 @@ async def resolve_report_context(job: Any, db_url: str, parent_job_id: str) -> R
     )
 
 
+def report_context_from_markdown(report_markdown: str, parent_job_id: str = "in-session") -> ReportContext:
+    """Build report context directly from report markdown — no job store, auth, or scheduler.
+
+    Used for report follow-up in the synchronous CLI, where the report was produced inline in the
+    current session rather than as a durable async job. Sources are reconstructed from the report's
+    own ``## Sources`` section via the same parser used for job-backed reports.
+    """
+    sources = _extract_sources_from_report_markdown(report_markdown)
+    return ReportContext(
+        parent_job_id=parent_job_id,
+        report_markdown=report_markdown,
+        source_summary_markdown=_source_summary_markdown(sources),
+        sources=sources,
+    )
+
+
 async def resolve_authorized_report_context(parent_job_id: str, principal: Principal) -> ReportContext:
     """Authorize and reconstruct a parent report context using configured async job storage."""
 
--- a/frontends/aiq_api/src/aiq_api/jobs/report_context.py
+++ b/frontends/aiq_api/src/aiq_api/jobs/report_context.py
@@ -69,8 +69,7 @@ def _event_data(event: dict[str, Any]) -> dict[str, Any]:
     return data if isinstance(data, dict) else {}
 
 
-async def _extract_report_from_events(db_url: str, job_id: str) -> str | None:
-    events = await EventStore.get_events_async(db_url, job_id, 0, _EVENT_SCAN_LIMIT)
+def _report_from_events(events: list[dict[str, Any]]) -> str | None:
     fallback: str | None = None
     final_report: str | None = None
     for event in events:
@@ -86,6 +85,11 @@ async def _extract_report_from_events(db_url: str, job_id: str) -> str | None:
         if data.get("output_category") == "final_report":
             final_report = content.strip()
     return final_report or fallback
+
+
+async def _extract_report_from_events(db_url: str, job_id: str) -> str | None:
+    events = await EventStore.get_events_async(db_url, job_id, 0, _EVENT_SCAN_LIMIT)
+    return _report_from_events(events)
 
 
 def _is_url(value: str | None) -> bool:
@@ -113,8 +117,7 @@ def _dedupe_sources(sources: list[ReportContextSource]) -> list[ReportContextSou
     return deduped
 
 
-async def _extract_sources_from_events(db_url: str, job_id: str) -> list[ReportContextSource]:
-    events = await EventStore.get_events_async(db_url, job_id, 0, _EVENT_SCAN_LIMIT)
+def _sources_from_events(events: list[dict[str, Any]]) -> list[ReportContextSource]:
     sources: list[ReportContextSource] = []
     for event in events:
         if event.get("type") != "artifact.update":
@@ -139,6 +142,11 @@ async def _extract_sources_from_events(db_url: str, job_id: str) -> list[ReportC
             )
         )
     return _dedupe_sources(sources)
+
+
+async def _extract_sources_from_events(db_url: str, job_id: str) -> list[ReportContextSource]:
+    events = await EventStore.get_events_async(db_url, job_id, 0, _EVENT_SCAN_LIMIT)
+    return _sources_from_events(events)
 
 
 def _sources_section(report_markdown: str) -> str:
@@ -185,11 +193,19 @@ def _source_summary_markdown(sources: list[ReportContextSource]) -> str:
 async def resolve_report_context(job: Any, db_url: str, parent_job_id: str) -> ReportContext:
     """Build report context from a previously authorized parent job."""
 
-    report = _extract_report_from_job_output(job) or await _extract_report_from_events(db_url, parent_job_id)
+    report = _extract_report_from_job_output(job)
+    # Durable events are only needed when the report isn't already in job output,
+    # or to reconstruct sources. Fetch the (potentially large) event log at most once.
+    events: list[dict[str, Any]] | None = None
+    if not report:
+        events = await EventStore.get_events_async(db_url, parent_job_id, 0, _EVENT_SCAN_LIMIT)
+        report = _report_from_events(events)
     if not report:
         raise HTTPException(409, f"Parent job has no durable report: {parent_job_id}")
 
-    event_sources = await _extract_sources_from_events(db_url, parent_job_id)
+    if events is None:
+        events = await EventStore.get_events_async(db_url, parent_job_id, 0, _EVENT_SCAN_LIMIT)
+    event_sources = _sources_from_events(events)
     report_sources = _extract_sources_from_report_markdown(report)
     sources = _dedupe_sources([*event_sources, *report_sources])
     return ReportContext(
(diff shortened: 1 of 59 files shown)


frontends/ui/src/features/chat/hooks/use-websocket-chat.ts
```diff
@@ -593,22 +616,28 @@ export const useWebSocketChat = (options: UseWebSocketChatOptions = {}): UseWebS
...
-        // Check for deep research escalation signal
-        // Backend sends: "Deep research job submitted. Job ID: {uuid}"
+        // Check for an async-job escalation signal. Two backend paths produce a
+        // pollable child job whose report is fetched over the same SSE/report
+        // endpoints, so both escalate the same way:
+        //   - "Deep research job submitted. Job ID: {uuid}"   (deep_researcher)
+        //   - "Report edit job submitted. Job ID: {uuid}"     (report_rewriter)
         const deepResearchMatch = content?.match(
-          /Deep research job submitted\. Job ID: ([a-f0-9-]+)/i
+          /(Deep research|Report edit) job submitted\. Job ID: ([a-f0-9-]+)/i
```
cdgamarose-nv: Worried about using regex to detect report edit/deep research escalation. If there is a change in punctuation/case it will disable job tracking. Can we replace this with structured metadata/an event type?

frontends/aiq_api/src/aiq_api/jobs/report_context.py
```diff
@@ -0,0 +1,265 @@
...
+    db_url = os.environ.get("NAT_JOB_STORE_DB_URL", "sqlite:///./data/jobs.db")
+    job_store = JobStore(scheduler_address=scheduler_address, db_url=db_url)
+    job = await authorize_job_access(job_store, db_url, parent_job_id, principal)
+    if getattr(job, "status", None) != JobStatus.SUCCESS.value:
+        raise HTTPException(409, f"Parent job is not complete: {parent_job_id}")
+    return await resolve_report_context(job, db_url, parent_job_id)
+
+
+def to_initial_files(context: ReportContext, instruction: str | None = None) -> dict[str, str]:
+    """Convert report context into PR 267 DeepAgents virtual filesystem seed files."""
```
cdgamarose-nv: Can we remove PR number?

frontends/aiq_api/src/aiq_api/jobs/report_context.py
```diff
@@ -0,0 +1,265 @@
...
+        # The report is already in hand (from job output); the event log is only
+        # needed to enrich sources. A transient fetch failure here must not abort
+        # report follow-up — fall back to report-only source extraction.
+        try:
+            events = await EventStore.get_events_async(db_url, parent_job_id, 0, _EVENT_SCAN_LIMIT)
+        except Exception:
+            events = []
+    event_sources = _sources_from_events(events)
+    report_sources = _extract_sources_from_report_markdown(report)
+    sources = _dedupe_sources([*event_sources, *report_sources])
```
cdgamarose-nv: Sources seem to be deduped in `_sources_from_events()`, `_extract_sources_from_report_markdown()`, and again after merging in `resolve_report_context()`. Is the final merge-level dedupe enough and can we avoid the earlier ones on large events?

frontends/aiq_api/src/aiq_api/jobs/submit.py
```diff
@@ -159,6 +185,11 @@ async def submit_agent_job(
     # Get agent configuration from registry
     agent_config = get_agent_config(agent_type)
 
+    # Enforce the internal-agent gate at the submission boundary (defense in depth):
+    # internal-only agents may only be launched by trusted callers that opt in.
+    if not agent_config.public and not allow_internal:
+        raise InternalAgentError(f"Agent type is internal-only and cannot be submitted directly: {agent_type}")
```
cdgamarose-nv, line 201: There are two different internal-agent rejection messages: the HTTP route returns `Agent type is internal-only: ...`, while the submission helper raises `Agent type is internal-only
  and cannot be submitted directly: ...`. Can we unify these strings?

frontends/aiq_api/src/aiq_api/routes/jobs.py
```diff
@@ -493,6 +558,67 @@ async def submit_job(
...
+    @app.post(
+        "/v1/jobs/async/job/{job_id}/report/edit",
+        response_model=ReportEditResponse,
+        tags=["async jobs"],
+        summary="Create a revised report from a completed report job",
+        description=(
+            "Authorize access to a completed parent report, reconstruct durable report context, "
+            "and submit an internal child job that emits a full revised report."
+        ),
+        responses={
```
cdgamarose-nv, line 644: The `/report/edit` route can raise `HTTPException(500, "Failed to submit report edit job")`, but the FastAPI route `responses` metadata only lists 404, 409, and 503.

frontends/aiq_api/src/aiq_api/jobs/report_context.py
```diff
@@ -0,0 +1,265 @@
...
+        report = _report_from_events(events)
+    if not report:
+        raise HTTPException(409, f"Parent job has no durable report: {parent_job_id}")
+
+    if events is None:
+        # The report is already in hand (from job output); the event log is only
+        # needed to enrich sources. A transient fetch failure here must not abort
+        # report follow-up — fall back to report-only source extraction.
+        try:
+            events = await EventStore.get_events_async(db_url, parent_job_id, 0, _EVENT_SCAN_LIMIT)
```
cdgamarose-nv: `resolve_report_context()` still scans up to 10k EventStore events even when the report is already present in `job.output`, to enrich sources. This runs on report ask/edit/
  delta follow-up paths, so large jobs can add foreground latency. Could we avoid the EventStore scan?

frontends/aiq_api/src/aiq_api/jobs/report_context.py
```diff
@@ -0,0 +1,265 @@
...
+
+    from nat.front_ends.fastapi.async_jobs.job_store import JobStatus
+    from nat.front_ends.fastapi.async_jobs.job_store import JobStore
+
+    scheduler_address = os.environ.get("NAT_DASK_SCHEDULER_ADDRESS")
+    if not scheduler_address:
+        raise HTTPException(503, "Async job storage is not configured")
+
+    db_url = os.environ.get("NAT_JOB_STORE_DB_URL", "sqlite:///./data/jobs.db")
+    job_store = JobStore(scheduler_address=scheduler_address, db_url=db_url)
```
cdgamarose-nv: Is it intentional to construct a new `JobStore` from env vars on each chat follow-up.

cdgamarose-nv:
Both report ask and report edit inline the full parent report into an LLM prompt. Since we don't have token budget or truncation, I'm worried that large completed reports can exceed the model context window or crowd out the user’s actual question/edit instruction. Not sure what the right thing to do here is. For first pass, we could probably add a size/token guard here? But wondering if it would be possible to use a simple agent with filesystem tools rather than individual LLM calls?

AjayThorve:
> Both report ask and report edit inline the full parent report into an LLM prompt. Since we don't have token budget or truncation, I'm worried that large completed reports can exceed the model context window or crowd out the user’s actual question/edit instruction. Not sure what the right thing to do here is. For first pass, we could probably add a size/token guard here? But wondering if it would be possible to use a simple agent with filesystem tools rather than individual LLM calls?

@cdgamarose-nv Good callout — I dug into this and it breaks down by path:

1. Delta (re-research) runs on the deepagents stack, which already has summarization middleware wired (create_summarization_middleware in deep_researcher/factory.py), so its context gets compacted mid-loop — that path is covered.

2. Ask and edit are single-shot LLM calls, so they do inline the full report. A summarizer can't help there: summarization happens between agent steps, but a single oversized prompt fails before there's any step boundary to compact at.


**On the actual risk**: both run on large context llm (gpt-oss-120b, large context window), and if a report ever did exceed the window, report_ask_node and report_edit_node each wrap the call in try/except and degrade to a "please try again" message — no crash, no data loss. So today it's a graceful tail-case, not a correctness bug.

On the filesystem-tools idea — that's the right long-term shape, and the deep_researcher already does exactly that (deepagents read_file/grep over the report seeded into its virtual FS). The subtlety: a lazy read_file only defers loading — once it returns, the whole report is in context. What actually bounds the window is selective reads (grep/offset) or truncation. So the proper version for ask is a small read/grep agent that pulls only the relevant sections — a real rearchitecture, and the synchronous CLI path has no shared FS so it'd need a temp-file shim.

My plan is to defer the hard guard for this PR (tail risk + graceful failure + large-context model) and track the read/grep-agent version as a follow-up

frontends/aiq_api/src/aiq_api/jobs/report_context.py
```diff
@@ -0,0 +1,305 @@
...
+        content = data.get("content")
+        if not isinstance(content, str) or not content.strip():
+            continue
+        fallback = content.strip()
+        if data.get("output_category") == "final_report":
+            final_report = content.strip()
+    return final_report or fallback
+
+
+async def _extract_report_from_events(db_url: str, job_id: str) -> str | None:
```
cdgamarose-nv: This function doesn't look like its being used anywhere. Can we remove?

frontends/aiq_api/src/aiq_api/jobs/report_context.py
```diff
@@ -0,0 +1,305 @@
...
+                citation_key=citation_key.strip() if isinstance(citation_key, str) and citation_key.strip() else None,
+                title=data.get("title") or event.get("name"),
+                source_type=str(data.get("source_type") or "parent_report"),
+                tool_name=str(data.get("tool_name") or "parent_report"),
+            )
+        )
+    return _dedupe_sources(sources)
+
+
+async def _extract_sources_from_events(db_url: str, job_id: str) -> list[ReportContextSource]:
```
cdgamarose-nv: This doesn't look like its being used anywhere either. Can we remove?

src/aiq_agent/agents/chat_researcher/prompts/intent_classification.j2
```diff
@@ -16,7 +16,30 @@ Generate a direct response to the user.
...
+- "summarize this report" -> route = "report_ask"
+- "make this shorter" -> route = "report_cosmetic_edit"
+- "change tone" -> route = "report_cosmetic_edit"
+- "format as bullets" -> route = "report_cosmetic_edit"
+- "rewrite this report from a player-performance POV" -> route = "report_delta_research"
+- "redo this with newer evidence" -> route = "report_delta_research"
+- "can we write a report on this from a supply-chain angle?" -> route = "report_delta_research"
+- "write a separate report on player performance" -> route = "standalone_research"
+
+Compatibility fields:
```
cdgamarose-nv: The prompt instructs the model to emit route, target, report_action, and use_parent_report_context - and then says the last three are determined by the route:

  The LLM is being asked to re-derive fields that Python derives deterministically from route via _route_to_fields. This is wasted token budget and increases the chance of
  inconsistent output (e.g., route="report_ask" with target="new_research").

cdgamarose-nv: for example, this was all the reasoning:
```

2026-06-26 11:45:13 - INFO     - aiq_agent.common.callbacks:192 - [Reasoning]
2026-06-26 11:45:13 - INFO     - aiq_agent.common.callbacks:196 - We need to edit the report to include more detail. The user asks to edit the report to include more detail. The context: active report is available. The user wants cosmetic edit: include more detail. So we need to output a JSON with route = report_cosmetic_edit, target = report, report_action = edit, etc. The output format is a JSON object with fields: intent, route, target, report_action, meta_response, research_depth, use_parent_report_context, route_reasoning.

We need to generate the JSON response only. The intent is "research"? Actually the user asks to edit the report to include more detail. That's a request to edit the report, which is a cosmetic edit. According to instructions: If intent is "research", we need to determine route. This is a request to edit the report, which is a cosmetic edit that does not require new evidence. So route = "report_cosmetic_edit". target = "report". report_action = "edit". meta_response should be null because intent is research? Actually meta_response is only filled if intent is meta. Since we are in research mode, meta_response = null. research_depth = null (since cosmetic edit). use_parent_report_context = false (as per schema for cosmetic edit). route_reasoning: one sentence explanation.

```

src/aiq_agent/agents/report_rewriter/agent.py
```diff
@@ -0,0 +1,148 @@
...
+        verbose: bool = False,
+        callbacks: list[Any] | None = None,
+        config: Any | None = None,
+        job_id: str | None = None,
+    ) -> None:
+        self.llm_provider = llm_provider
+        self.tools = list(tools or [])
+        self.verbose = verbose
+        self.callbacks = callbacks or []
+        self.config = config
```
cdgamarose-nv: config is never used.

frontends/aiq_api/src/aiq_api/jobs/report_context.py
```diff
@@ -0,0 +1,305 @@
...
+
+def _is_url(value: str | None) -> bool:
+    return bool(value and value.lower().startswith(("http://", "https://")))
+
+
+def _normalize_url_key(url: str) -> str:
+    return url.rstrip("/").lower()
+
+
+def _dedupe_sources(sources: list[ReportContextSource]) -> list[ReportContextSource]:
```
cdgamarose-nv, line 98: `_sources_from_events` calls `_dedupe_sources` at the end. `_extract_sources_from_report_markdown` also calls `_dedupe_sources`. Then `resolve_report_context` calls `_dedupe_sources` on the
  merged result. The first two seem like wasted work when both lists are immediately merged and re-deduped.

src/aiq_agent/agents/chat_researcher/nodes/intent_classifier.py
```diff
@@ -136,18 +142,58 @@ async def run(self, state: ChatResearcherState) -> dict[str, Any]:
...
+            route = _normalize_route(parsed.get("route"))
+            target = _normalize_target(parsed.get("target"))
+            report_action = _normalize_report_action(parsed.get("report_action"))
+            use_parent_report_context = _normalize_bool(parsed.get("use_parent_report_context")) and active_report
+
+            if intent == "meta":
+                target = "meta"
+                report_action = None
+                use_parent_report_context = False
+            elif route is not None:
```
cdgamarose-nv, line 165: `_route_to_fields` returning target="meta" is reachable only by a contradictory LLM output, and produces broken state when it fires
```

    if route == _ROUTE_META:
        return "meta", None, False, research_depth, depth_reasoning
```

    But _route_to_fields is only called when intent != "meta":
```
    if intent == "meta":
        target = "meta"; ...
    elif route is not None:
        target, ... = _route_to_fields(route=route, ...)
```

    If the model contradictorily emits intent="research" + route="meta", `_route_to_fields` returns target="meta". Then the second branch of the downstream update logic fires:
```
    elif target != "report":
        update["depth_decision"] = DepthDecision(...)
```
    "meta" != "report" is true, so a `depth_decision` is set, but no `meta_response` message is added and no research runs. The graph routes to shallow_research with no coherent answer.

cdgamarose-nv: Wondering if LLM should only own the decision for route and research_depth, and python should derive everything else from it deterministically

AjayThorve: Agree. This is the same ownership issue as the prompt/schema comment above: `route` should be the single semantic routing decision, and Python should derive the workflow fields from it.

I’ll change this so `route="meta"` is normalized to the meta path before `_route_to_fields()` is called, and `_route_to_fields()` no longer has a `meta` branch. That should make contradictory output like `intent="research", route="meta"` resolve to a coherent meta response instead of producing `target="meta"` plus a research depth decision.

frontends/aiq_api/src/aiq_api/jobs/report_context.py
AjayThorve: Fixed in `aea3375`: removed `_extract_report_from_events`; `resolve_report_context` now fetches events and calls `_report_from_events` directly.

frontends/aiq_api/src/aiq_api/jobs/report_context.py
AjayThorve: Fixed in `aea3375`: removed `_extract_sources_from_events`; the context resolver now calls `_sources_from_events` directly after it has already fetched events.

src/aiq_agent/agents/chat_researcher/prompts/intent_classification.j2
AjayThorve: Fixed in `aea3375`: the prompt schema now asks the LLM only for `intent`, `route`, `meta_response`, `research_depth`, and `route_reasoning`. `target`, `report_action`, and parent-context routing are derived in Python from `route`.

src/aiq_agent/agents/report_rewriter/agent.py
AjayThorve: Fixed in `aea3375`: removed the unused `config` parameter/storage from `ReportRewriterAgent`. The async runner now has a job-id-only constructor path so the rewriter still receives `job_id`.

frontends/aiq_api/src/aiq_api/jobs/report_context.py
AjayThorve: Fixed in `aea3375`: `_sources_from_events` and `_extract_sources_from_report_markdown` now return raw source lists. Dedupe happens at context boundaries in `resolve_report_context` and `report_context_from_markdown`, with tests for duplicate preservation and boundary dedupe.

src/aiq_agent/agents/chat_researcher/nodes/intent_classifier.py
AjayThorve: Fixed in `aea3375`: `route` is now authoritative. `route="meta"` is normalized to the meta path before `_route_to_fields`, and `_route_to_fields` no longer has a meta branch. Added a regression for `intent="research"` plus `route="meta"`.

tanleach:
on https://github.com/NVIDIA-AI-Blueprints/aiq/blob/523602dff7e9fd7c928e618dc30f6151ce4732be/src/aiq_agent/agents/report_rewriter/agent.py#L63 do you need to run through the citations verification/sanitization again?

src/aiq_agent/agents/report_rewriter/agent.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+def _source_text(value: Any) -> str | None:
+    if isinstance(value, str):
+        stripped = value.strip()
+        return stripped or None
+    return None
+
+
+def _source_entries_from_parent_context(parent_context: str) -> list[SourceEntry]:
```
cdgamarose-nv: Can we reuse `_seed_parent_sources` here for code consistencies?

src/aiq_agent/agents/report_rewriter/agent.py
```diff
@@ -0,0 +1,208 @@
...
+        self,
+        llm_provider: LLMProvider,
+        tools: Sequence[BaseTool] | None = None,
+        *,
+        verbose: bool = False,
+        callbacks: list[Any] | None = None,
+        job_id: str | None = None,
+    ) -> None:
+        self.llm_provider = llm_provider
+        self.tools = list(tools or [])
```
cdgamarose-nv: `ReportRewriter` isn't going to use any of these tools right? Do we need to pass and store them?

src/aiq_agent/agents/chat_researcher/utils.py
```diff
@@ -97,34 +106,57 @@ def _extract_text_from_message(message: Any) -> str | None:
 
 
 def _extract_query_from_text(text: str) -> tuple[str, list[str] | None]:
+    """Extract query text and data sources from a text payload.
```
cdgamarose-nv: I don't think this is used except by tests (docstring confirms it). Can we remove it?

src/aiq_agent/agents/chat_researcher/prompts/intent_classification.j2
```diff
@@ -16,18 +16,37 @@ Generate a direct response to the user.
...
 - **shallow**: Single main question, factual lookup, 2-3 tool calls, no complex comparison.
-- **deep**: Multi-faceted, explicit comparisons, trend analysis, strategy/roadmaps, or "comprehensive" requests.
+- **deep**: Explicit report-generation requests ("write/create/generate a report"), multi-faceted requests,
+  explicit comparisons, trend analysis, strategy/roadmaps, or "comprehensive" requests.
 - **Tie-break**: If unsure, choose "shallow".
 
 ### OUTPUT FORMAT
 You must respond ONLY with a JSON object using this schema:
 {
   "intent": "meta" | "research",
```
cdgamarose-nv, line 45: Wondering if we even need the intent field if meta is a value for route field anyway? Why add dual signal?
We can come back to this in a different PR though

src/aiq_agent/agents/chat_researcher/models/intent.py
cdgamarose-nv, line 33: Can we document the new fields?

cdgamarose-nv (commented):
Left some cosmetic comments. Everything worked functionally for me! Good to go in.

src/aiq_agent/agents/report_rewriter/agent.py
AjayThorve: Extracted source_entries_from_parent_context() in citation_verification and use it from both the deep-research seed path and report rewriter. _seed_parent_sources now owns only middleware registration.

src/aiq_agent/agents/report_rewriter/agent.py
AjayThorve: Removed ReportRewriterAgent.tools storage. The generic async runner still supplies tools for constructor compatibility, but the rewriter neither retains nor uses them.

src/aiq_agent/agents/chat_researcher/utils.py
AjayThorve: Removed _extract_query_from_text and the tests that only exercised that compatibility wrapper.

src/aiq_agent/agents/chat_researcher/prompts/intent_classification.j2
AjayThorve: Keeping intent in this PR. route is authoritative for research routing now, but collapsing the meta dual signal is a separate contract simplification as suggested.

src/aiq_agent/agents/chat_researcher/models/intent.py
AjayThorve: Documented target, report_action, and use_parent_report_context in IntentResult.

cdgamarose-nv (approved):

tanleach (approved):
LGTM

PR #276, opened 2026-06-16
Add per-user MCP auth flow

ashan-nv:
## Summary

Adds **per-user MCP authentication** for protected data sources: a user connects a source (e.g. Google Drive via MCP-as-a-Service) through an OAuth popup, AIQ stores that user's token, and async research jobs later call the source's MCP tools **as that user** using their stored token. Also gates job submission so a job can't start against a protected source the owner hasn't connected, and binds each job to its owner's identity for token lookup.

### Why a shared token store (Redis)

The flow spans **two separate processes**, so an in-memory cache won't work:

- The **API process** runs the OAuth connect/callback and **writes** the token.
- The **Dask job worker** (a separate process, potentially a separate replica, already existing code) **reads** the token at job time to call the MCP server.

They coordinate through a **shared, persistent object store** keyed per user. `config_web_frag.yml` wires this as a NAT `object_stores` entry (`mcp_token_store`). Redis is the default backend (`nvidia-nat-redis`), but it's interchangeable with `s3` / `mysql` — the only requirement is that both processes can reach the same store by key. Tokens are isolated by key: each is stored under `principal_user_id = "{principal.type}:{principal.sub}"` derived from the *verified* identity, so a worker only ever retrieves the job owner's token.

### How the auth flow works

**Connect (interactive, API process)** — new routes in `routes/auth.py`:
- `POST /v1/auth/mcp/{source_id}/connect` — starts (or resumes) the OAuth flow via the NAT `mcp_oauth2` provider, returns a provider login URL.
- `GET /v1/auth/mcp/{source_id}/status` — current per-user status (`connected` / `not_connected` / `expired` / `error`).
- `GET /v1/auth/mcp/{source_id}/callback` — OAuth redirect target; exchanges the code for a token and persists it to the shared store, then serves a small HTML page that `postMessage`s the result to the opener and closes.

`NatMcpAuthProvider` (`mcp_auth/nat_provider.py`) is the control plane over NAT's public OAuth/token storage; `factory.py` builds it from the `authentication.mcp_oauth2_*` config block (deriving redirect_uri, scopes, client_id, and the shared token storage). The connect/callback are keyed by an unguessable OAuth `state` bound to `(principal, source)`.

**Middleware (`auth/middleware.py`)** — only the `…/callback` path is made auth-exempt (the provider redirects the browser back with no AIQ token; it's secured by the OAuth `state`). `…/status` and `…/connect` still require a verified principal.

**Job time (headless worker)** — `mcp_auth/runtime_tools.py::open_per_user_mcp_tools` builds the per-user MCP client **in code** per job (after setting `Context.user_id` to the job owner), reads the MCP endpoint from the source's `mcp_oauth2` provider, connects with the owner's stored token, enumerates tools, wraps them for the agent framework, and maps them back to their data source for citation capture. It is deliberately **not** declared as a `per_user_mcp_client` in config, because NAT's interactive (WebSocket) session builder would fail to build it for a user with no token and break interactive chat.

### Submission gating + owner binding

- `mcp_auth/preflight.py::evaluate_mcp_auth` is the single source of truth for "can this job be enqueued given the selected protected sources?" It's enforced at **two** points so they can't drift: the REST `/v1/jobs/async/submit` route (returns a structured **409 `mcp_auth_required`** with connect URLs) and `submit_agent_job` itself (raises `McpAuthRequiredError`), so programmatic submitters like the chat researcher's deep-research path can't bypass the route check.
- `jobs/submit.py` now resolves and passes the verified `principal`, and threads `principal_user_id(principal)` to the worker so job-time token lookup matches the connect-time key.
- The Next.js proxy (`app/api/jobs/async/[...path]/route.ts`) now forwards non-2xx **JSON** bodies verbatim (instead of wrapping them in `BACKEND_ERROR`), so the UI can read the 409's `sources` / `auth_url` and prompt the user to connect.

### Frontend

- `adapters/api/mcp-auth-client.ts` — client for status/connect plus `openAuthPopupAndWait` (opens the OAuth popup, resolves on `postMessage` or popup close, caller re-checks status).
- `DataConnectionCard.tsx` — protected sources show a Connect/Reconnect button and a status line instead of the toggle until connected.
- `DataSourcesPanel.tsx` — drives the connect flow, refreshes statuses afterward, and surfaces an error banner on failure.
- `data-sources.ts` / `data-sources-client.ts` — carry the new `per_user_auth` metadata through to the UI.

### CLI

- `skills/aiq-research/scripts/aiq.py` — adds status/connect handling so the CLI consumer can connect protected sources and see their state.

### Other

- Pinned `nvidia-nat*` from `1.8.0rc4` to the `1.8.0` release across `pyproject.toml` files and `uv.lock`.

## Test plan
- New backend tests: `test_mcp_auth_factory.py`, `test_mcp_auth_provider.py`, `test_mcp_auth_routes.py`, `test_submit_mcp_auth_guard.py`, `test_submit_owner_user_id.py`, plus additions to `test_auth.py` and `test_job_submit_data_sources.py`.
- Pre-commit hooks passed on the committed changes.
- `uv lock` completed successfully after the rebase onto `develop`.
- Not run here: full backend/frontend test suites.

## Summary

* **New Features**
  * Added a protected per-user MCP OAuth source (example: Google Drive) with status/connect/callback flows and OAuth popup-based connection
  * Displayed per-user connection/auth state on the data sources UI, including “Connect/Reconnect” and actionable connect errors
  * Added per-run resolution of per-user MCP tools with correct user context and shared token storage
  * Extended CLI with commands for listing sources, checking auth status, and connecting
* **Bug Fixes**
  * Improved async job/SSE reliability using keepalive signals during idle periods
* **Documentation**
  * Included configuration guidance for redirect/callback wiring and token storage setup


ashan-nv:
@coderabbitai review

frontends/ui/src/features/layout/components/DataSourcesPanel.tsx
```diff
@@ -114,6 +126,21 @@ export const DataSourcesPanel: FC<DataSourcesPanelProps> = memo(function DataSou
...
+    async (sourceId: string) => {
+      const client = createMcpAuthClient({ authToken: idToken })
+      const { status, auth_url } = await client.connect(sourceId)
+      if (status === 'auth_required' && auth_url) {
+        await openAuthPopupAndWait(auth_url, sourceId)
+      }
+      await fetchDataSources(idToken)
+    },
+    [idToken, fetchDataSources]
+  )
```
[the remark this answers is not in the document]
ashan-nv: Fixed

frontends/aiq_api/src/aiq_api/mcp_auth/preflight.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    from aiq_agent.common.data_source_registry import get_all_sources
+    from aiq_agent.common.data_source_registry import get_source
+
+    if data_sources is None:
+        protected_ids = [s.id for s in get_all_sources() if s.per_user_auth and s.per_user_auth.required]
+    else:
+        protected_ids = [
+            sid for sid in data_sources if (src := get_source(sid)) and src.per_user_auth and src.per_user_auth.required
+        ]
```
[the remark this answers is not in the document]
ashan-nv: Will fix

frontends/ui/src/adapters/api/mcp-auth-client.ts
```diff
@@ -0,0 +1,150 @@
...
+    const onMessage = (event: MessageEvent) => {
+      const data = event.data
+      if (data && data.type === 'mcp-auth' && data.source_id === sourceId) {
+        try {
+          popup.close()
+        } catch {
+          /* ignore */
+        }
+        finish({ ok: !!data.ok, sourceId })
+      }
```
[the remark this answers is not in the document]
ashan-nv: Fixed

pyproject.toml
```diff
@@ -31,10 +31,11 @@ readme = "README.md"
...
 dependencies = [
-    "nvidia-nat-core==1.7.0",
-    "nvidia-nat[langchain,async_endpoints,phoenix,mcp]==1.7.0",
-    "nvidia-nat-eval==1.7.0",
-    "nvidia-nat-profiler==1.7.0",
+    "nvidia-nat-core==1.8.0rc4",
+    "nvidia-nat[langchain,async_endpoints,phoenix,mcp]==1.8.0rc4",
+    "nvidia-nat-eval==1.8.0rc4",
+    "nvidia-nat-profiler==1.8.0rc4",
+    "nvidia-nat-redis==1.8.0rc4",
```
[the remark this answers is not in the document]
ashan-nv: 1.8.0 got released after. Updated the version to that which will fix this

frontends/aiq_api/src/aiq_api/routes/auth.py
```diff
@@ -0,0 +1,158 @@
...
+_CALLBACK_HTML = """<!doctype html>
+<html><head><meta charset="utf-8"><title>{title}</title></head>
+<body style="font-family: sans-serif; padding: 2rem; text-align: center;">
+<p>{message}</p>
+<script>
+  try {{ if (window.opener) window.opener.postMessage(
+    {{ type: "mcp-auth", source_id: {source_id_js}, ok: {ok_js} }}, "*"); }} catch (e) {{}}
+  setTimeout(function () {{ window.close(); }}, 400);
+</script>
+</body></html>"""
```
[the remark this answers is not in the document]
ashan-nv: no secret is transmitted, so this is not an issue

frontends/aiq_api/src/aiq_api/mcp_auth/preflight.py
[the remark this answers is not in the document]
ashan-nv: Not triggerable by any shipped config. 

And -- the literal fix isn't viable where the finding points. evaluate_mcp_auth is transport-agnostic — it has no WorkflowBuilder, and resolving "agent-available sources" requires one (_get_agent_available_source_ids calls builder.get_tools). There's no registry-only call to swap in

configs/config_web_frag.yml
```diff
@@ -179,6 +221,63 @@ functions:
...
+#     API process writes the token and a separate worker process reads it.
+#   * redirect_uri MUST point at AIQ's callback: <public-url>/v1/auth/mcp/gdrive/callback
+#     and be reachable by the user's browser (i.e. AIQ's public URL).
+#   * client_id is OPTIONAL. Omitted -> NAT dynamic client registration (DCR):
+#     connect works, but silent token refresh won't survive the API->worker
+#     process boundary, so users Reconnect when a source shows "expired" (the UI
+#     supports this). Set a fixed (e.g. ECI public) client_id for silent refresh.
+#
+# Example below is wired for NVIDIA MaaS (MCP-as-a-Service) Google Drive.
+# Set: MCP_GDRIVE_URL=https://maas.prd.astra.nvidia.com/maas/gdrive/mcp
```
AjayThorve: This public configuration comment exposes the NVIDIA production Astra/MaaS hostname. Replace it with a neutral example host and describe the source generically; keep the real endpoint only in internal deployment configuration.

AjayThorve: something like `https://your-mcp-server.example.com/mcp`

frontends/ui/src/features/layout/components/DataSourcesPanel.tsx
```diff
@@ -114,6 +129,35 @@ export const DataSourcesPanel: FC<DataSourcesPanelProps> = memo(function DataSou
...
+        try {
+          await fetchDataSources(idToken)
+        } catch (err) {
+          console.error('[DataSourcesPanel] Failed to refresh data sources', err)
+        }
+      }
+    },
+    [idToken, fetchDataSources, displaySources]
+  )
+
```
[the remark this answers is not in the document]
ashan-nv: Fixed

frontends/ui/src/features/layout/components/DataSourcesPanel.tsx
```diff
@@ -142,13 +186,15 @@ export const DataSourcesPanel: FC<DataSourcesPanelProps> = memo(function DataSou
...
+  // is enabled, and turns everything off when clicked.
+  const anyAvailableEnabled = enabledAvailableCount > 0
 
   const handleToggleAll = useCallback(() => {
-    const updatedIds = allAvailableEnabled ? [] : availableSources.map((s) => s.id)
+    const updatedIds = anyAvailableEnabled ? [] : availableSources.map((s) => s.id)
     setEnabledDataSources(updatedIds)
     saveDataSourcesToConversation(updatedIds)
-  }, [allAvailableEnabled, setEnabledDataSources, availableSources, saveDataSourcesToConversation])
+  }, [anyAvailableEnabled, setEnabledDataSources, availableSources, saveDataSourcesToConversation])
```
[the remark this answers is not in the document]
ashan-nv: Fixed

configs/config_web_frag.yml
```diff
@@ -179,6 +221,63 @@ functions:
...
+
+authentication:
+  mcp_oauth2_gdrive:
+    _type: mcp_oauth2
+    server_url: ${MCP_GDRIVE_URL:-https://maas.prd.astra.nvidia.com/maas/gdrive/mcp}
+    redirect_uri: ${AIQ_PUBLIC_URL:-http://localhost:8000}/v1/auth/mcp/gdrive/callback
+    token_storage_object_store: mcp_token_store
+    # scopes are discovered from MaaS protected-resource metadata; pin only to override.
+    # client_id: ${MCP_GDRIVE_CLIENT_ID}        # optional — see note above
+    # client_secret: ${MCP_GDRIVE_CLIENT_SECRET}
```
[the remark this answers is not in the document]
ashan-nv: Skipping — the current default is already a generic placeholder (`${MCP_GDRIVE_URL:-https://your-mcp-server.example.com/mcp}`), not the NVIDIA-internal endpoint. The internal URL only appears in commented docs/examples. No change needed.

deploy/compose/docker-compose.yaml
```diff
@@ -135,11 +142,31 @@ services:
...
+    volumes:
+      - redis-data:/data
+    networks:
+      - aiq-network
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
+    restart: unless-stopped
```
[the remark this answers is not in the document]
ashan-nv: Skipping — the `redis` service already publishes no host port (reachable only on the internal Docker network), so the token store isn't exposed on the host. Already addressed.

frontends/aiq_api/src/aiq_api/mcp_auth/factory.py
```diff
@@ -0,0 +1,221 @@
...
+        authorization_url=authorization_url,
+        token_url=token_url,
+        client_id=client_id,
+        client_secret=client_secret,
+        scopes=scopes,
+        redirect_uri=redirect_uri,
+        use_pkce=bool(getattr(cfg, "use_pkce", True)),
+        token_endpoint_auth_method=(getattr(cfg, "token_endpoint_auth_method", None) or "client_secret_post"),
+        resource=resource,
+    )
```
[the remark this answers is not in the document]
ashan-nv: Skipping — the private-attribute access is wrapped in defensive `getattr` + `try/except`, NAT is version-pinned to 1.8.0, and discovery is unit-tested in `test_mcp_auth_factory.py`. A live-NAT integration test is out of scope for this PR.

pyproject.toml
```diff
@@ -200,7 +201,7 @@ dev = [
 ]
 
 [tool.uv]
-prerelease = "if-necessary-or-explicit"
+prerelease = "allow"
```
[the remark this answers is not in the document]
ashan-nv: Skipping (non-blocking, as noted) — tightening the prerelease policy risks resolution for other workspace members that may need prereleases. Leaving `allow` for safety.

deploy/.env.example
```diff
@@ -47,6 +47,16 @@ DASK_DISTRIBUTED__LOGGING__DISTRIBUTED=warning
...
+# -----------------------------------------------------------------------------
+# Per-user MCP auth token store (config_web_frag.yml)
+# Shared, persistent store the API process (connect) and the Dask worker (job)
+# both reach to read/write per-user MCP tokens. Compose defaults to the bundled
+# `redis` service; point these at a managed Redis in production and set a password.
+# -----------------------------------------------------------------------------
+# REDIS_HOST=redis
+# REDIS_PORT=6379
+# REDIS_PASSWORD=
+
```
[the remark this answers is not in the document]
ashan-nv: The bundled Redis is internal-only by design (no host port; internal network). `REDIS_PASSWORD` is scoped to an external/managed Redis, which `config_web_frag.yml` consumes via the commented `password: ${REDIS_PASSWORD}`; for Compose, `env_file: ../.env` forwards it to the backend. Enforced auth on the bundled dev Redis (requirepass + Secret) is a heavy lift left out of this PR's scope.

deploy/compose/docker-compose.yaml
```diff
@@ -44,6 +46,9 @@ services:
       - NAT_JOB_STORE_DB_URL=${NAT_JOB_STORE_DB_URL:-postgresql+asyncpg://aiq:aiq_dev@postgres:5432/aiq_jobs}
       - AIQ_CHECKPOINT_DB=${AIQ_CHECKPOINT_DB:-postgresql://aiq:aiq_dev@postgres:5432/aiq_checkpoints}
       - AIQ_SUMMARY_DB=${AIQ_SUMMARY_DB:-postgresql+psycopg://aiq:aiq_dev@postgres:5432/aiq_jobs}
+      # Shared token store for per-user MCP auth (config_web_frag.yml)
+      - REDIS_HOST=${REDIS_HOST:-redis}
+      - REDIS_PORT=${REDIS_PORT:-6379}
```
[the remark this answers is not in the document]
ashan-nv: Skipping — Compose loads `env_file: ../.env`, which already forwards `REDIS_PASSWORD` into the backend container, so an explicit `environment:` entry is redundant. The bundled Redis is internal-only by design.

deploy/helm/deployment-k8s/values.yaml
```diff
@@ -85,6 +85,9 @@ aiq:
         NAT_JOB_STORE_DB_URL: postgresql+asyncpg://$(DB_USER_NAME):$(DB_USER_PASSWORD)@aiq-postgres:5432/aiq_jobs
         AIQ_CHECKPOINT_DB: postgresql://$(DB_USER_NAME):$(DB_USER_PASSWORD)@aiq-postgres:5432/aiq_jobs
         AIQ_SUMMARY_DB: postgresql+asyncpg://$(DB_USER_NAME):$(DB_USER_PASSWORD)@aiq-postgres:5432/aiq_jobs
+        # Per-user MCP auth token store (config_web_frag.yml)
+        REDIS_HOST: aiq-redis
+        REDIS_PORT: '6379'
```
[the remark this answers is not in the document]
ashan-nv: Acknowledged — wiring `REDIS_PASSWORD` in Helm needs a deployment-specific Secret source, so it's left out of this PR; the bundled Redis is internal-only (ClusterIP). Tracking enforced auth as a follow-up.

configs/config_web_frag.yml
ashan-nv: sure, just updated to something generic

frontends/ui/src/features/layout/components/DataConnectionCard.tsx
```diff
@@ -72,11 +104,15 @@ export const DataConnectionCard: FC<DataConnectionCardProps> = ({
...
+      tabIndex={canToggle ? 0 : -1}
       onClick={handleCardClick}
       onKeyDown={handleCardKeyDown}
       className={`border-base rounded-lg border p-3 transition-colors ${
-        isDisabled ? 'cursor-not-allowed opacity-50' : 'cursor-pointer hover:bg-surface-raised-50'
+        isDisabled
+          ? 'cursor-not-allowed opacity-50'
+          : canToggle
+            ? 'cursor-pointer hover:bg-surface-raised-50'
+            : ''
```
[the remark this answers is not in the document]
ashan-nv: Resolved

src/aiq_agent/common/data_source_registry.py
```diff
@@ -248,6 +295,23 @@ def set_group_source_map(mapping: dict[str, str]) -> None:
...
+    Registering them lets ``get_source_id_for_tool`` resolve their source so
+    citation/source capture treats their results as sources (otherwise shallow
+    research raises EmptySourceRegistryError). Keys are deterministic, so merging
+    globally is idempotent.
+    """
+    if not mapping:
+        return
+    global _tool_source_map
+    _tool_source_map = {**_tool_source_map, **mapping}
+    _rebuild_prefix_index()
```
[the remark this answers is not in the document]
ashan-nv: Fixed

frontends/aiq_api/src/aiq_api/routes/auth.py
```diff
@@ -0,0 +1,158 @@
...
+    )
+    async def mcp_auth_callback(source_id: str, request: Request) -> HTMLResponse:
+        _require_protected_source(source_id)
+        state = request.query_params.get("state")
+        if not state:
+            return _callback_page(source_id, ok=False, message="Missing state. Please restart the connection.")
+
+        error = request.query_params.get("error")
+        if error:
+            return _callback_page(source_id, ok=False, message=f"Authorization was denied: {error}")
```
AjayThorve, line 146: **[P1] Escape OAuth errors before rendering HTML.** The auth-exempt callback reflects the provider-controlled `error` query parameter directly into `_CALLBACK_HTML`. Repro: request this callback with `error=</p><script>document.body.dataset.pwned=1</script><p>`; the response contains that executable script unescaped in the AIQ origin. Please avoid echoing the provider value or HTML-escape it before interpolation, and add a regression test using an HTML payload.

frontends/aiq_api/src/aiq_api/mcp_auth/factory.py
```diff
@@ -0,0 +1,238 @@
...
+    """Resolve the token storage for a source from the mcp_oauth2 config.
+
+    Uses the configured object store (shared across processes) when present;
+    otherwise falls back to a process-local in-memory store with a loud warning
+    (dev only — tokens will not be visible to job workers).
+    """
+    object_store_name = getattr(cfg, "token_storage_object_store", None)
+    if object_store_name:
+        object_store = await builder.get_object_store_client(object_store_name)
+        return ObjectStoreTokenStorage(object_store)
```
AjayThorve, line 73: **[P1] Namespace tokens by protected source.** Every source resolves `ObjectStoreTokenStorage` over the same object-store namespace, while NAT keys tokens only by hashed user ID. Repro: create two token-storage instances over the configured `mcp-tokens` bucket, store `source-a-token` and then `source-b-token` for the same user, and retrieve through both instances; both return `source-b-token`. A second protected source therefore overwrites the first source credentials. Please require distinct object-store namespaces per provider or add a source-aware prefix used consistently by both the API and worker, with a two-source isolation test.

frontends/aiq_api/src/aiq_api/mcp_auth/sqlite_object_store.py
```diff
@@ -0,0 +1,210 @@
...
+        self._db_path = db_path
+        self._prefix = f"{bucket_name}/" if bucket_name else ""
+        self._ttl = ttl
+        self._db: aiosqlite.Connection | None = None
+        self._lock = asyncio.Lock()
+
+    # ── connection / helpers ──
+    async def _conn(self) -> aiosqlite.Connection:
+        if self._db is None:
+            db = await aiosqlite.connect(self._db_path)
```
AjayThorve, line 141: **[P1] Protect the plaintext token database.** SQLite creates this database using the process umask; with the normal `022` umask, creating the store and checking it with `stat -f "%Sp %OLp %N" <db>` produces mode `0644`. Because the serialized objects include plaintext access and refresh tokens, other local users can read the credential store. Please create or chmod the database to `0600`, ensure WAL/SHM files remain restricted, and assert the resulting modes in the SQLite-store tests.

frontends/aiq_api/src/aiq_api/mcp_auth/runtime_tools.py
```diff
@@ -0,0 +1,173 @@
...
+                name: MCPToolOverrideConfig(alias=ov.get("alias"), description=ov.get("description"))
+                for name, ov in (pua.tool_overrides or {}).items()
+            }
+            client_cfg = PerUserMCPClientConfig(
+                server=MCPServerConfig(transport="streamable-http", url=server_url, auth_provider=pua.auth_provider),
+                tool_overrides=tool_overrides,
+            )
+            group = await exit_stack.enter_async_context(per_user_mcp_client_function_group(client_cfg, builder))
+            fns = await group.get_accessible_functions()
+            wrapper = builder._registry.get_tool_wrapper(llm_framework=wrapper_type)
```
AjayThorve: **[P1] Runtime MCP tools cannot be wrapped from a `ChildBuilder`.** In E2E server mode, OAuth completes and the MCP client discovers the Drive tools, but this line raises `AttributeError: 'ChildBuilder' object has no attribute '_registry'`. `open_per_user_mcp_tools()` then swallows the failure and silently falls back to web search—or leaves zero tools when only GDrive is selected. Repro: connect GDrive, select it, and ask to summarize a Drive document; the log shows all MCP tools being added immediately before this exception. Please wrap through a builder-supported API rather than assuming the private `_registry`, add a test using the real `ChildBuilder` registration path, and surface a user-visible failure instead of silently dropping an explicitly selected source.

frontends/aiq_api/src/aiq_api/jobs/runner.py
```diff
@@ -471,27 +483,43 @@ async def run_agent_job(
...
+                    from ..mcp_auth.runtime_tools import open_per_user_mcp_tools
+
+                    async with AsyncExitStack() as mcp_stack:
+                        mcp_tools = await open_per_user_mcp_tools(
+                            builder=builder,
+                            data_sources=data_sources,
+                            exit_stack=mcp_stack,
+                            wrapper_type=LLMFrameworkEnum.LANGCHAIN,
+                        )
+                        agent_tools = [*tools, *mcp_tools] if mcp_tools else tools
```
AjayThorve, line 709: [P1] Keep runtime MCP tools out of the deep-research orchestrator's callable catalog

This works for shallow research, but it currently breaks the deep-research path. Repro: select GDrive (web search can also be enabled), submit `write a deep research analysis on this doc: <private Google Doc URL>`, and approve the plan. The worker successfully logs `Resolved 4 per-user MCP tool(s) for source 'gdrive'`, but passing these tools into `DeepResearcherAgent` also places their names in `tool_set.tools_info`, which is rendered under the orchestrator prompt's `Available Tools` section. The actual orchestrator created in `build_deep_research_graph`, however, is bound only to helper tools plus `run_research_batch`. In the live run it directly called `per_user_mcp_client__health_check`, `per_user_mcp_client__google_drive_read_file`, and `advanced_web_search_tool`; each was rejected with `is not a valid tool, try one of [..., get_verified_sources, run_research_batch]`. No planner/research batch/writer output was produced, and the job eventually failed with the downstream `writer-agent did not produce a final Markdown answer` error.

Potential fix: do not advertise source tools as directly callable in the orchestrator prompt. Keep them bound to the planner/researcher workers, make the orchestrator route all source access through `run_research_batch`, and make each role's tool-name sanitizer allowlist match the tools actually bound to that role. A regression test should assert that a GDrive deep-research request follows planner -> `run_research_batch` -> researcher -> writer and never emits a direct orchestrator MCP call.

frontends/ui/src/features/layout/store.ts
```diff
@@ -116,6 +116,27 @@ export const useLayoutStore = create<LayoutStore>()(
         }
       },
 
+      refreshDataSourceStatus: async (authToken?: string) => {
+        // Silent, selection-preserving refresh: update the source list/auth status
```
AjayThorve: [P1] Do not keep an unusable protected source in the effective selection

This selection-preserving refresh leaves `enabledDataSourceIds` unchanged when a protected source transitions from `connected` to `expired`, `not_connected`, or `error`. Repro: connect and enable GDrive, let its token expire, then reopen the data-sources panel. The card correctly changes to `Session expired` / `Reconnect`, but the response still displays `Selected Data Sources: ... GDrive`, and the outgoing message still includes `gdrive`. With Web Search also enabled, runtime resolution can drop the expired GDrive tools and return a web-only answer while the UI claims GDrive was selected.

Please distinguish desired selection from effective executable selection. At minimum, reconcile protected IDs whose refreshed status is not `connected` out of the IDs rendered/sent for the message (and persist that conversation state), or block submission with a reconnect-required error. The backend should also fail an explicitly selected unavailable protected source instead of silently continuing without it.

frontends/aiq_api/src/aiq_api/jobs/runner.py
```diff
@@ -471,27 +483,43 @@ async def run_agent_job(
...
-                        input_text=input_text,
-                        monitor=cancellation_monitor,
-                        available_documents=available_documents,
-                        data_sources=data_sources,
-                        event_store=event_store,
-                    )
+                    from ..mcp_auth.runtime_tools import open_per_user_mcp_tools
+
+                    async with AsyncExitStack() as mcp_stack:
+                        mcp_tools = await open_per_user_mcp_tools(
```
AjayThorve, line 703: [P1] Let runtime MCP sources reach resolution before declaring that no tools exist

A connected GDrive-only deep-research request is rejected before this code runs. Repro: connect and enable GDrive, disable Web Search, submit an explicit deep-research request. `ChatResearcherAgent` first calls `validate_deep_research_tools()` against the startup-built static tool list; per-user MCP tools are deliberately absent from that list because they are resolved here in the worker. Filtering the static list for `gdrive` therefore yields zero tools and returns `Cannot start deep research: No tools are available`, so the job is never submitted and `open_per_user_mcp_tools()` never gets a chance to resolve the four connected tools.

For the async path, treat a selected configured per-user MCP source as a runtime tool candidate, let MCP auth preflight verify connectivity, and perform the authoritative availability check only after this resolver runs. The synchronous deep path similarly needs to resolve per-user tools before validation, matching shallow research. Please cover connected GDrive-only (submits), expired/disconnected GDrive-only (reconnect required), and genuinely empty selection (no-tools error).

frontends/aiq_api/src/aiq_api/routes/auth.py
ashan-nv: fixed in 22ce041 — html-escape the provider-controlled error before it hits the callback page, with a regression test that feeds an html/script payload.

frontends/aiq_api/src/aiq_api/mcp_auth/sqlite_object_store.py
ashan-nv: fixed in 22ce041 — create the db (and its wal/shm sidecars) at 0600 and tighten an existing 0644 file on open; tests assert the modes.

frontends/aiq_api/src/aiq_api/mcp_auth/runtime_tools.py
ashan-nv: fixed in 22ce041 — resolve the tool wrapper through the builder's registry chain instead of builder._registry (ChildBuilder doesn't have it). verified live: gdrive tools now resolve in server mode.

frontends/aiq_api/src/aiq_api/mcp_auth/factory.py
ashan-nv: a per-source key prefix would desync NAT's job-time read (it looks up tokens/{user} unprefixed via the source's mcp_oauth2 provider), so 22ce041 fails closed instead — two sources sharing one token store is rejected, each needs its own bucket. added a two-source isolation test. open to switching to per-source namespaces end-to-end if you'd prefer that over failing closed.

frontends/aiq_api/src/aiq_api/jobs/runner.py
ashan-nv: fixed in 98519f0 — validate_deep_research_tools now treats a selected configured per-user mcp source as a runtime tool candidate instead of rejecting it against the static startup list, so a connected-gdrive-only request submits. connectivity stays enforced by the submit preflight (evaluate_mcp_auth -> mcp_auth_required when not connected), and the real tool check happens after the worker resolves tools. note: this covers the async path; the synchronous deep path still validates the static list — leaving that as a follow-up since the shipped config uses async.

frontends/aiq_api/src/aiq_api/jobs/runner.py
ashan-nv: fixed in 98519f0 — the orchestrator prompt now advertises only the tools it's actually bound to (helpers + run_research_batch), not the source tools, and a dedicated orchestrator middleware restricts the sanitizer allowlist to match. source access stays delegated through run_research_batch to the researcher. heads up: this touches the shared orchestrator prompt/middleware, so worth a deepresearch eval run before merge to confirm no quality regression.

frontends/aiq_api/src/aiq_api/mcp_auth/factory.py
AjayThorve: if the current implementation prevents the credential overwrite, we could keep it as is

frontends/ui/src/adapters/api/mcp-auth-client.ts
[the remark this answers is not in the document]
AjayThorve: Reopening: this is still reproducible on `66d2fb7`. `openAuthPopupAndWait()` accepts a synthetic `mcp-auth` message without checking `event.source === popup` or a trusted `event.origin`; the existing test dispatches exactly such a message and resolves success.

frontends/aiq_api/src/aiq_api/routes/jobs.py
```diff
@@ -591,6 +643,13 @@ async def submit_job(
         if _sandbox_caps_configured() and _agent_uses_sandbox(builder, agent_config.config_name):
             await _enforce_sandbox_concurrency(db_url, principal)
 
+        # Preflight protected MCP sources: block before enqueue if a selected
+        # protected source is not connected. When data_sources is None the job
+        # may use any tool, so every protected source must be connected.
+        mcp_block = await _preflight_mcp_auth(mcp_auth_provider, principal, req.data_sources)
```
AjayThorve, line 662: REST source validation still rejects runtime-only MCP sources before auth preflight. On `66d2fb7`, submitting `deep_researcher` with `data_sources: ["gdrive"]` returns 422 and never calls submission. `_get_agent_available_source_ids()` derives availability only from static tools, but protected MCP sources intentionally have no static tools. Treat configured per-user MCP sources as runtime candidates before this call, then let this preflight return 409 when disconnected.

frontends/aiq_api/src/aiq_api/jobs/submit.py
```diff
@@ -252,6 +253,21 @@ async def submit_agent_job(
...
+    # when no MCP auth provider is active in this process (nothing to enforce).
+    from aiq_api.mcp_auth.active import get_active_mcp_auth_provider
+    from aiq_api.mcp_auth.preflight import McpAuthRequiredError
+    from aiq_api.mcp_auth.preflight import evaluate_mcp_auth
+
+    mcp_provider = get_active_mcp_auth_provider()
+    if mcp_provider is not None:
+        block = await evaluate_mcp_auth(mcp_provider, principal, data_sources)
+        if block is not None:
+            raise McpAuthRequiredError(block)
```
AjayThorve, line 269: Preserve `mcp_auth_required` when the submit-time preflight detects a late auth-state change. A deterministic current-head repro makes `submit_agent_job()` raise this `McpAuthRequiredError`; the REST route catches it as generic `Exception` and returns 500 `Failed to persist async job authorization metadata`. Catch this exception in the REST caller and serialize `error.response` with the same 409 contract used by route preflight.

src/aiq_agent/agents/deep_researcher/factory.py
```diff
@@ -484,13 +516,21 @@ def build_deep_research_graph(
...
             clarifier_result=context.state.clarifier_result,
-            tools=context.tool_set.tools_info,
+            # Advertise only the tools the orchestrator can actually call. Source
+            # tools (incl. per-user MCP tools like Google Drive) are NOT directly
+            # callable here — the orchestrator delegates all source access through
+            # run_research_batch to the researcher, which holds those tools. Listing
+            # them under "Available Tools" made the orchestrator call them directly
+            # (e.g. per_user_mcp_client__google_drive_read_file), which the runtime
+            # rejects since they aren't bound to this agent.
+            tools=[{"name": t.name, "description": t.description} for t in orchestrator_tools],
```
AjayThorve, line 538: Preserve downstream source awareness while hiding direct source-tool calls. E2E on `66d2fb7` completed OAuth/DCR and logged `Resolved 4 per-user MCP tool(s) for source 'gdrive'`, but this prompt exposes only `orchestrator_tools`; the model then concluded it had no Google Drive access, listed `/` and `/shared`, skipped source-router/planner/writer, and the job failed with `writer-agent did not produce a final Markdown answer`. Keep MCP tools non-callable here, but pass selected source IDs plus the researcher tool catalog in a separate downstream-capabilities section and enforce source-router/planner sequencing outside soft prompt instructions (middleware or an explicit state-graph transition). Add an E2E regression that selects `gdrive` and proves a resolved runtime MCP tool is invoked and `/shared/output.md` is produced.

frontends/aiq_api/src/aiq_api/routes/jobs.py
ashan-nv: good catch — fixed in b00497d. _get_agent_available_source_ids only looked at static tools, so a runtime-only source like gdrive got 422'd before preflight. now configured per-user mcp sources count as runtime candidates, so it falls through to the 409 when disconnected instead.

frontends/aiq_api/src/aiq_api/jobs/submit.py
ashan-nv: yep, fixed in b00497d — the REST route now catches McpAuthRequiredError and returns the same 409 mcp_auth_required (error.response) instead of letting it hit the generic 500 handler on a late auth-state change.

frontends/ui/src/adapters/api/mcp-auth-client.ts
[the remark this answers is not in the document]
ashan-nv: fixed in b00497d — openAuthPopupAndWait now bails unless event.source === popup, so a random page can't spoof the auth-complete message. status poll is still the fallback if COOP cuts the opener. also added a test for the spoofed-source case and fixed the old test to actually set the source.

AjayThorve (approved):
Works great! LGTM

PR #278, opened 2026-06-22
feat(paper-search): add SerpAPI and SearchAPI provider support

gcasilva:
#### Overview

This PR adds a `provider` config field to the Google Scholar paper search tool (`sources/google_scholar_paper_search/`) so developers can choose between three backend APIs:

| Provider | `provider` value | Env var | Sign-up |
|----------|------------------|---------|---------|
| **Serper** (default) | `serper` | `SERPER_API_KEY` | [serper.dev](https://serper.dev/) |
| **SerpAPI** | `serpapi` | `SERPAPI_API_KEY` | [serpapi.com](https://serpapi.com/) |
| **SearchAPI** | `searchapi` | `SEARCHAPI_API_KEY` | [searchapi.io](https://www.searchapi.io/) |

All three query Google Scholar and return results normalized to the same shape (title, year, snippet, link, publication info, citations), so the agent-facing tool behavior is identical regardless of provider.

**Why**: The tool was previously hard-wired to Serper. Organizations that already have SerpAPI or SearchAPI subscriptions — or that need provider-specific features — can now use their preferred backend without forking the tool.

**Backward compatibility**: `provider` defaults to `serper`, the existing `serper_api_key` config field and `SERPER_API_KEY` env var path are unchanged, and all existing workflow configs work without modification.

**Bonus fix**: `pyproject.toml` listed `httpx` as a dependency but the code imports `aiohttp`. Fixed to declare `aiohttp>=3.8.0`.

#### Validation

Ran the narrowest commands covering the changed package, plus full-repo lint:

**Lint (whole repo):**
```
$ uv run ruff check .
All checks passed!

$ uv run ruff format --check .
223 files already formatted
```

**Tests (changed package):**
```
$ uv run pytest sources/google_scholar_paper_search/tests/ -v
============================= 58 passed in 0.06s ==============================
```

Test coverage expanded from 21 → 58 tests, covering:
- Provider init, enum dispatch, and backward-compatible defaults
- Year extraction from publication strings (SerpAPI/SearchAPI embed the year in the summary)
- Response normalization for SerpAPI (`publication_info.summary`, `inline_links.cited_by.total`) and SearchAPI (`publication`, `inline_links.cited_by.total`)
- Fetch payload construction per provider (POST vs GET, `X-API-KEY` header vs `api_key` query param)
- Pagination including SearchAPI's 1-based `page` parameter (vs Serper/SerpAPI `start` offset)
- End-to-end `search()` success and error handling per provider

`sources/google_scholar_paper_search/src/paper_search.py` — the `PaperSearchProvider` enum (line 41) and the `search()` dispatch method (line 107) show the design. `_normalize_serpapi` (line 363) and `_normalize_searchapi` (line 448) show how each provider's raw response maps to the common shape consumed by `format_results`.

The key design decision was a **single `_type: paper_search` with an enum `provider` field** rather than three separate registered types (`paper_search_serper/serpapi/searchapi`). This keeps every existing config working unchanged while adding the provider choice as an opt-in field.

#### Related Issues

- Relates to #

## Summary by CodeRabbit

* **New Features**
  * Paper search now supports multiple backend providers (Serper, SerpAPI, SearchAPI) with configuration-based selection, defaulting to Serper.
* **Documentation**
  * Updated setup guides and environment templates to describe provider selection, required API keys, and example configuration.
* **Bug Fixes**
  * Improved error and warning messages when the selected provider’s API key is missing or a provider request fails.
* **Tests**
  * Expanded automated coverage for provider initialization, dispatch, normalization, pagination, and year-range filtering.


gcasilva:
@AjayThorve What is needed to run the other validations for this PR? Thanks

gcasilva:
@AjayThorve there were false positives related to secret strings used in the google scholar testing that were being caught on the lint validation test that I've fixed with the # pragma: allowlist secret in the last commit. Can you please re-run /ok to test
Appreciate it, thanks

gcasilva:
@AjayThorve can you please re-run /ok to test
Appreciate it, thanks

gcasilva:
Hi @AjayThorve, following up to re-run /ok to test, thanks!

gcasilva:
Hi @AjayThorve, following up to re-run /ok to test, all tests executed successfully locally, so we just need to re-run this on the pipeline as well. Appreciate your support so we can provide these additional Google Scholar backends for developers! Thanks!

AjayThorve:
[gcasilva](https://github.com/gcasilva) can you resolve the conflicts?

sources/google_scholar_paper_search/src/paper_search.py
```diff
@@ -12,57 +12,207 @@
...
+    def _extract_year(publication_info: Any) -> str:
+        """Extract a 4-digit year from a publication summary string.
+
+        SerpAPI and SearchAPI embed the year inside the publication summary
+        (e.g. ``"JL Harper - ..., 1977 - cabdirect.org"``). Serper returns
+        a clean ``year`` field, so this is only used for the other two.
+        """
+        if not isinstance(publication_info, str):
+            return "Unknown Year"
+        match = _YEAR_RE.search(publication_info)
```
AjayThorve: [P1] Parse the actual publication year, not the first four-digit token.

`_YEAR_RE.search()` selects the first `19xx`/`20xx` occurrence. With an official SerpAPI-shaped summary such as `Science Progress in the Twentieth Century (1919-1933 …, 1926 - JSTOR`, this normalizes the publication year to `1919` instead of `1926`; it also forces every pre-1900 paper to `Unknown Year`. That violates the advertised common result shape and feeds factually incorrect metadata to the agent. Please parse the provider's publication-year position—typically the final year before the source suffix—with a defensive fallback, and add tests covering embedded date ranges, arXiv identifiers, and pre-1900 publications.

AjayThorve (commented):
Two findings from current-head code inspection and local runtime validation.

sources/google_scholar_paper_search/src/register.py
```diff
@@ -53,41 +57,66 @@ class PaperSearchToolConfig(FunctionBaseConfig, name="paper_search"):
...
+                provider.value,
+                env_var,
             )
             _missing_key_warned = True
 
-        # Yield a stub function that returns an error message
         async def _paper_search_stub(
             query: str = Field(..., validation_alias=AliasChoices("query", "question")),
-            year: str | None = None,
+            year: str | int | None = None,
```
AjayThorve, line 112: [P2] Preserve a meaningful description for the missing-key stub.

The stub has no docstring, but `FunctionInfo.from_fn()` receives `_paper_search_stub.__doc__`, so the resulting `FunctionInfo.description` is `None`. I reproduced NAT warning that it falls back to the instance name `paper_search_tool`, weakening tool selection in the graceful-degradation path. Please pass an explicit provider-aware description (or restore a useful docstring) and add a registration-level test asserting the missing-key function's description.

sources/google_scholar_paper_search/src/paper_search.py
gcasilva: Addressed in b078d1f.

_extract_year now takes the final year token before the source/host suffix instead of the first 19xx/20xx match, so a summary like (1919-1933, 1926 - JSTOR) now yields 1926 instead of 1919. The year window is widened from 19xx/20xx to 1500-2099 so pre-1900 works are no longer forced to Unknown Year (e.g. 1687 and 1859 now resolve). A negative lookahead also prevents an arXiv id such as 1910.12345 from matching as the year 1910.

Added tests covering date ranges, arXiv identifiers (ignored in favor of the real year, or Unknown when no year is present), and pre-1900 publications. Full suite: 64 passed.

sources/google_scholar_paper_search/src/register.py
gcasilva: Addressed in https://github.com/NVIDIA-AI-Blueprints/aiq/pull/278/changes/b078d1f168e31e1dcadfc1d011881b72dc322326.

The missing-key stub now passes an explicit provider-aware description to FunctionInfo.from_fn instead of _paper_search_stub.__doc__ (which was None). The description names the provider and the unconfigured env var, so tool selection no longer falls back to the instance name.

Added TestRegisterMissingKeyStub, parametrized over all three providers, asserting the FunctionInfo description is non-None and contains both the provider value and the missing env var.

AjayThorve:
@gcasilva This is a good extension to the existing paper_search source. I’m comfortable with the provider-selection surface here, especially since Serper remains the default and existing configs keep working.

Two things before merge:
1. Please fix the remaining direct-tool edge cases from the latest review: don’t log/return the raw query on timeout, and short-circuit direct PaperSearchTool usage when the selected provider key is missing.
2. For SerpAPI/SearchAPI long-term maintenance, please provide CI-usable credentials or a reliable sandbox/test-mode path if you want these treated as core-maintained providers.

If we don’t have CI credentials, I’m still okay merging this as a best-effort/community-supported provider extension, provided the docs make that maintenance boundary clear. The unit coverage is strong enough for the normalization/config surface; live provider CI would be the bar for us owning provider drift.

gcasilva:
> @gcasilva This is a good extension to the existing paper_search source. I’m comfortable with the provider-selection surface here, especially since Serper remains the default and existing configs keep working.
> 
> Two things before merge:
> 
> 1. Please fix the remaining direct-tool edge cases from the latest review: don’t log/return the raw query on timeout, and short-circuit direct PaperSearchTool usage when the selected provider key is missing.
> 2. For SerpAPI/SearchAPI long-term maintenance, please provide CI-usable credentials or a reliable sandbox/test-mode path if you want these treated as core-maintained providers.
> 
> If we don’t have CI credentials, I’m still okay merging this as a best-effort/community-supported provider extension, provided the docs make that maintenance boundary clear. The unit coverage is strong enough for the normalization/config surface; live provider CI would be the bar for us owning provider drift.

Thanks @AjayThorve , on bullet point 2, not sure how I would be able to provide CI credentials for SerpAPI/SearchAPI, as I'm not part of those companies, I'm just an individual developer so I cannot provide my own credentials for your CI to use for testing, how have this been resolved for Serper? Can't you use the same process?

gcasilva:
@AjayThorve Both direct-tool edge cases from bullet 1 are addressed in cde2c3f.

1. **Timeout no longer leaks the query.** The `TimeoutError` handler now logs and returns only the timeout duration and provider — `Paper search timed out after 30s. Try again or narrow the query.` The input query is no longer echoed into logs or tool output on a timeout.

2. **Direct `PaperSearchTool` short-circuits on a missing provider key.** `search()` now resolves the selected provider's key up front and returns a provider-named "unavailable" message instead of dispatching an upstream call with an empty credential. This mirrors the register stub's graceful-degradation path, so direct construction behaves consistently with the registered function.

Tests: added a regression assertion that the query is absent from the timeout result, a parametrized short-circuit test across all three providers, and a no-dispatch test confirming `_search_*` is never reached when the key is missing. Full suite: 68 passed.

AjayThorve:
> not sure how I would be able to provide CI credentials for SerpAPI/SearchAPI, as I'm not part of those companies, I'm just an individual developer so I cannot provide my own credentials for your CI to use for testing, how have this been resolved for Serper? Can't you use the same process?

@gcasilva generally this is a slippery slope, we want to be mindful of the third-party providers we add to the repo, since its meant to be inherently extensible, so these providers can be managed in third-party forks as well, does not need to be on AI-Q team to maintain these unless there is a clear resource allotment for running the maintaince checks via CI. 

Since this is closely integrated with an existing tools, i'll allow it for this specific case, since the paper search tool already exists.

If you can resolve the conflcits and CI, i'll go ahead and merge this

tanleach:
@gcasilva please take a look at teh CICD failures

AjayThorve (approved):

gcasilva:
> @gcasilva please take a look at teh CICD failures

I'm not seeing any CI/CD failures from the latest commit. Can you please link? Thanks

PR #289, opened 2026-06-26
fix(deep-research): salvage inline report + persist planner plan

KyleZheng1284:
## Summary
Two sandbox-independent reliability fixes for the deep researcher agent, split out of the sandbox/artifact-runtime PR so they can be reviewed and merged on their own (PR-271 and other consumers hit these on shared `deep_researcher` code).

- **Writer-agent salvage.** When the orchestrator synthesizes the report inline instead of delegating to `writer-agent`, no `/shared/output.md` is written and the job fails with `writer-agent did not produce a final Markdown answer`, discarding a complete report. `_extract_final_markdown` now falls back to the final assistant message, gated to a substantive report (>=400 chars + a markdown heading, never the completion marker), so workflow chatter is still rejected. The orchestrator Return-Answer step is reworded to return the writer marker rather than synthesize the report itself.
- **Planner plan persistence.** The planner previously wrote `/shared/plan.json` itself via the LLM file tools, which thrashed (`write_file` refuses to overwrite, `edit_file` needs an exact match) whenever the file already existed. `PlanPersistenceMiddleware` now writes the planner's schema-validated `ResearchPlan` to `/shared/plan.json` deterministically via the overwrite-safe `backend.upload_files`; the planner does no file I/O. `planner.j2`/`orchestrator.j2` updated to match.

## Scope
7 files, no sandbox coupling: `agent.py`, `custom_middleware.py`, `factory.py`, `prompts/orchestrator.j2`, `prompts/planner.j2`, and tests (`test_agent.py`, `test_custom_middleware.py`).

## Test plan
- [x] `ruff check` on changed files
- [x] `interrogate` docstring coverage passes (>=80%)
- [x] `pytest tests/aiq_agent/agents/deep_researcher/test_agent.py test_custom_middleware.py test_factory.py` (79 passed)

## Summary by CodeRabbit

* **New Features**
  * Deep research now persistently saves planner output automatically to a shared plan file.
  * Added optional “source routing” flow to insert a source-routing step before planning.
  * Final reports can be recovered from inline completion messages when the main report file isn’t available.

* **Bug Fixes**
  * Reduced cases where substantive reports were missed at the end of a run.
  * Improved progress tracking and final-turn behavior to enforce correct todo/status handling.

* **Chores**
  * Updated deep-research LLM configuration to remove `gpt_oss_llm` and consistently use `nemotron_super_llm`.

### DIFF
--- a/src/aiq_agent/agents/deep_researcher/custom_middleware.py
+++ b/src/aiq_agent/agents/deep_researcher/custom_middleware.py
@@ -23,6 +23,7 @@ from pathlib import Path
 from langchain.agents.middleware import AgentMiddleware
 from langchain.agents.middleware.types import ModelResponse
 from langchain_core.messages import AIMessage
+from langchain_core.messages import SystemMessage
 from langchain_core.messages import ToolMessage
 
 from aiq_agent.common import get_source_id_for_tool
@@ -197,6 +198,47 @@ class ToolVisibilityMiddleware(AgentMiddleware):
     async def awrap_model_call(self, request, handler):
         """Filter hidden tools before an asynchronous model call."""
         return await handler(request.override(tools=self._filter_tools(request.tools)))
+
+
+class TodoSuppressionMiddleware(AgentMiddleware):
+    """Strip the framework's ``write_todos`` tool and its injected prompt for a subagent.
+
+    deepagents attaches ``TodoListMiddleware`` to every subagent, which adds the
+    ``write_todos`` tool plus a system-prompt block telling the agent to use it.
+    Agents that own no progress list - e.g. the planner, which returns a single
+    structured ``ResearchPlan`` - should not have it. Placed after the framework's
+    ``TodoListMiddleware`` in the stack, this removes both the tool and the injected
+    prompt block from the model request, keeping todo tracking solely with the
+    orchestrator. It is a no-op when neither is present.
+    """
+
+    _TODO_TOOL = "write_todos"
+    _TODO_PROMPT_MARKER = "## `write_todos`"
+
+    def _clean_request(self, request: object) -> object:
+        """Return the request with the write_todos tool and its prompt block removed."""
+        overrides: dict[str, object] = {
+            "tools": [tool for tool in request.tools if _request_tool_name(tool) != self._TODO_TOOL]
+        }
+        system_message = getattr(request, "system_message", None)
+        if system_message is not None:
+            blocks = system_message.content_blocks
+            kept = [
+                block
+                for block in blocks
+                if not (isinstance(block, dict) and self._TODO_PROMPT_MARKER in str(block.get("text", "")))
+            ]
+            if len(kept) != len(blocks):
+                overrides["system_message"] = SystemMessage(content=kept)
+        return request.override(**overrides)
+
+    def wrap_model_call(self, request, handler):
+        """Strip write_todos and its prompt before a synchronous model call."""
+        return handler(self._clean_request(request))
+
+    async def awrap_model_call(self, request, handler):
+        """Strip write_todos and its prompt before an asynchronous model call."""
+        return await handler(self._clean_request(request))
 
 
 class ToolRetryMiddleware(AgentMiddleware):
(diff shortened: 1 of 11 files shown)


cdgamarose-nv (approved):
Can you implement coderabbit's suggestion? Otherwise, looks good!

src/aiq_agent/agents/deep_researcher/custom_middleware.py
```diff
@@ -406,6 +407,80 @@ def get_source_list_text(self, mode: str = "compact") -> str | None:
...
+        if plan is None:
+            return
+        if hasattr(plan, "model_dump"):
+            payload = plan.model_dump(mode="json", exclude_none=True)
+        elif isinstance(plan, dict):
+            payload = plan
+        else:
+            return
+        content = json.dumps(payload, indent=2, ensure_ascii=False).encode("utf-8")
+        responses = self.backend.upload_files([(self.path, content)])
```
AjayThorve, line 496: This should not be best-effort now that the planner prompt no longer writes `/shared/plan.json` itself. If `upload_files()` returns an error or raises, the orchestrator still proceeds to Step 4 and reads `/shared/plan.json`, which will either fail later in a less actionable way or potentially consume stale plan content if the file already existed. Since this middleware is the sole producer of a required workflow artifact, please make persistence failure fatal to the planner task and surface the upload error directly.

AjayThorve (changes requested):

src/aiq_agent/agents/deep_researcher/custom_middleware.py
AjayThorve: Suggested fix: make `_persist_plan()` raise when `upload_files()` returns any error response, and let that exception propagate from `awrap_model_call` / `aafter_agent` instead of catching and logging it. The existing `test_backend_failure_does_not_propagate` should flip to assert the planner task fails with the upload error.

src/aiq_agent/agents/deep_researcher/custom_middleware.py
```diff
@@ -406,6 +407,74 @@ def get_source_list_text(self, mode: str = "compact") -> str | None:
...
+        errors = [f"{response.path}: {response.error}" for response in responses if getattr(response, "error", None)]
+        if errors:
+            msg = f"Failed to persist plan to {self.path}: {'; '.join(errors)}"
+            logger.error(msg)
+            raise RuntimeError(msg)
```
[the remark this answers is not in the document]
KyleZheng1284: Are you sure this applies to our backend?

AjayThorve (approved):
LGTM

src/aiq_agent/agents/deep_researcher/prompts/planner.j2
```diff
@@ -22,7 +20,9 @@ Before generating the research plan, deeply analyze the user's request:
 5. **Language**: What language is the user's request in?
 
 ## Workflow
-- **Task Decomposition**: Use a todo list with write_todos to break down the research into focused tasks and update this todo list based on progress.
+- **Task Decomposition**: Break the research into focused answer components and ResearchQuery objects in the final
+  ResearchPlan. Do not call `write_todos`. Todo tracking is owned by the orchestrator; the planner's job is to use
```
cdgamarose-nv: If we dont want planner to write todos, we should just remove the todo list middleware rather than bloating prompts. 
Todo list middleware adds todo related instructions to system prompt, so there'll be conflicting instructions

src/aiq_agent/agents/deep_researcher/prompts/orchestrator.j2
```diff
@@ -38,27 +38,35 @@ Step 1. **Task Decomposition and Progress Tracking**: Use `write_todos` to track
...
 
-Step 5. **Return Answer**: Return only the final Markdown answer in the final message. Do not add workflow commentary.
+Step 5. **Return Answer**: After writer-agent returns, return only its short completion marker. The runtime loads the final report from `/shared/output.md`; do not synthesize or write the report yourself.
 {% endif %}
 
 ## Progress Tracking
-Use write_todos to update progress after each workflow step. Use status values: "pending", "in_progress", or "completed".
-
-Before returning the final answer, invoke write_todos to mark all tasks as "completed". Do not output tool calls as text.
+Use write_todos only after a workflow-advancing tool has already run. Use status values: "pending", "in_progress",
```
cdgamarose-nv, line 59: Does the orchestrator with nemotron super mess it up without this block? If it does, then okay to leave it in.

cdgamarose-nv (approved):
Minor comment, otherwise approved

src/aiq_agent/agents/deep_researcher/prompts/orchestrator.j2
AjayThorve: yeah, had to do a lot of trial and error, but seemed to follow this specifically

src/aiq_agent/agents/deep_researcher/prompts/planner.j2
AjayThorve: fair enough

src/aiq_agent/agents/deep_researcher/prompts/orchestrator.j2
cdgamarose-nv: Cool

src/aiq_agent/agents/deep_researcher/prompts/planner.j2
AjayThorve: @KyleZheng1284 can you do the needful

PR #291, opened 2026-06-29
feat(api): encrypt async job content artifacts

tanleach:
#### Overview

Add application-level envelope encryption at rest for async job final reports
and selected artifact event content. The implementation supports disabled,
static-key, and HashiCorp Vault Transit modes; uses per-job AES-256-GCM data
encryption keys; and fails closed when encrypted report or event data cannot be
authenticated or decrypted.

The hardening in this branch:

- protects shared DEK-cache, manager, readiness, and Vault-client state under
  concurrent access;
- offloads synchronous Vault and decryption operations from FastAPI async paths;
- verifies both generate and unwrap permissions during cached Vault readiness;
- retries only classified transient Vault and network failures;
- propagates a versioned, non-secret encryption-policy identity from the API to
  Dask workers and rejects mismatched worker configurations before `RUNNING`;
- propagates encrypted event-persistence failures so affected jobs fail rather
  than silently losing protected data;
- surfaces encrypted state and report failures through explicit HTTP and SSE
  contracts;
- separates process liveness (`/live`) from dependency readiness (`/health`) and
  configures Helm to use the appropriate endpoint for each probe;
- maps report-edit submission-time encryption outages to HTTP 503 without
  exposing exception details; and
- documents the protected fields, plaintext boundary, rollout constraints, and
  operational behavior.

Encryption remains disabled by default. This release intentionally does not
include plaintext migration, backfill, application-managed key rotation,
rewrap, or decrypt-on-rollback tooling. Opt-in operators must follow the
documented forward-only rollout constraints and retain the original encryption
configuration while encrypted jobs exist. Database-writer replay, deletion,
and reordering remain outside the initial confidentiality-at-rest threat model.

#### Validation

```console
$ uv lock --check
Resolved 355 packages in 2ms

$ uv run ruff check \
    frontends/aiq_api/src/aiq_api/auth/middleware.py \
    frontends/aiq_api/src/aiq_api/routes/jobs.py \
    frontends/aiq_api/tests/test_auth.py \
    frontends/aiq_api/tests/test_content_encryption_routes.py \
    frontends/aiq_api/tests/test_report_edit.py \
    tests/deploy/test_helm_deployment_k8s.py
All checks passed!

$ uv run ruff format --check \
    frontends/aiq_api/src/aiq_api/auth/middleware.py \
    frontends/aiq_api/src/aiq_api/routes/jobs.py \
    frontends/aiq_api/tests/test_auth.py \
    frontends/aiq_api/tests/test_content_encryption_routes.py \
    frontends/aiq_api/tests/test_report_edit.py \
    tests/deploy/test_helm_deployment_k8s.py
6 files already formatted

$ uv run pytest frontends/aiq_api/tests -q
303 passed, 1 skipped, 74 warnings in 15.03s

$ uv run pytest tests/aiq_agent/jobs/test_runner.py -q
106 passed in 3.89s

$ helm version --short
v3.21.2+g1259634

$ uv run pytest tests/deploy/test_helm_deployment_k8s.py -q
4 passed in 0.16s

$ uv run --extra docs sphinx-build -M html docs/source docs/build
build succeeded, 3 warnings

$ git diff --check
# passed
```

The documentation warnings are pre-existing missing-reference warnings.
The API suite also emits existing dependency deprecation and periodic-cleanup
warnings without test failures.

Manual Docker validation: rebuilt the images from the local Dockerfiles and
successfully ran the application in Docker.

Start with `frontends/aiq_api/src/aiq_api/jobs/crypto.py` for the envelope,
readiness, Vault retry, concurrency, and policy-identity model. Then review:

- `frontends/aiq_api/src/aiq_api/jobs/submit.py` and `runner.py` for the
  API-to-worker policy boundary;
- `frontends/aiq_api/src/aiq_api/jobs/event_store.py` for encrypted persistence
  and failure propagation;
- `frontends/aiq_api/src/aiq_api/routes/jobs.py` for readiness, report/state
  decryption, report-edit submission, and HTTP/SSE error contracts; and
- `frontends/aiq_api/src/aiq_api/auth/middleware.py` plus
  `deploy/helm/deployment-k8s/values.yaml` for liveness/readiness exposure.

Focused regression coverage is in the content-encryption, report-edit, runner,
authentication, and Helm deployment tests.

#### Related Issues

- N/A

## Summary by CodeRabbit

* **New Features**
  * Added async job content encryption, with support for protected job reports and selected artifact content.
  * Introduced a separate `/live` endpoint for basic process health checks.

* **Bug Fixes**
  * Improved readiness handling so encryption or dependency issues return clear 503 responses without restarting healthy processes.
  * Updated job report, state, and stream behavior to surface encryption-related errors consistently.

* **Documentation**
  * Added deployment and API documentation for encryption setup, health checks, and Kubernetes/Helm usage.


frontends/aiq_api/src/aiq_api/routes/jobs.py
```diff
@@ -428,7 +458,8 @@ async def health_check():
         responses={
             400: {"description": "Unknown agent type or invalid request"},
             422: {"description": "One or more unknown or agent-unavailable data source IDs"},
-            503: {"description": "Dask scheduler not available"},
+            500: {"description": "Content encryption configuration is invalid or job submission failed"},
+            503: {"description": "Content encryption is not ready"},
```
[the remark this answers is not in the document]
tanleach: This behavior is intentional. Content encryption is opt-in, but once enabled, startup validation is a security gate. Missing, invalid, or unusable encryption configuration must prevent startup rather than leave a partially configured service running.

Encryption-disabled deployments are unaffected, and runtime failures remain fail-closed without plaintext fallback.

frontends/aiq_api/src/aiq_api/jobs/runner.py
```diff
@@ -324,6 +325,20 @@ async def run_agent_job(
 
     try:
         job_store = JobStore(scheduler_address=scheduler_address, db_url=db_url)
+        try:
+            from .crypto import ContentEncryptionError
+            from .crypto import create_job_content_cipher
+
+            job_output_cipher = create_job_content_cipher(job_id)
```
AjayThorve, line 553: **[P1] Carry the encryption requirement across the API → worker boundary.**

The submit path validates only the API process, while this worker reconstructs the cipher solely from its local environment, where `AIQ_CONTENT_ENCRYPTION=off` is the valid default. An external Dask worker missing the encryption configuration will therefore accept a job from an encryption-enabled API and persist its final output and artifact content as plaintext; the API readiness check cannot detect that drift. I reproduced this with an API manager ready in `key` mode and a worker manager in `off` mode—the worker returned the original plaintext JSON without an `aiqenc:` envelope.

Please capture a non-secret encryption policy identity at submission, pass it with the Dask job, and make the worker compare it with its local configuration before setting the job to `RUNNING`. At minimum this must reject required-encryption → `off`; ideally it also verifies the mode and stable key identity/fingerprint. Add a test with intentionally different API and worker configurations.

frontends/aiq_api/src/aiq_api/jobs/event_store.py
```diff
@@ -398,11 +458,12 @@ def store_batch(self, events: list[dict]):
 
         rows = []
         for event in events:
+            stored_event = self._prepare_event_for_storage(event)
```
AjayThorve, line 464: **[P1] Propagate encrypted event persistence failures to the job.**

The event is encrypted here, but the subsequent `store()` / `store_batch()` transaction catches every database exception, logs it, and returns normally. `BatchingEventStore.flush()` therefore reports success and the runner can mark the job `SUCCESS` even though encrypted artifact events were dropped. A timer-triggered flush also has no path to report encryption or persistence failures back to the owning job. This contradicts the documented guarantee that encrypted persistence failures mark the job failed.

In encrypted mode, re-raise persistence/encryption failures and make background flush failures observable by the runner before final success. Preserve best-effort event persistence only for `off` mode if that remains the intended legacy contract, and add a regression test for a transient database failure during an encrypted batch flush.

AjayThorve (commented):
Two blocking encryption-boundary issues found in the current PR head.

frontends/aiq_api/src/aiq_api/routes/jobs.py
```diff
@@ -415,6 +424,27 @@ async def health_check():
 
             return JSONResponse(status_code=503, content=result)
 
+        try:
+            encryption = await get_content_encryption_health_async()
```
AjayThorve: **[P1] Ensure the encryption health check is the route served at runtime.**

NAT registers its own `/health` route before `register_job_routes()` adds this one. FastAPI keeps both routes, but request matching selects the first registration, so a real `nat serve` process returns the earlier NAT `{"status":"healthy"}` response and never executes this encryption readiness check. I reproduced this in both `off` and `key` modes; `/openapi.json` describes this handler while `GET /health` still returns the earlier response. An invalid or unavailable Vault/key configuration can therefore remain HTTP 200 healthy, defeating the deployment readiness gate.

Replace or extend the existing NAT health route rather than registering a duplicate path, and add an integration test against the actual assembled `nat serve` app that asserts enabled-but-unready encryption returns 503 and includes the encryption readiness object.

frontends/aiq_api/src/aiq_api/jobs/runner.py
tanleach: good catch on this. I've added a check to make sure that both policies match

tanleach:
@coderabbitai review

frontends/aiq_api/src/aiq_api/routes/jobs.py
```diff
@@ -498,6 +520,92 @@ async def register_job_routes(app: FastAPI, builder: WorkflowBuilder, worker: Fa
...
+
+        # Check DB connectivity using any cached async engine
+        try:
+            cache = EventStore._async_engine_cache
+            if cache:
+                engine = next(iter(cache.values()))[0]
+                async with engine.connect() as conn:
+                    await asyncio.wait_for(conn.execute(text("SELECT 1")), timeout=3.0)
+            else:
+                result["db"] = "no_engine"
```
AjayThorve: This `no_engine` branch is the normal fresh-process path, but it leaves `status` as `healthy` and returns HTTP 200 without touching `db_url`. I reproduced this against the assembled app: after startup I made the SQLite database genuinely unusable, confirmed a direct connection failed, and `GET /health` still returned `200` with `{"status":"healthy","db":"no_engine",...}` because the async-engine cache was empty. Since Helm now uses `/health` for readiness, a fresh pod can remain ready through a database outage even though this endpoint promises database readiness.

Always obtain or create an engine for the configured `db_url` and execute the bounded `SELECT 1`; an absent cached engine must not be treated as healthy. Add an assembled-app regression test that makes the database unavailable after startup, before any `EventStore` async engine is cached, and asserts `/health` returns 503.

tanleach: pushed a new commti for this

AjayThorve (approved):
Looks great, thanks for adding this feature

PR #292, opened 2026-06-29
feat: add workflow and agent boundary guardrails

ericevans-nv:
#### Overview

Adds NeMo Guardrails support at the workflow, shallow researcher, and deep researcher boundaries through NAT middleware aliases. The implementation lets configured rails evaluate selected boundary fields, pass content through unchanged, block content with a configured refusal, or modify selected fields before execution continues.

This PR includes:

- Middleware aliases for the supported boundaries: `workflow_guardrails`, `shallow_agent_guardrails`, and `deep_agent_guardrails`.
- A guarded default web configuration at `configs/config_web_default_guardrails.yml`.
- Dynamic field selection for guardrail evaluation and modification.
- Message-type field selection for agent states, including `HumanMessage.content` and `AIMessage.content`.
- Dotted field paths for nested response objects, such as selecting `message.content` from response choices.
- Focused unit coverage for workflow, shallow agent, deep agent, and shared guardrails field-selection behavior.
- Documentation for configuring guardrails at the supported workflow and agent boundaries.

#### Implementation Notes

Guardrails are applied at these AI-Q boundaries:

| Boundary | Middleware | Applies To |
| --- | --- | --- |
| Workflow | `workflow_guardrails` | Workflow input and final assistant response. |
| Shallow researcher | `shallow_agent_guardrails` | Shallow researcher input and output message content. |
| Deep researcher | `deep_agent_guardrails` | Deep researcher input and output message content. |

The `workflow_functions` configuration defines both where middleware is attached and which string fields guardrails evaluate and can modify. For nested response objects, selected fields can use dotted paths. For agent state messages, selected fields can be scoped by message type:

```yaml
workflow_functions:
  shallow_research_agent:
    messages:
      HumanMessage:
        - content
      AIMessage:
        - content
```

This allows input rails to evaluate user message content and output rails to evaluate assistant message content while leaving non-selected fields untouched.

#### Validation

Focused unit tests validated the guardrails middleware behavior at the workflow, shallow researcher, and deep researcher boundaries. The guardrails test suite passed with 43 tests, covering pass, block, and modify behavior; dynamic field selection; dotted field paths; message-type field selection for `HumanMessage.content` and `AIMessage.content`; and blocked agent responses returning through the expected state/message schema.

Smoke testing used the guarded web configuration with deterministic regex rails to exercise the configured boundary behavior end to end. The regex rails are used as simple trigger policies so the smoke tests can reliably force pass, block, and modify outcomes without depending on model output. These runs validated that workflow input/output guardrails and shallow/deep agent input/output guardrails are attached at the expected boundaries and apply the configured decision to the selected fields.

Documentation validation confirmed that the new guardrails page is linked into the docs navigation and builds successfully with Sphinx. Pre-commit checks for the touched docs files also passed, including markdown link check.

Start with `src/aiq_agent/guardrails/register.py` for the middleware registration surface, then review the workflow and agent middleware implementations under `src/aiq_agent/guardrails/`. The most important behavioral tests are in `tests/aiq_agent/guardrails/`.

#### Related Issues

- Relates to #

## Summary by CodeRabbit

* **New Features**
  * Added guardrails middleware support for workflow, shallow-agent, and deep-agent boundaries.
  * Introduced a default Web-mode guardrails configuration with multiple LLM providers and targeted input/output rewriting.
* **Documentation**
  * Added a Guardrails customization guide and linked it from the customization docs.
* **Bug Fixes**
  * Improved boundary blocking/modify handling with state-aware updates to affected messages.
* **Tests**
  * Added unit tests covering workflow, shallow-agent, and deep-agent guardrails behaviors.
* **Chores**
  * Updated NAT/evaluation/security versions and added a Guardrails plugin entry point.


configs/config_web_default_guardrails.yml
```diff
@@ -0,0 +1,237 @@
...
+
+  shallow_agent_guardrails:
+    _type: shallow_agent_guardrails
+    workflow_functions:
+      shallow_research_agent:
+        messages:
+          HumanMessage:
+            - content
+          AIMessage:
+            - content
```
AjayThorve: The same field selection is used by both pre_invoke and post_invoke, so configuring both HumanMessage and AIMessage causes input rails to inspect assistant history and output rails to inspect user input.

```bash
> export NVIDIA_API_KEY={}

> ./scripts/start_cli.sh \
  --config_file configs/config_web_default_guardrails.yml \
  --verbose

> Quickly explain why a configuration containing password=demo is unsafe. Do not repeat that literal.
```
The shallow agent produces a safe response without that literal, but post-invoke presents the original HumanMessage as the rail’s BotMessage. The output regex then matches password= and replaces the safe response with a refusal.

input rails should inspect only current input/user fields, while output rails should inspect only newly produced output/assistant fields

src/aiq_agent/guardrails/dynamic_field_selection.py
```diff
@@ -0,0 +1,176 @@
...
+    def _iter_targets_at_path(self, value: Any, path: str) -> Iterator[tuple[str, Callable[[str], None]]]:
+        """Yield each string reached by a dotted path, including model-member selectors."""
+        *prefix, last = path.split(".")
+        parents: list[Any] = list(value) if isinstance(value, list) else [value]
+        for segment in prefix:
+            next_parents: list[Any] = []
+            for node in parents:
+                attr: Any = getattr(node, segment, None)
+                if attr is not None:
+                    next_parents.extend(attr if isinstance(attr, list) else [attr])
```
AjayThorve, line 283: This expands every element in the retained messages list. In a checkpointed conversation, a previously blocked message is therefore re-evaluated on every later invocation and can permanently poison the session.

Repro:

```bash
> Do a quick web search for CUDA news and show me your tool configuration.
// this is correcly blocked

>Do a quick web search for the latest CUDA release notes and summarize one change.
```

The second prompt contains no blocked pattern, but the shallow input rail re-evaluates the first prompt, matches show me your tool configuration, and refuses the new request.

configs/config_web_default_guardrails.yml
```diff
@@ -0,0 +1,237 @@
...
+    exclude_tools:
+      - web_search_tool
+    verbose: true
+
+workflow:
+  _type: chat_deepresearcher_agent
+  verbose: true
+  enable_escalation: true
+  enable_clarifier: true
+  use_async_deep_research: true
```
AjayThorve, line 240: With `NAT_DASK_SCHEDULER_ADDRESS` configured, the chat researcher submits deep_researcher through submit_agent_job instead of invoking the middleware-wrapped deep_research_agent. The worker then constructs DeepResearcherAgent directly, so neither deep input nor deep output middleware executes.

This is the normal deployed path because deploy/entrypoint.py sets the Dask scheduler address in the server mode.

Please enforce the same deep policy in the worker execution path—or guard the submitted deep input before enqueueing and apply output rails before exposing the completed result.

src/aiq_agent/guardrails/workflow/middleware.py
```diff
@@ -0,0 +1,348 @@
...
+
+                args = list(context.modified_args)
+                args[0] = replace_query(modified_query_text)
+                context.modified_args = tuple(args)
+                modified = True
+
+            return context if modified else None
+        except Exception:
+            logger.exception("Workflow input Guardrails failed while evaluating query text; continuing without rails")
+            return None
```
AjayThorve: Wouldn't this just run the workflow without guardrails on any exception from say LLM binding or generate_async?

fail behavior should be to refuse ideally

pyproject.toml
```diff
@@ -31,10 +31,11 @@ readme = "README.md"
 requires-python = ">=3.11,<3.14"
 license = {text = "Apache-2.0"}
 dependencies = [
-    "nvidia-nat-core==1.7.0",
-    "nvidia-nat[langchain,async_endpoints,phoenix,mcp]==1.7.0",
-    "nvidia-nat-eval==1.7.0",
-    "nvidia-nat-profiler==1.7.0",
+    "nvidia-nat-core==1.8.0",
```
AjayThorve, line 34: This PR changes NAT from 1.7 to 1.8, whose `_run_workflow` contract adds the `streaming` keyword. AIQ’s existing override implements the NAT 1.7 signature, so every UI WebSocket request now fails with unexpected keyword argument 'streaming'. This is a regression introduced by the dependency upgrade.

to test his, use `./scripts/setup.sh` followed by `./scripts/start_e2e.sh --config_file ./configs/..yaml`

configs/config_web_default_guardrails.yml
```diff
@@ -0,0 +1,237 @@
...
+  checkpoint_db: ${AIQ_CHECKPOINT_DB:-./checkpoints.db}
+  middleware:
```
AjayThorve, line 242: Attaching this middleware to the streaming web workflow changes the emitted payload from response tokens/content into the Python string representation of ChatResponse.


Repro:
```
./scripts/start_e2e.sh \
  --config_file configs/config_web_default_guardrails.yml
```
Send Hello in the UI. Instead of rendering the assistant content, the UI renders:
```
id='research_response' object='chat.completion' ...
choices=[ChatResponseChoice(...)]
```

NAT’s non-streaming guardrail path buffers the stream and stringifies each structured chunk. Enabling stream_output_rails is not directly available here because NAT rejects it with mapping-form workflow_functions.
Please preserve the original ChatResponse/streaming schema while applying final-output rails, and add a WebSocket E2E test asserting that the UI receives assistant content rather than a model representation.

AjayThorve (changes requested):
did e2e testing with cli mode and e2e (web ui mode), requested a few changes. This for most part works, great work!!

configs/config_web_default_guardrails.yml
ericevans-nv: Fixed. The dynamic field selection now supports phase-specific targets under `workflow_functions`: `pre_invoke` selections are used for input rails and `post_invoke` selections are used for output rails. The shallow/deep guardrail config now targets `HumanMessage.content` only for pre-invoke and `AIMessage.content` only for post-invoke, so output rails no longer inspect the original user input.

src/aiq_agent/guardrails/dynamic_field_selection.py
ericevans-nv: The shallow and deep agent guardrail middleware now override field selection for agent message history so rails evaluate only the current/latest matching message instead of expanding all retained messages.

configs/config_web_default_guardrails.yml
ericevans-nv: The async Dask worker path now preserves the configured NAT middleware boundary for deep_research_agent. Since the worker executes DeepResearcherAgent.run(...) directly instead of invoking the registered NAT function, the worker now adapts that direct call through NAT’s FunctionMiddlewareChain using the configured function name/config. This applies middleware targeted through both the function’s middleware list and workflow_functions, while ignoring unrelated full-config middleware entries.

src/aiq_agent/guardrails/workflow/middleware.py
ericevans-nv: Agreed. The workflow middleware was failing open when guardrail evaluation raised after selecting input text. I updated the input-rail failure path to fail closed and return the boundary’s normal refusal shape. Since the same runtime failure mode applies to shallow and deep agent input rails, the shared guardrail mixin now applies the same fail-closed behavior for those boundaries as well, using each boundary’s existing block-response adaptation so schema/state shape is preserved.

pyproject.toml
ericevans-nv: Updated the WebSocket workflow override to match the NAT 1.8 _run_workflow signature by accepting the streaming argument and passing it through to generate_streaming_response. This preserves the existing streaming behavior by default while fixing the UI/WebSocket path when NAT calls the override with the new keyword.

configs/config_web_default_guardrails.yml
ericevans-nv: Updated this path so output rails can run against structured streaming responses without converting the streamed payload into a Python object string. The middleware now applies rails to the configured text fields and writes pass/modify/block results back into the original response shape, so the UI continues to receive assistant content instead of a ChatResponse / ChatResponseChoice representation.

src/aiq_agent/guardrails/shallow_agent/middleware.py
```diff
@@ -0,0 +1,141 @@
...
+
+        When a shallow-agent state is available, the refusal is appended as the
+        next assistant message so downstream callers still receive
+        ``ShallowResearchAgentState``. If no state can be resolved, return the
+        raw refusal text.
+        """
+        state = self._get_shallow_agent_state_from_invocation_context(context, original_output)
+        if state is None:
+            return block_message
+        return state.model_copy(update={"messages": [*state.messages, AIMessage(content=block_message)]})
```
AjayThorve, line 61: This appends the refusal without removing or replacing the blocked assistant output.

Repro: return a state containing a human message followed by `AIMessage("password=secret")`, then have the output rail block it. The returned state still contains that original assistant message, followed by the refusal. The immediate caller may display only the last message, but any serializer, history consumer, or debugger can still read the content that the output rail rejected.

Please replace or remove the blocked assistant message while preserving the state schema, and cover the same behavior in both shallow and deep middleware tests.

src/aiq_agent/guardrails/interface/middleware.py
```diff
@@ -0,0 +1,178 @@
...
+
+        if all(isinstance(chunk, str) for chunk in buffered):
+            ctx.output = "".join(buffered)
+            result = await self.post_invoke(ctx)
+            if result is not None:
+                ctx = result
+            yield ctx.output
+            return
+
+        for chunk in buffered:
```
AjayThorve: This checks each structured chunk independently, so output rails can be bypassed when blocked text spans chunk boundaries.

Repro: stream two ChatResponse chunks whose message contents are `system ` and `prompt:`. Neither chunk matches the configured `system\s+prompt\s*[:=]` pattern alone, so both are emitted and the consumer reconstructs `system prompt:`.

Since the stream is already buffered here, please evaluate the combined logical assistant output before emitting it, then preserve the structured response shape when applying a block or rewrite. A regression test should split a blocked pattern across two chunks.

src/aiq_agent/guardrails/shallow_agent/middleware.py
ericevans-nv: Output-rail blocks now replace the rejected assistant message in the shallow/deep agent state instead of appending a refusal after the blocked content.

src/aiq_agent/guardrails/interface/middleware.py
ericevans-nv: Output-rail blocks now replace the rejected assistant message in the shallow/deep agent state instead of appending a refusal after the blocked content.

AjayThorve (approved):
All issues addressed, LGTM! Approved, pending CI fixes which seem unrelated

PR #294, opened 2026-06-29
chore: transfer signed SPDX commit

AjayThorve:
Temporary draft used only to introduce the GPG-signed SPDX commit object into the upstream repository so the origin-owned branch for #293 can point to it. This PR will be closed immediately after the ref transfer.


AjayThorve:
Transfer complete: the GPG-verified commit now backs the origin-owned branch for #293.

Issue #287, opened 2026-06-26
OpenShell fail-closed policy attestation


KyleZheng1284:
OpenShell is not truly job-scoped in the provided config. The code normalizes a job-scoped name, but then attaches to the configured providers.openshell.sandbox_name when present, and the example uses one fixed aiq-openshell-demo sandbox with delete_on_exit: false

_Originally posted by @cdgamarose-nv in https://github.com/NVIDIA-AI-Blueprints/aiq/issues/280#issuecomment-4794452275_

- the provider still declares capabilities without verifying the attached sandbox's applied policy/network mode; tracked as a follow-up pending OpenShell SDK introspection.

cdgamarose-nv:
This was addressed by #298: OpenShell now creates and owns a sandbox per deep-research job by default, shared-sandbox attachment is explicitly debug-only, and execution fails closed unless the effective policy and status attestation match.

PR #298, opened 2026-07-01
feat: isolate and attest OpenShell jobs

KyleZheng1284:
#### Overview

Create and own one physical OpenShell sandbox per AI-Q deep-research job. This change provides per-job isolation, fail-closed policy attestation, discoverable ownership labels, truthful cleanup, authenticated gateway readiness checks, pytest-owned live acceptance, and a canonical operator guide.

The previous upstream blocker is resolved by [NVIDIA/OpenShell#2170](https://github.com/NVIDIA/OpenShell/pull/2170) and released in OpenShell `0.0.80`. AI-Q now requires `openshell>=0.0.80,<0.1`; the setup script defaults to and enforces the `0.0.80` supported floor. Runtime security decisions remain capability- and state-based rather than branching on an OpenShell version.

Key behavior:

- Create one owned physical sandbox lazily for each job and reuse it for subsequent execution within that job.
- Apply `aiq=deep-research` and a normalized `aiq-job-id` to both OpenShell request metadata and template/container metadata. Active jobs can be queried with `openshell sandbox list --selector aiq=deep-research`.
- Keep shared-sandbox attachment behind explicit `allow_shared_sandbox=true` debug configuration. Shared attachment is non-production and does not transfer cleanup ownership to the job.
- Enforce AI-Q's declared network boundary:
  - blocked mode rejects every configured endpoint;
  - allowlist mode requires non-empty normalized hosts contained in `network.allow`;
  - hostless endpoints and `allowed_ips`/CIDR exceptions are rejected.
- Attest the exact submitted policy through authoritative OpenShell policy-status and effective-config RPCs before exposing the execution adapter.
- Require `READY`, a `LOADED` revision without a load error, sandbox policy source, matching positive current/active/revision/config versions, structurally identical policies, and matching deterministic hashes.
- Treat a matching effective policy that remains `PENDING` as `policy_status_inconsistent`; never weaken attestation because the sandbox is otherwise executable.
- Use the canonical read-only `/proc` baseline so OpenShell does not enrich the submitted policy and create an unexpected revision.
- Delete and verify deletion of partially created sandboxes after attestation, startup, execution, timeout, failure, or cancellation errors.
- Emit structured, sanitized `sandbox.attestation` and `sandbox.cleanup` events without exception messages, policy contents, credentials, SDK response bodies, or tracebacks.
- Preserve artifact harvest-before-cleanup ordering from `develop`/PR #314. Generated files remain job-scoped, receive durable artifact references, and can be rendered or downloaded after the physical sandbox is deleted.
- Add one writer-local corrective turn when the writer claims `Wrote /shared/output.md` without creating a non-empty output file. A second false completion fails with the stable `writer_output_missing` reason instead of restarting research.
- Split provisioning from gateway lifecycle:
  - `setup_openshell.sh` installs the pinned dependencies, generates policy, and builds the image;
  - `start_openshell_gateway.sh` validates an authenticated registered or packaged gateway and performs a disposable strict readiness probe;
  - AI-Q never launches a raw gateway binary or broadly kills externally owned processes.
- Make the readiness probe verify CLI/SDK/gateway version agreement, request labels, selector behavior, `READY`, `LOADED`, effective policy source/content/hash/version, command execution, deletion, and confirmed absence.
- Move live assertions, resources, and verified teardown into pytest. `scripts/smoke_openshell_isolation.py` is now only a thin compatibility launcher.
- Add `docs/source/deployment/openshell.md` as the canonical operator guide for platform support, ownership, policy/config pairing, startup, acceptance, and troubleshooting.

#### Reviewer Test Instructions

On macOS with Docker Desktop, this is a functional local-demo flow using explicit Landlock `best_effort`:

```bash
gh pr checkout 298
./scripts/setup.sh
source .venv/bin/activate

/opt/homebrew/bin/bash ./scripts/setup_openshell.sh \
  --openshell-version 0.0.80 \
  --local-demo \
  --policy offline

export AIQ_OPENSHELL_GATEWAY_NAME=openshell
export AIQ_OPENSHELL_IMAGE=aiq-openshell-demo:latest
export AIQ_OPENSHELL_POLICY_FILE="$PWD/configs/openshell/generated/aiq-openshell-policy.yaml"
export AIQ_OPENSHELL_EXPECTED_GATEWAY_VERSION=0.0.80
export AIQ_OPENSHELL_REQUIRE_HARD_LANDLOCK=false

nat validate --config_file configs/config_openshell.yml

./scripts/start_openshell_gateway.sh \
  --gateway-name openshell \
  --image-name aiq-openshell-demo:latest \
  --policy-file configs/openshell/generated/aiq-openshell-policy.yaml

./scripts/start_e2e.sh \
  --config_file configs/config_openshell.yml
```

Open `http://localhost:3000`, create two sessions, and start two concurrent deep-research jobs that request a CSV and PNG chart.

While both jobs are running:

```bash
.venv/bin/openshell sandbox list \
  -g openshell \
  --selector aiq=deep-research \
  -o json
```

Expected behavior:

1. Each job creates one distinct physical sandbox when sandbox execution is first needed.
2. Later execution calls from the same job reuse that sandbox.
3. Both sandboxes have `aiq=deep-research` and distinct `aiq-job-id` labels.
4. Attestation succeeds before sandbox execution is exposed to the agent.
5. Generated files appear under the correct job and remain downloadable after cleanup.
6. PNG artifacts render through durable artifact references in the final report.
7. A false writer completion receives one writer-local retry.
8. Cancelling one job deletes only that job's sandbox; the other job remains usable.
9. Completing the remaining job deletes its sandbox.
10. The selector returns no matching sandboxes after both jobs terminate.

Run the automated macOS live suite with:

```bash
.venv/bin/python scripts/smoke_openshell_isolation.py \
  --gateway openshell \
  --policy configs/openshell/generated/aiq-openshell-policy.yaml \
  --image aiq-openshell-demo:latest \
  --expected-gateway-version 0.0.80 \
  --allow-best-effort-landlock
```

Expected result: all three live isolation/attestation, failure-cleanup/redaction, and shared-policy-mismatch tests pass.

A passing macOS run is functional demo evidence only. Production acceptance requires Linux, Docker, and a policy/config pairing that both require Landlock `hard_requirement`, as documented in `docs/source/deployment/openshell.md`.

#### Validation

Current-head scoped regression suite:

```bash
.venv/bin/pytest -q \
  tests/aiq_agent/agents/deep_researcher \
  tests/aiq_agent/jobs/test_runner.py \
  tests/scripts/test_openshell_lifecycle_scripts.py \
  tests/scripts/test_openshell_readiness_checker.py \
  tests/scripts/test_openshell_smoke_wrapper.py
```

Result: **485 passed, 3 skipped in 6.58s**. The three live tests were collected and skipped before optional OpenShell imports or gateway access because live testing was not enabled.

Lint, formatting, and shell syntax:

```bash
.venv/bin/ruff check \
  src/aiq_agent/agents/deep_researcher \
  tests/aiq_agent/agents/deep_researcher \
  tests/aiq_agent/jobs/test_runner.py \
  tests/scripts \
  scripts/check_openshell_readiness.py \
  scripts/smoke_openshell_isolation.py

.venv/bin/ruff format --check \
  src/aiq_agent/agents/deep_researcher \
  tests/aiq_agent/agents/deep_researcher \
  tests/aiq_agent/jobs/test_runner.py \
  tests/scripts \
  scripts/check_openshell_readiness.py \
  scripts/smoke_openshell_isolation.py

bash -n \
  scripts/setup_openshell.sh \
  scripts/start_openshell_gateway.sh \
  scripts/start_e2e.sh
```

Result: **all checks passed; 53 files already formatted; shell syntax passed**.

Configuration validation:

```bash
AIQ_OPENSHELL_REQUIRE_HARD_LANDLOCK=false \
  .venv/bin/nat validate \
  --config_file configs/config_openshell.yml
```

Result: **configuration valid**.

Documentation:

```bash
cd docs
PATH="../.venv/bin:$PATH" make html
```

Result: **build succeeded**. Sphinx reported two existing unresolved links in `docs/source/integration/agent-skills.md`; neither warning originates from the OpenShell documentation.

Live OpenShell validation recorded on macOS/Docker Desktop after the `0.0.80` rebaseline:

- **3 live tests passed**.
- Proved concurrent per-job isolation and selector visibility.
- Proved source/content/hash/revision attestation.
- Proved isolated cancellation and verified terminal deletion.
- Proved failure cleanup and credential-canary redaction.
- Proved shared-policy mismatch rejection.

Linux/Docker acceptance with Landlock `hard_requirement` has **not** been run on this macOS host and remains the production acceptance gate.

1. `src/aiq_agent/agents/deep_researcher/sandbox/providers/openshell.py`
   - Per-job creation, labels, network validation, authoritative attestation, and cleanup.

2. `scripts/check_openshell_readiness.py` and `scripts/start_openshell_gateway.sh`
   - Version/capability validation, selector proof, policy verification, execution, and verified deletion.

3. `tests/aiq_agent/agents/deep_researcher/sandbox/test_openshell_live.py`
   - Pytest-owned live isolation, cancellation, cleanup/redaction, and shared-policy rejection.

4. `src/aiq_agent/agents/deep_researcher/custom_middleware.py`
   - Deterministic filesystem arguments, sandbox-path validation, and the bounded writer-output guard.

5. `docs/source/deployment/openshell.md`
   - Canonical supported-platform, ownership, provisioning, startup, acceptance, and troubleshooting contract.

#### Related Issues

- Relates to #287
- AIQ-3531
- Partial AIQ-3433 lifecycle coverage
- Upstream issue: [NVIDIA/OpenShell#2159](https://github.com/NVIDIA/OpenShell/issues/2159)
- Upstream fix released in OpenShell `0.0.80`: [NVIDIA/OpenShell#2170](https://github.com/NVIDIA/OpenShell/pull/2170)

## Summary by CodeRabbit

* **New Features**
  * Experimental per-job OpenShell sandboxes with policy binding, attestation, and network allowlist support; debug shared-sandbox attachment is explicitly opt-in.
  * Added OpenShell lifecycle tooling (setup, gateway startup, readiness/version checks, smoke/live acceptance).
  * Strengthened chart-generation/artifact workflow with runtime-driven chart execution and output validation.
* **Bug Fixes**
  * Improved deep-research job cancellation/terminal teardown reliability with safer event store flushing and sandbox finalization.
  * Tightened fail-closed policy/network handling and reduced sensitive error text leakage in logs and events.
* **Documentation**
  * Updated deep-research sandboxing/ownership architecture guidance, OpenShell deployment/operator docs, and REST SSE artifact event details.


tests/aiq_agent/agents/deep_researcher/sandbox/test_openshell_provider.py
```diff
@@ -63,18 +65,382 @@ def __exit__(self, *_args: object) -> None:
...
+    ):
+        provider = OpenShellSandboxProvider(
+            SandboxConfig(provider="openshell", providers={"openshell": {"policy": str(policy_path)}}),
+            "job-123",
+        )
+        with pytest.raises(ValueError, match="bad adapter"):
+            provider._create_session()
+
+    assert context.exit_calls == 1
+    assert provider._os_context is None
```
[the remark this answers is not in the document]
KyleZheng1284: Retaining the cumulative failure is intentional fail-closed behavior. `cleanup_succeeded` represents every owned-sandbox cleanup attempt during the job, including stale-session teardown during retry/reset. If an earlier deletion fails, a later successful final close cannot prove that every owned sandbox was deleted, so terminal cleanup must remain `failed`. Commit `8e2287b` documents this contract and `test_retry_cleanup_failure_remains_terminal_failure` proves the earlier failure remains observable after the final close succeeds. Validation on the exact commit: OpenShell provider 35 passed; deterministic remediation suite 239 passed; full PR-scoped suite 293 passed.

KyleZheng1284:
Final CodeRabbit remediation is in `8e2287b` (single DCO-signed commit). It serializes and caches terminal finalization, hardens OpenShell context ownership across creation/termination races, delays attestation success until adapter usability, adds owned/non-owned NAT cancellation coverage, and retains cumulative fail-closed cleanup status. Follow-up findings are also addressed: shared-sandbox aliases must agree, cleanup warnings contain exception types only, gateway identifiers are excluded from attestation logs, shared attachments never delete unowned sandboxes, false cleanup results remain observable, event-store flush failures cannot mask job outcomes, and public shared attachment requires explicit debug opt-in. Validation on the exact commit: provider 35 passed; deterministic remediation suite 239 passed; full PR-scoped suite 293 passed; Ruff check/format passed; config and shell validation passed; OpenShell 0.0.72 live smoke passed distinct creation, cancellation isolation, continued peer usability, and terminal deletion.

src/aiq_agent/agents/deep_researcher/sandbox/providers/openshell.py
```diff
@@ -241,36 +369,153 @@ def _create_session(self) -> BaseSandbox:
...
+            self._ensure_context_active(os_sandbox)
+            backend = OpenShellSandbox(sandbox=os_sandbox, timeout=cfg.timeout, shell=oscfg.shell)
+            self._ensure_context_active(os_sandbox)
+            sandbox_ref = os_sandbox.sandbox
+            logger.info(
+                "OpenShell sandbox attested: id=%s name=%s gateway=%s policy_version=%s shared=%s",
+                backend.id,
+                getattr(sandbox_ref, "name", None),
+                oscfg.gateway,
+                policy_version,
```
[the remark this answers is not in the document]
KyleZheng1284: Addressed in `8e2287b`: the attestation info log no longer includes `oscfg.gateway`. The per-job creation test now configures a sentinel gateway name and asserts it is absent from captured logs. Validation: OpenShell provider 35 passed; deterministic remediation suite 239 passed; full PR-scoped suite 293 passed.

src/aiq_agent/agents/deep_researcher/deepagents_runtime.py
```diff
@@ -229,6 +278,50 @@ def terminate(self) -> None:
...
+                return self._finalize_result
+
+            self._emit_cleanup("started", interrupted=interrupted)
+            try:
+                if interrupted:
+                    self.terminate()
+                else:
+                    self.close()
+            except Exception:  # noqa: BLE001 - terminal cleanup cannot replace the job result
+                logger.warning("Sandbox cleanup failed for job %s", self._job_id, exc_info=True)
```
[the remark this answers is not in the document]
KyleZheng1284: Addressed in `8e2287b`: terminal cleanup and cleanup-event warning paths now log only the exception class, without `exc_info` or raw exception text. `test_finalize_logs_only_cleanup_exception_type` proves a credential-like exception message is absent while the cleanup result remains false. Validation: deterministic remediation suite 239 passed; full PR-scoped suite 293 passed.

src/aiq_agent/agents/deep_researcher/deepagents_runtime.py
```diff
@@ -106,19 +133,35 @@ class DeepResearchSandboxConfig(FunctionBaseConfig, name="deep_research_sandbox"
...
     )
 
+    @model_validator(mode="after")
+    def _validate_openshell_and_network(self) -> DeepResearchSandboxConfig:
+        if self.network == "allowlist" and not self.network_allow:
+            raise ValueError("network='allowlist' requires a non-empty network_allow list")
+        if self.provider == "openshell":
+            shared_name = self.existing_sandbox_name or self.sandbox_name
+            if shared_name and not self.allow_shared_sandbox:
+                raise ValueError("existing_sandbox_name/sandbox_name requires allow_shared_sandbox=true")
```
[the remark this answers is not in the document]
KyleZheng1284: Addressed in `8e2287b`: `DeepResearchSandboxConfig` now rejects differing `existing_sandbox_name` / deprecated `sandbox_name` values before any shared attachment is selected. `test_public_shared_attachment_rejects_conflicting_aliases` covers the fail-closed behavior. Validation: deterministic remediation suite 239 passed; full PR-scoped suite 293 passed.

KyleZheng1284:
@coderabbitai review

scripts/smoke_openshell_isolation.py
```diff
@@ -0,0 +1,197 @@
...
+from aiq_agent.agents.deep_researcher.sandbox.providers.openshell import _policy_network_hosts
+from aiq_agent.agents.deep_researcher.sandbox.providers.openshell import _read_policy_data
+
+
+def _args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--gateway", help="Registered OpenShell gateway name; default uses the active gateway")
+    parser.add_argument(
+        "--policy",
+        default="configs/openshell/aiq-research-policy.yaml",
```
AjayThorve: The default smoke path points at `configs/openshell/aiq-research-policy.yaml`, but that checked-in policy still contains `/app`. The latest commit explains that `/app` does not exist in the demo image and makes `hard_requirement` fail during sandbox prepare; it only removed the path from the generator. Can we remove `/app` from the checked-in policy too, or point this default at the generated policy? Otherwise the no-args smoke test still hits the exact failure this branch just fixed.

src/aiq_agent/agents/deep_researcher/deepagents_runtime.py
```diff
@@ -229,6 +284,50 @@ def terminate(self) -> None:
...
             provider.terminate()
 
+    def finalize(self, *, interrupted: bool) -> bool:
+        """Release the sandbox once and emit a truthful, sanitized cleanup outcome."""
+        with self._finalize_lock:
+            if self._finalized:
+                assert self._finalize_result is not None
+                return self._finalize_result
+
+            self._emit_cleanup("started", interrupted=interrupted)
```
AjayThorve, line 359: `finalize()` emits sandbox cleanup events even when this runtime has no sandbox provider. The API runner finalizes every Deep Research runtime, so regular non-sandbox jobs now get `sandbox.cleanup` started/succeeded events with `provider: null` and `sandbox: null`. Can we short-circuit the event emission when `_sandbox_provider is None`? These events should represent a real sandbox lifecycle.

tests/aiq_agent/jobs/test_runner.py
```diff
@@ -1908,6 +1908,41 @@ def test_none_runtime_is_noop(self):
...
+        from aiq_api.jobs.runner import _teardown_sandbox
+
+        runtime = MagicMock(spec=["finalize"])
+        runtime.finalize.return_value = False
+
+        with caplog.at_level("WARNING", logger="aiq_api.jobs.runner"):
+            _teardown_sandbox(runtime, job_id="job-1", interrupted=False)
+
+        assert "Sandbox cleanup reported failure for job job-1" in caplog.text
+
```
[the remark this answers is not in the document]
KyleZheng1284: Addressed in `9cec3b4`: `_teardown_sandbox` now sanitizes exceptions from both request-owned `finalize()` and the legacy close/terminate fallback by logging only the exception class. `test_runtime_finalizer_exception_is_nonfatal_and_sanitized` proves a credential-like exception message is absent, and the fallback regression now enforces the same contract. Validation: `tests/aiq_agent/jobs/test_runner.py` — 98 passed; Ruff check/format passed; CodeRabbit and DCO are green on the pushed revision.

src/aiq_agent/agents/deep_researcher/sandbox/providers/openshell.py
```diff
@@ -125,21 +129,148 @@ def _is_openshell_not_found_error(exc: Exception) -> bool:
...
+        return set()
+    hosts: set[str] = set()
+    for policy in policies.values():
+        if not isinstance(policy, dict):
+            continue
+        endpoints = policy.get("endpoints") or []
+        if not isinstance(endpoints, list):
+            continue
+        for endpoint in endpoints:
+            if isinstance(endpoint, dict) and isinstance(endpoint.get("host"), str):
```
AjayThorve, line 248: `_policy_network_hosts()` only accounts for `host`, but OpenShell also permits hostless endpoints with `allowed_ips`. A policy with `allowed_ips: [0.0.0.0/0]` currently passes both `network="blocked"` and an unrelated hostname allowlist, while OpenShell parses it as `host=""` and permits any domain resolving into that range. Can we reject hostless/CIDR endpoints unless the public config grows an explicit CIDR contract, and add blocked/allowlist regressions? Otherwise this upper-bound check is bypassable.

src/aiq_agent/agents/deep_researcher/sandbox/providers/openshell.py
```diff
@@ -241,36 +369,152 @@ def _create_session(self) -> BaseSandbox:
...
+
+        sandbox_ref = os_sandbox.sandbox
+        phase = getattr(sandbox_ref, "phase", None)
+        policy_version = getattr(sandbox_ref, "current_policy_version", 0)
+        if phase != openshell_pb2.SANDBOX_PHASE_READY:
+            self._emit_attestation(phase=phase, policy_version=policy_version, status="failed")
+            raise RuntimeError(f"OpenShell sandbox attestation failed: phase={phase!r} is not READY")
+
+        oscfg = self.config.providers.openshell
+        if oscfg.attest and (not isinstance(policy_version, int) or policy_version <= 0):
```
AjayThorve: A positive `current_policy_version` proves that some revision loaded, not that the policy we validated and submitted is the effective policy. This path skips the status/config RPC entirely when the version is already positive, and OpenShell can replace the sandbox policy with a gateway-global policy. Can we attest the effective policy source/hash against the submitted proto and require the reported `active_version` to match the loaded revision before emitting success? Version alone is not a sufficient security invariant.

src/aiq_agent/agents/deep_researcher/sandbox/providers/openshell.py
```diff
@@ -282,14 +526,78 @@ def _terminate_session(self, session: BaseSandbox | None) -> None:
...
-        """Exit the OpenShell context once, swallowing cleanup errors on the terminal path."""
-        ctx = self._os_context
-        self._os_context = None
-        if ctx is not None and hasattr(ctx, "__exit__"):
+        """Exit once, deferring deletion to the creator while ``__enter__`` is in flight."""
+        with self._state_lock:
+            ctx = self._os_context
+            if ctx is None:
+                return
+            if self._os_context_entering:
```
AjayThorve: When cancellation lands during `ctx.__enter__()`, this path records a deferred exit and returns immediately. I reproduced `finalize(interrupted=True)` returning `True` and emitting `sandbox.cleanup=succeeded` while creation was still running and `__exit__()` had not executed; a later exit failure cannot change the cached result. Can we track deferred teardown with a completion primitive and make finalization wait boundedly for the actual exit outcome, reporting failure on timeout or cleanup error?

scripts/setup_openshell.sh
```diff
@@ -57,7 +58,7 @@ GATEWAY_ENDPOINT=""
 GATEWAY_DISABLE_TLS=false
 RESTART_GATEWAY=true
 BUILD_IMAGE=true
-CREATE_SANDBOX=true
+CREATE_SANDBOX=false
```
AjayThorve: Changing `CREATE_SANDBOX` to `false` also disables the raw-gateway rejection in `configure_gateway_mode()`, because that guard is conditional on this flag. The script can now start plaintext `openshell-gateway`, report setup success, and defer the known JWT/mTLS Docker-creation failure to the first AI-Q job. Can we make the gateway capability guard independent of shared-sandbox creation, ideally with a create/delete readiness probe before setup reports success?

src/aiq_agent/agents/deep_researcher/sandbox/providers/openshell.py
```diff
@@ -282,14 +526,78 @@ def _terminate_session(self, session: BaseSandbox | None) -> None:
...
+            self._os_context_exit_requested = False
+        self._close_os_context(ctx)
+
+    def _close_os_context(self, ctx: object) -> None:
+        """Drive one detached SDK context exit without replacing the job result."""
+        if hasattr(ctx, "__exit__"):
             try:
                 ctx.__exit__(None, None, None)
             except Exception:  # noqa: BLE001 - cleanup must never raise on the terminal path
+                self._cleanup_failed = True
```
AjayThorve: This provider-layer cleanup path still logs `exc_info=True` on the next line, so the full SDK exception and traceback bypass the sanitized runner/runtime logging. I reproduced a credential-like value appearing in the log; the same pattern remains in `sandbox/base.py` for event delivery and session cleanup. Can we log only classified exception metadata here and in those base paths, or route all of them through one redactor?

AjayThorve (commented):
I found five remaining blockers on the current head.

AjayThorve:
Can we split durable OpenShell provisioning from gateway lifecycle before merging? `setup_openshell.sh` is documented as a run-once setup, but it currently installs dependencies, builds the policy/image, and also `pkill`s and restarts a long-lived gateway. I’d keep setup idempotent and limited to install/configure/build, then add `scripts/start_openshell_gateway.sh` to start or reuse and register the gateway, with an actual create/delete probe as the readiness check. `start_e2e.sh` can optionally invoke that thin launcher; per-job sandbox creation should remain runtime-owned in Python. This gives the gateway one explicit lifecycle owner and prevents setup from reporting success for a service that cannot create the required sandbox.

src/aiq_agent/agents/deep_researcher/sandbox/providers/openshell.py
KyleZheng1284: Addressed in `ef823fb`. The public network-policy boundary now rejects every endpoint in blocked mode; allowlist mode requires a non-empty normalized host, requires each host to be in `network.allow`, and rejects every non-empty `allowed_ips` entry because AI-Q exposes no IP/CIDR exception contract. Regressions cover hostless endpoints, empty hosts, `0.0.0.0/0`, IPv4/IPv6 CIDRs, and mixed host/IP forms. Post-merge focused suite: 139 passed. Leaving this thread open until required CI runs.

src/aiq_agent/agents/deep_researcher/sandbox/providers/openshell.py
KyleZheng1284: Addressed in `ef823fb`. Attestation now always calls both `GetSandboxPolicyStatus` and `GetSandboxConfig`, including when `current_policy_version` is positive. It requires READY + LOADED/no error, matching active/revision/config/current/expected versions, per-job sandbox policy source, structurally identical submitted/revision/effective policies, and matching deterministic OpenShell 0.0.72 hashes. Missing RPCs or any source/content/hash/version mismatch fail closed and delete the partial sandbox. Regressions cover every mismatch plus zero-version 0.0.72 compatibility and strict/reduced shared-debug modes. Leaving this thread open until required CI runs.

src/aiq_agent/agents/deep_researcher/sandbox/providers/openshell.py
KyleZheng1284: Addressed in `15d3f73`. In-flight context entry now has a generation-specific completion event. Cancellation requests exactly one deferred exit and waits outside locks up to `cleanup_timeout_seconds`; exit failure or timeout becomes a monotonic terminal cleanup failure, so a late successful exit cannot overwrite it. Concurrent finalizers share one cached real result. Regressions cover cancellation during enter, exit failure, timeout, duplicate termination, and concurrent finalization. Leaving this thread open until required CI runs.

scripts/setup_openshell.sh
KyleZheng1284: Addressed in `fc96a86`. `setup_openshell.sh` is provisioning-only and rejects legacy lifecycle flags. The new `start_openshell_gateway.sh` rejects raw/plaintext/insecure gateways, accepts only authenticated registered gateways or packaged Homebrew/systemd services, runs status, and requires a disposable create/readiness/delete probe with trap cleanup. Stub-executable regressions prove setup performs no lifecycle operations, invalid gateways fail, the probe is mandatory, and deletion is attempted after probe failure. Leaving this thread open until required CI runs.

src/aiq_agent/agents/deep_researcher/sandbox/providers/openshell.py
KyleZheng1284: Addressed in `15d3f73`. Sandbox runtime paths now use centralized sanitized exception logging: operation, safe provider/sandbox identifiers, reason code, and exception class only—no `str(exc)`, traceback, commands, policy/endpoint/environment values, credentials, or SDK response bodies. The audit covers provider/context/session cleanup, event delivery, workspace prep, registry loading, artifact scan/store, and terminal harvest. Secret-canary regressions assert raw values do not reach logs or events. Leaving this thread open until required CI runs.

KyleZheng1284:
Lifecycle follow-up: commit `fc96a86` now makes `setup_openshell.sh` provisioning-only (pinned dependencies, policy generation/validation, image build). It cannot start/stop/kill/register/select a gateway or create a persistent debug sandbox, and legacy lifecycle flags fail with a migration message. Gateway ownership moved to `start_openshell_gateway.sh`, which accepts only an authenticated registered gateway or the official packaged Homebrew/systemd service, runs `openshell status`, and performs a mandatory disposable create/readiness/delete probe with trap cleanup. `start_e2e.sh` invokes it only with `--start-openshell-gateway` and never stops externally managed services. The stub lifecycle suite proves the auth rejection paths, mandatory probe, cleanup after probe failure, setup non-interference, and explicit E2E opt-in (6 passed in the post-merge run). The real Linux OpenShell 0.0.72 + Landlock `hard_requirement` probe is still pending on a suitable host; I am not substituting macOS `best_effort` output for that acceptance evidence.

src/aiq_agent/agents/deep_researcher/sandbox/providers/openshell.py
KyleZheng1284: Live validation follow-up in `7baf293`: the authenticated OpenShell 0.0.72 gateway reports `current_policy_version=0` and `active_version=0` for an initial per-sandbox policy, while the LOADED revision and effective config both report version 1 with identical sandbox source, submitted protobuf content, gateway hashes, and the ported deterministic hash. A literal active==revision rule therefore blocks every real 0.0.72 job. The compatibility is now bound to gateway health version exactly `0.0.72`, requires both generic/active values to be zero, and still requires the positive revision/config pair plus source/content/hash to agree. Any positive value, later gateway version, or other mismatch fails closed. New regressions cover accepted 0.0.72 zero-active behavior and rejection for 0.0.71/0.0.73/positive-current disagreement; the expanded live smoke passed all seven assertions. Leaving this thread open for refreshed CI.

scripts/smoke_openshell_isolation.py
```diff
@@ -0,0 +1,328 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Live, deterministic proof of OpenShell per-job isolation and cleanup.
```
AjayThorve: This is a 300+ line live acceptance test: it imports internal provider helpers, creates real resources, asserts security and lifecycle behavior, and owns cleanup. Keeping that logic under `scripts/` makes it look like an operator utility and leaves it outside pytest discovery, markers, and fixtures. Can we move the test implementation to `tests/aiq_agent/agents/deep_researcher/sandbox/test_openshell_live.py`, mark it `integration`, and gate it with `AIQ_OPENSHELL_LIVE_TESTS=1`, following the existing OpenSearch live-test pattern? If the one-command entry point is useful, keep a thin wrapper under `scripts/`, but the assertions and cleanup fixtures should be test-owned.

AjayThorve (commented):
One test-ownership concern on the current head.

AjayThorve:
Can we add a canonical OpenShell setup and deployment guide under `docs/source`—for example, `docs/source/deployment/openshell.md`—and have the existing script, sandbox, and architecture docs link to it? The current information is split across `scripts/README.md`, the sandbox implementation README, the architecture page, and config comments. The guide should provide one coherent operator contract: supported-platform matrix; macOS local-demo versus Linux production-acceptance flows; gateway installation and version compatibility; Docker/Podman and service ownership; Landlock policy/config pairing; provisioning versus startup responsibilities; deterministic live acceptance; and cleanup/troubleshooting. The architecture page can stay focused on invariants and `scripts/README.md` on command discovery. This feature introduces a real external runtime and security boundary, so it deserves a first-class user guide rather than requiring users to assemble the workflow from implementation notes.

scripts/smoke_openshell_isolation.py
KyleZheng1284: Addressed in `dc9839c`.

All live assertions and resource ownership now live in `tests/aiq_agent/agents/deep_researcher/sandbox/test_openshell_live.py`. The module is marked `integration`, is gated by `AIQ_OPENSHELL_LIVE_TESTS=1`, and does not import the optional OpenShell SDK or contact a gateway before the gate. Fixtures own configuration, provider creation, direct sandbox creation, immediate resource registration, reverse-order teardown, and gateway-verified deletion; teardown verification fails the test even if the test body also failed.

`scripts/smoke_openshell_isolation.py` is now only a `subprocess.run()` launcher that maps the compatibility flags to the documented environment variables, targets the single pytest module, and propagates pytest's exit code. `tests/scripts/test_openshell_smoke_wrapper.py` covers argument/environment mapping, best-effort opt-in, target selection, and exit-code propagation.

Local result without opt-in: **10 passed, 3 live tests collected/skipped**, with no gateway access. The broader OpenShell regression set is **292 passed, 2 skipped**. The required Linux/OpenShell 0.0.72/hard-Landlock live run is still pending, so I am intentionally leaving this thread open until that run and refreshed CI are green.

KyleZheng1284:
Addressed in `450df92` with the canonical guide at `docs/source/deployment/openshell.md`.

The guide now owns the operator contract: security boundary; Linux/Docker, macOS demo, Podman evaluation, remote gateway, and Windows/WSL support matrix; pinned CLI/SDK/gateway/adapter compatibility; lifecycle/service ownership; exact Linux, macOS, and remote flows; policy/config and Landlock pairing; expected per-job behavior; pytest-owned acceptance; sanitized inspection; and safe cleanup/troubleshooting.

One-way references were added from `docs/source/index.md`, `docs/source/deployment/index.md`, `docs/source/deployment/production.md`, `docs/source/resources/troubleshooting.md`, `docs/source/architecture/agents/sandbox.md`, `scripts/README.md`, the sandbox implementation README, `configs/config_openshell.yml`, and setup/launcher help. Operational duplication was removed from the script and implementation READMEs.

The normal Sphinx build succeeds and the new guide/backlinks render without warnings. The requested warning-as-error build completes but remains nonzero on two pre-existing `docs/source/integration/agent-skills.md` links to `.agents/skills/README`; neither warning is in an OpenShell-touched source. I am leaving the review item open until refreshed CI and the required Linux/OpenShell 0.0.72/hard-Landlock live suite are green.

KyleZheng1284:
## BLOCKED — upstream OpenShell policy state and SDK capabilities

Marking this PR blocked for the AI-Q 2.2 OpenShell integration.

The policy-status failure has now been reproduced **without AI-Q** by creating a sandbox directly through the OpenShell 0.0.77 CLI on macOS/Docker Desktop:

- the sandbox reaches `Ready`;
- `policy get` reports revision 2 as `Effective`;
- `policy list` leaves revision 2 `Pending` while revision 1 is `Superseded`.

AI-Q therefore correctly times out instead of exposing the execution adapter without authoritative `LOADED` status. This is not explained by `best_effort`: that setting changes Landlock compatibility, not the policy-revision acknowledgement contract. The direct CLI reproduction also removes AI-Q's sandbox-creation path from the failure.

The upstream lifecycle/API work is tracked in [NVIDIA/OpenShell#2159](https://github.com/NVIDIA/OpenShell/issues/2159). That issue also tracks the Python SDK gap for request-level labels and selector-based listing; template/container labels alone do not make `openshell sandbox list --selector aiq=deep-research` work against gateway metadata.

### Unblock criteria

1. OpenShell reports an initially applied policy as `LOADED` (or `FAILED`) with an authoritative version, or documents another security-equivalent success contract.
2. The public Python SDK supports request-level labels, selector filtering, and returned gateway metadata labels.
3. The fixes are available in a tagged OpenShell release with compatible CLI, SDK, and gateway versions.
4. AI-Q pins that release and passes the strict readiness probe.
5. The pytest-owned live suite passes on macOS/Docker Desktop with explicit `best_effort` as demo validation and on Linux/Docker with `hard_requirement` as production acceptance.

Until those gates are met, this PR must not claim production acceptance or merge by weakening attestation, treating `Pending` as success, or pinning an unreleased fork. AI-Q-side fail-fast diagnostics, cleanup verification, tests, and documentation can continue while the upstream dependency is resolved.

Current scope note: the policy-state contradiction is confirmed on macOS with OpenShell 0.0.77; a minimal Linux reproduction is still required to determine whether the runtime manifestation is driver/platform-specific.

KyleZheng1284:
Matched-stack macOS validation after `cdc3891` and `7b0ca2f`:

- AI-Q head: `7b0ca2f3d5d806a4f5b173bc546b1150bb149f61`
- OpenShell source: `616ff2f6` (`fix/2159-initial-policy-ack-and-sdk-labels/KyleZheng1284`)
- SDK wheel: `0.0.79.dev3+g616ff2f6`
- CLI/gateway: `0.0.79-dev.3+g616ff2f6`
- Platform: macOS ARM64 + Docker Desktop, explicit `best_effort` Landlock demo mode
- Policy: generated AI-Q policy with the canonical read-only `/proc` proxy baseline

The strict disposable readiness probe passed version equivalence, request labels/selectors, READY state, authoritative `LOADED` policy status, source/content/hash/version agreement, execution, and verified deletion.

The pytest-owned live suite then passed all three tests:

```text
test_live_per_job_isolation_attestation_and_cancellation PASSED
test_live_failure_cleanup_and_log_redaction PASSED
test_live_shared_policy_mismatch_is_rejected PASSED
3 passed in 8.78s
```

Additional validation:

```text
Scoped OpenShell/runtime suite: 306 passed
Full repository suite: 1455 passed, 9 skipped
Ruff: all checks passed; 310 files already formatted
Both configs/config_openshell.yml and the untracked local config copy validate
Sphinx HTML build succeeded
All disposable OpenShell containers were removed
```

The original permanent `PENDING` problem is fixed by the branch-built supervisor. AI-Q now also generates read-only `/proc` rather than `/proc/self`, preventing OpenShell from creating an enriched revision that would correctly fail exact content/hash attestation.

This is development evidence only. The `BLOCKED` label should remain until the OpenShell fix is merged and published as a matched release and the Linux `hard_requirement` live suite passes.

cdgamarose-nv:
If i run `./scripts/setup.sh` followed by `./scripts/setup_openshell.sh`, it will reinstall a bunch of packages with newer versions. Can we not just keep the versions in the two files in sync? And also not duplicate packages unless they're intended to be in different virtual environments?

scripts/openshell/check_openshell_readiness.py
cdgamarose-nv, line 1: Can we move all under an openshell directory? There are a ton of scripts

src/aiq_agent/agents/deep_researcher/sandbox/providers/openshell.py
```diff
@@ -119,27 +128,239 @@ def _normalize_openshell_name(job_id: str, prefix: str = "aiq-deep-research") ->
...
+
+def _has_message_field(message: Any, field: str) -> bool:
+    """Return protobuf message-field presence without assuming a concrete SDK class."""
+    try:
+        return bool(message.HasField(field))
+    except (AttributeError, ValueError):
+        return getattr(message, field, None) is not None
+
+
+def _deterministic_policy_hash(policy: Any) -> str:
```
cdgamarose-nv: Is it safe that AI-Q is reinventing the recipe to compute the hash? It seems like there could be chances of it drifting from the openshell one, and keeping it up to date would be a lot of maintenance work

scripts/openshell/check_openshell_readiness.py
KyleZheng1284: yes done

KyleZheng1284:
> If i run `./scripts/setup.sh` followed by `./scripts/setup_openshell.sh`, it will reinstall a bunch of packages with newer versions. Can we not just keep the versions in the two files in sync? And also not duplicate packages unless they're intended to be in different virtual environments?

there are some deps mismatches since the published langchain-openshell pkg `0.1.0` still declares deepagents < 0.6 while AIQ is currently 0.6.8. I removed the forced reinstall and now restore AI-Q’s locked dependencies after installing OpenShell; we can switch to a single uv sync --extra openshell once the adapter supports DeepAgents 0.6.x.

src/aiq_agent/agents/deep_researcher/sandbox/providers/openshell.py
KyleZheng1284: Removed the custom hash calculation and now validating that OpenShell’s effective and revision policies match the submitted policy and report the same non-empty hash. 

Missing or mismatched hashes fail closed; if we need local hash verification later we can look into using an official OpenShell SDK helper

KyleZheng1284:
@coderabbitai full review

KyleZheng1284:
@coderabbitai review

AjayThorve (approved):
Looks good, thanks for this

PR #299, opened 2026-07-02
skills-eval: add nightly scheduled trigger

sarathm-nvidia:
AI-Q's skills-eval workflow (`.github/workflows/skills-eval.yml`) currently triggers **only** on `push` (develop / release/** / pull-request/<N>) and `workflow_dispatch`. It has no automated cadence, so it never runs on its own — unlike **VSS** (`skills-eval-daily.yml`) and **RAG** (`skills-eval.yml`), both of which run a nightly full sweep. This means AI-Q eval results only reach the skills dashboard when someone pushes or manually dispatches.

## Change

Add a `schedule:` cron nightly trigger mirroring the RAG/VSS convention, so a full harbor sweep runs automatically each night and results flow to the skills dashboard.

```yaml
schedule:
  - cron: "0 2 * * *" # 2am UTC nightly — full sweep of all skills
```

(2am UTC matches RAG's `skills-eval.yml` nightly.)

A scheduled run carries no diff, so it is gated like a manual dispatch to guarantee a full sweep (never path-gated to zero):

- **detect-changes** short-circuits to `relevant=true` on `schedule` (same as `workflow_dispatch`), bypassing the path gate.
- **harbor-eval** `if:` now includes `github.event_name == 'schedule'` so the harbor trial runs automatically on the nightly.
- **SCOPE** selectors are unchanged: on a scheduled run `github.ref_name` is `develop` (not `pull-request/<N>`), so the existing default `*) SCOPE="--all"` case already produces a full sweep.

Existing `push` / `workflow_dispatch` behavior is **not** altered — this only adds the schedule trigger plus the minimal gating needed for it to run a full sweep.

## Validation

- YAML parses (`yaml.safe_load`).
- Diff is limited to the schedule trigger, the two gating changes, and comment updates.


.github/workflows/skills-eval.yml
```diff
@@ -5,7 +5,9 @@
 #   - generate-datasets: spec validation, no credentials.
 #   - harbor-eval:       full Harbor trial; auto on push, or dispatch w/ run_harbor.
 # Push-triggered on develop/release/pull-request/<N> (copy-pr-bot mirror after a
-# vetter's `/ok to test`); workflow_dispatch for manual full sweeps.
+# vetter's `/ok to test`); workflow_dispatch for manual full sweeps; schedule for
+# a nightly full sweep (mirrors the RAG/VSS nightly convention so eval results
```
cdgamarose-nv, line 9: can you remove the comment about RAG and VSS please

PR #306, opened 2026-07-02
fix(deep-research): source-routing guard recognizes sandbox 

KyleZheng1284:
#### Overview

Sandboxed deep-research runs could get stuck: the orchestrator was rejected on every tool call with "Source routing is required before any other tool call," so it never made progress.

Root cause seems to be a key-name mismatch:
- The source-routing guard decides it's safe to proceed by checking `state["files"]`
  for `/shared/source_routing.json`.
- But when a sandbox provider is enabled, `CompositeBackend` drops the `/shared/`
  prefix before storing the file, so it actually lands under `/source_routing.json`.
- The guard never found its expected key, so it kept blocking. (Non-sandbox runs
  keep the `/shared/` prefix, so they were fine — this only hit sandbox runs.)

Fix: the guard now accepts **both** key forms (`/shared/source_routing.json` and the
sandbox `/source_routing.json`). Small, no behavior change for non-sandbox runs.

#### Validation

- `pytest test_custom_middleware.py test_factory.py` → **52 passed**, including a new
  test that seeds the sandbox key `/source_routing.json` and confirms the guard opens.
- `ruff check` on changed files → clean.

- [ ] I updated documentation for user-facing or contributor-facing changes.

'custom_middleware.py` → `SourceRoutingGuardMiddleware._routing_complete`. Context for
why the key gets stripped is in `deepagents_runtime.py::_normalize_state_files`.

#### Related Issues

- Relates to #

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved source-routing detection so the app recognizes routing completion from either the shared or sandbox-local routing file.
  * Normal tool actions now proceed correctly in sandbox environments once routing data is present.
* **Tests**
  * Added coverage for sandbox-based routing to confirm the expected behavior and response flow.

### DIFF
--- a/src/aiq_agent/agents/deep_researcher/custom_middleware.py
+++ b/src/aiq_agent/agents/deep_researcher/custom_middleware.py
@@ -38,6 +38,11 @@ logger = logging.getLogger(__name__)
 # Path to this agent's prompts directory
 _PROMPTS_DIR = Path(__file__).parent / "prompts"
 _SOURCE_ROUTING_PATH = "/shared/source_routing.json"
+# When a sandbox provider is configured, CompositeBackend strips the /shared/ route
+# before delegating to StateBackend, so the router's file is stored under the
+# route-local key. The guard reads raw state, so it must accept both forms or it
+# blocks the orchestrator forever on sandboxed runs.
+_SOURCE_ROUTING_STATE_KEYS = (_SOURCE_ROUTING_PATH, "/source_routing.json")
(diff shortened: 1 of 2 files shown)


soumilinandi (approved):
tested [PR](https://github.com/NVIDIA-AI-Blueprints/aiq/pull/300) with this, LGTM!

AjayThorve (approved):

PR #308, opened 2026-07-03
feat(knowledge): add Azure AI Search backend

harmke:
#### Overview

Adds Azure AI Search as a Knowledge Layer backend for AI-Q. This gives deployments a managed alternative to the existing LlamaIndex/ChromaDB and Foundational RAG backends while preserving the shared Knowledge API, NAT function registration, document summaries, and citation formatting.

The adapter generates embeddings client-side through the existing NVIDIA embedding integration and stores document chunks in one Azure AI Search index per logical collection.

This PR includes:

- `azure_ai_search` retriever and ingestor adapters with vector, hybrid, and opt-in semantic-ranked retrieval.
- API-key authentication through `AZURE_SEARCH_API_KEY` or Azure identity authentication through `DefaultAzureCredential`.
- Managed-identity setup guidance for the required `Search Service Contributor` and `Search Index Data Contributor` roles.
- Shared embedding defaults through `AIQ_EMBED_BASE_URL` and `AIQ_EMBED_MODEL`, with Azure index dimensions configured through `AIQ_EMBED_DIM`.
- Collision-safe physical index names and ownership/schema markers so AI-Q only lists or mutates indexes it owns.
- PDF, DOCX, TXT, and Markdown ingestion through LlamaIndex readers, including chunk metadata, page-aware citations, and optional summaries.
- Canonical UUID file IDs, asynchronous ingestion status, same-name replacement handling, batch-size enforcement, upload rollback, delete retries, eventual-consistency-aware validation, pagination, and TTL cleanup.
- Azure-specific installation, configuration, and operating documentation.
- Focused unit tests and adapter compliance coverage.

This PR does not provision Azure infrastructure. The Azure AI Search service, required permissions, and embedding endpoint must already exist.

#### Validation

All commands were run from the repository root unless noted.

```text
$ uv run pytest tests/knowledge_layer_tests/test_azure_ai_search.py -q
..................................                                       [100%]
34 passed in 12.96s

$ uv run ruff check sources/knowledge_layer/src/register.py sources/knowledge_layer/src/azure_ai_search tests/knowledge_layer_tests/test_azure_ai_search.py tests/knowledge_layer_tests/run_adapter_compliance.py
All checks passed!

$ uv run ruff format --check sources/knowledge_layer/src/register.py sources/knowledge_layer/src/azure_ai_search tests/knowledge_layer_tests/test_azure_ai_search.py tests/knowledge_layer_tests/run_adapter_compliance.py
5 files already formatted
```

Adapter registration and factory compliance:

```text
$ uv run python tests/knowledge_layer_tests/run_adapter_compliance.py --backend azure_ai_search --quick --config '{"endpoint":"https://example.search.windows.net","start_ttl_cleanup":false}'
Summary: 4/4 passed, 0 failed
All tests PASSED - Adapter is compliant!
```

Documentation build:

```text
$ cd docs
$ uv run make html
build succeeded, 3 warnings.
```

The three Sphinx warnings are unresolved cross-reference warnings in unrelated existing pages (`architecture/agents/sandbox.md` and `integration/agent-skills.md`).

Live Azure service validation was also completed against a Basic-tier Azure AI Search service using NVIDIA-hosted `nvidia/nv-embed-v1` embeddings (4096 dimensions). The full adapter compliance flow passed `14/14` under admin-key authentication, covering collection creation, TXT ingestion, asynchronous status polling, file listing, hybrid retrieval with semantic ranking, chunk normalization, file deletion, and collection cleanup. `DefaultAzureCredential` also selected a system-assigned managed identity in Azure Container Apps, and index creation/deletion, document upload/deletion, and querying all succeeded with the documented `Search Service Contributor` and `Search Index Data Contributor` roles.

Start with `sources/knowledge_layer/src/azure_ai_search/adapter.py` for index ownership, schema validation, ingestion, replacement, and retrieval behavior. Then review `sources/knowledge_layer/src/register.py` for NAT configuration and authentication defaults, followed by `tests/knowledge_layer_tests/test_azure_ai_search.py` for the expected lifecycle and failure handling.

The main design decision is that AI-Q creates and manages only namespaced indexes carrying its ownership marker; unmanaged or incompatible indexes are ignored rather than modified.

#### Related Issues

- None.

## Summary by CodeRabbit

* **New Features**
  * Added **Azure AI Search** as a knowledge-retrieval backend, including installation extras and YAML configuration support.
  * Included Azure authentication modes (API key or managed identity), embedding dimension settings, index-prefix isolation, and collection TTL/cleanup controls.
* **Documentation**
  * Updated configuration reference and Knowledge Layer setup with Azure AI Search backend options and environment variables.
  * Added an end-to-end Azure AI Search example, backend README, and new Web-mode provided configuration profile.
  * Updated Examples navigation to include the Azure AI Search knowledge layer.
* **Tests**
  * Added Azure AI Search adapter tests and expanded adapter compliance coverage.


KyleZheng1284:
@coderabbitai full review

sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -0,0 +1,1174 @@
...
+        metadata: dict[str, Any] | None = None,
+    ) -> tuple[SearchIndex, dict[str, Any]]:
+        index_name = self._physical_index_name(collection_name)
+        try:
+            return self._get_owned_index(collection_name)
+        except ResourceNotFoundError:
+            pass
+
+        marker = _new_marker(collection_name, self.cfg, description, metadata)
+        schema = _build_index_schema(index_name, self.cfg.embed_dim, _encode_marker(marker))
```
KyleZheng1284: Could we take another look at creating one physical Azure index per logical collection? Since the UI maps each document-bearing session to a collection, it seems possible to reach Azure's index limits fairly quickly (Basic allows 5 or 15 indexes, and S1 allows 50). Would a shared index with a required collection/tenant filter, or a bounded sharding approach, fit the session model better? It may also be worth requiring a unique prefix per deployment when a search service is shared.

sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -0,0 +1,1174 @@
...
+        "embed_model": os.environ.get("AIQ_EMBED_MODEL", "nvidia/nv-embed-v1"),
+        "embed_dim": int(os.environ.get("AIQ_EMBED_DIM", "4096")),
+        "embed_api_key": None,
+        "use_hybrid": True,
+        "use_semantic_ranker": False,
+        "chunk_size": 512,
+        "chunk_overlap": 64,
+        "summary_max_chars": 1000,
+        "collection_name": "default",
+        "cleanup_files": True,
```
KyleZheng1284: Would it be safer for `cleanup_files` to default to `False` here both llamaindex and opensearch default to  `False` ? It looks like the public `upload_file()` path inherits this value and may unlink the caller's original file after ingestion. Since the HTTP upload route already passes `cleanup_files=True` for its temporary files, could cleanup stay opt-in for other callers? A small regression test that checks the source file still exists after `upload_file()` might help.

sources/knowledge_layer/pyproject.toml
```diff
@@ -52,8 +57,17 @@ foundational_rag = [
     "docx2txt>=0.8",
     "python-pptx>=0.6.21",
 ]
+azure_ai_search = [
+    "azure-identity>=1.15.0,<1.26",
+    "azure-search-documents>=11.5.0,<11.7",
```
KyleZheng1284: Could we raise this minimum to `azure-search-documents>=11.6.0`? I tried the focused suite with 11.5.3, which is currently allowed, and `SearchIndex.description` was ignored/missing, resulting in 12 test failures. The same 34 tests pass with 11.6.0. Since the ownership marker depends on `description`, 11.6.0 seems like the practical minimum.

sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -0,0 +1,1174 @@
...
+                updated_at=_utc_now().isoformat(),
+            )
+        return self._collection_info(name, index, marker)
+
+    def delete_collection(self, name: str) -> bool:
+        try:
+            index, _marker = self._get_owned_index(name)
+        except ResourceNotFoundError:
+            return False
+        self._index_client.delete_index(index.name)
```
KyleZheng1284: Could this delete use the fetched `SearchIndex` object with `MatchConditions.IfNotModified` instead of deleting by name? Since ownership is checked before the delete, there may be a small window where another process replaces the index and this call deletes the replacement. Using the ETag from the object and treating a 412 as a conflict/retry might preserve the ownership guarantee.

KyleZheng1284 (commented):
A few questions and suggestions on my end

sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -0,0 +1,1174 @@
...
+        collection_name: str,
+        file_id: str,
+        file_name: str,
+        file_size: int,
+        uploaded_at: datetime,
+        metadata: dict[str, Any],
+    ) -> int:
+        from llama_index.core import SimpleDirectoryReader
+
+        old_file_ids = self._find_file_ids_by_name(file_name, collection_name) - {file_id}
```
KyleZheng1284: How do we want overlapping uploads of the same filename to behave? It looks like two workers could both snapshot the same old generation, upload separate new generations, and each delete only the old one, leaving both new generations searchable. Would a durable generation/manifest record with a conditional cutover help here? That could also keep a partially uploaded generation out of retrieval results.

sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -0,0 +1,1174 @@
...
+            collection_name,
+            {"original_filenames": [path.name], "metadata": metadata or {}},
+        )
+        with self._jobs_lock:
+            file_id = self._jobs[job_id].file_details[0].file_id
+            info = self._files[file_id].model_copy(deep=True)
+            info.metadata["job_id"] = job_id
+            return info
+
+    def delete_file(self, file_id: str, collection_name: str) -> bool:
```
KyleZheng1284, line 1251: Could we guard deletion while a file is still `UPLOADING` or `INGESTING`? It looks possible for this method to remove the tracking entry while the background worker continues and later indexes the file, or for the worker to lose its `_files[file_id]` entry. Would it make sense either to reject deletion for nonterminal files or add a cancellation/tombstone state that the worker checks?

sources/knowledge_layer/src/register.py
```diff
@@ -72,6 +84,55 @@ class KnowledgeRetrievalConfig(FunctionBaseConfig, name="knowledge_retrieval"):
...
+    # Azure AI Search options
+    azure_search_endpoint: HttpUrl | None = Field(
+        default_factory=lambda: _url_from_env("AZURE_SEARCH_ENDPOINT"),
+        description="Azure AI Search service URL; defaults to AZURE_SEARCH_ENDPOINT",
+    )
+    azure_search_api_key: SecretStr | None = Field(
+        default_factory=lambda: _secret_from_env("AZURE_SEARCH_API_KEY"),
+        description="Optional Azure AI Search admin key; defaults to AZURE_SEARCH_API_KEY",
+    )
+    azure_search_auth_mode: Literal["managed_identity", "api_key"] = Field(
```
KyleZheng1284: Could the default auth mode be resolved after the model fields are parsed? When I provide `azure_search_api_key` directly in YAML but do not set the environment variable or `azure_search_auth_mode`, the resulting mode is still `managed_identity`, so the configured key is not used. Maybe `None`/`auto` internally, followed by a validator based on the parsed key, would make the behavior less surprising.

sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -0,0 +1,1174 @@
...
+
+def _validate_index_name(name: str) -> None:
+    if not re.fullmatch(r"[a-z0-9](?:[a-z0-9-]{0,126}[a-z0-9])?", name) or "--" in name:
+        raise ValueError(
+            f"Invalid Azure AI Search index name {name!r}. Names must use lowercase letters, digits, and single "
+            "hyphens; be 2-128 characters; and start and end with a letter or digit."
+        )
+
+
+def _encode_marker(marker: dict[str, Any]) -> str:
```
KyleZheng1284, line 161: Would it be worth keeping this ownership marker intentionally small, or validating its serialized size before calling Azure? The marker currently includes the full user description and metadata, while Azure index descriptions must be under 4,000 characters. Perhaps the ownership/schema fields could stay here and arbitrary collection metadata could live in a manifest document.

sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -0,0 +1,1174 @@
...
+
+
+def _build_search_credential(cfg: SimpleNamespace):
+    """Pick the Azure SDK credential without exposing secret values."""
+    if cfg.auth_mode == "api_key":
+        api_key = _secret_value(cfg.api_key)
+        if api_key is None:
+            raise ValueError("auth_mode=api_key requires the `api_key` field to be set")
+        return AzureKeyCredential(api_key)
+    return DefaultAzureCredential()
```
KyleZheng1284: Would `default_credential` be a clearer name for this mode, or should this instantiate `ManagedIdentityCredential` directly? A parameterless `DefaultAzureCredential` can also select environment, workload-identity, Azure CLI, and other credentials, so `managed_identity` may set a narrower expectation than the implementation provides.

docs/source/examples/azure-ai-search.md
```diff
@@ -0,0 +1,82 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Example: Azure AI Search knowledge layer
+
+Use Azure AI Search as the document store while retaining AI-Q's existing
+upload API, per-conversation collection routing, document summaries, and
+citations. This example assumes the Azure AI Search service and embedding
```
KyleZheng1284, line 10: Could we add a short Azure setup prerequisite here, or link directly to the Microsoft quickstart? The page currently assumes the service already exists, so a new user may not know which Azure AI Search resource/tier to create, where to find the endpoint or admin key, or how to enable RBAC for DefaultAzureCredential. Since AI-Q creates one index per logical collection, it may also help to mention the index limits when choosing a tier. The detailed provisioning steps could stay in the Azure docs; a brief checklist and link here would probably be enough.

docs/source/customization/configuration-reference.md
```diff
@@ -260,6 +276,19 @@ functions:
...
+| `azure_search_index_prefix` | `str` | `AIQ_AZURE_SEARCH_INDEX_PREFIX` or `aiq` | Namespace prefix for AI-Q-owned indexes. |
+| `embed_base_url` | `URL` | `AIQ_EMBED_BASE_URL` or NVIDIA API | OpenAI-compatible embedding base URL. |
+| `embed_model` | `str` | `AIQ_EMBED_MODEL` or `nvidia/nv-embed-v1` | Embedding model used for Azure ingestion and retrieval. |
+| `embed_dim` | `int` | `AIQ_EMBED_DIM` or `4096` | Embedding dimensions; must match the model and existing index schema. |
+| `embed_api_key` | `SecretStr` | `None` | Optional embedding API key; falls back to `NVIDIA_API_KEY`. |
+| `use_hybrid` | `bool` | `true` | Combine lexical and vector retrieval. |
+| `use_semantic_ranker` | `bool` | `false` | Apply Azure semantic ranking when supported; requires `use_hybrid: true`. |
+| `chunk_size` | `int` | `512` | Tokens per Azure-ingested chunk. |
+| `chunk_overlap` | `int` | `64` | Token overlap; must be smaller than `chunk_size`. |
+| `summary_max_chars` | `int` | `1000` | Maximum document characters sent to the summary model. |
```
KyleZheng1284: These options are useful, but exposing all of them increases the configuration and support surface. Have we considered which ones users really need to change? If we keep them all, can we make sure invalid values and unsupported combinations fail early with clear error messages?

sources/knowledge_layer/pyproject.toml
harmke: Done, raised the minimum to `11.6.0`.

sources/knowledge_layer/src/azure_ai_search/adapter.py
harmke: Done,  `cleanup_files` now defaults to `False`.

sources/knowledge_layer/src/register.py
harmke: Addressed. I removed `azure_search_auth_mode` entirely. Authentication is now inferred from the parsed configuration: when `azure_search_api_key` is provided through YAML or `AZURE_SEARCH_API_KEY`, API-key authentication is used; otherwise, it falls back to managed identity.

docs/source/examples/azure-ai-search.md
harmke: Addressed. I added a short prerequisites checklist covering service creation and tier selection, where to find the endpoint and admin key, RBAC setup for `DefaultAzureCredential`, and the one-index-per-collection impact on index limits. I also linked to the relevant Microsoft setup and service-limit documentation for the detailed steps.

sources/knowledge_layer/src/azure_ai_search/adapter.py
harmke: Addressed in `94c8bf8z`. Azure now uses one physical index per deployment prefix and embedding configuration. Logical collections use manifest records and mandatory `collection_id`/`record_type` filters. The prefix is documented as deployment-unique.

sources/knowledge_layer/src/azure_ai_search/adapter.py
harmke: Addressed in `94c8bf8`. `file_id` is now authoritative, so concurrent same-name uploads coexist independently. A file manifest is created before ingestion, chunks are scoped by file ID, and failed uploads roll back their chunks.

sources/knowledge_layer/src/azure_ai_search/adapter.py
harmke: Addressed in `94c8bf8`. File and collection deletion now reject files in `UPLOADING` or `INGESTING` state, with regression coverage.

sources/knowledge_layer/src/azure_ai_search/adapter.py
harmke: Addressed in `94c8bf8`. The marker now contains only ownership/schema/prefix/embedding fields and creation time, with an explicit 4,000-character limit. Collection descriptions and metadata live in manifest documents.

docs/source/customization/configuration-reference.md
harmke: Addressed in `94c8bf8`. Removed Azure-specific chunking, hybrid, semantic-ranker, and summary-length options. Chunking and hybrid retrieval now use fixed internal behavior.

sources/knowledge_layer/src/azure_ai_search/adapter.py
harmke: Addressed by removing the explicit auth mode.

docs/source/customization/configuration-reference.md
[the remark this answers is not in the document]
harmke: Addressed by removing the explicit auth mode.

sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -0,0 +1,1298 @@
...
+from aiq_agent.knowledge import unregister_summary
+from aiq_agent.knowledge.base import CollectionInfo
+from aiq_agent.knowledge.base import FileInfo
+from aiq_agent.knowledge.base import TTLCleanupMixin
+from aiq_agent.knowledge.schema import FileStatus
+
+logger = logging.getLogger(__name__)
+
+_BACKEND_NAME = "azure_ai_search"
+_SCHEMA_VERSION = 1
```
KyleZheng1284, line 62: Could we align the schema version here with the tests and documentation? The focused Azure suite currently fails because this is `1`, while the schema validation test expects `2`, and the documentation says schema-v1 indexes are ignored. If the shared-index layout is intended to be version 2, could we bump this constant? Otherwise, the test and migration documentation may need to remain on version 1.

sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -0,0 +1,1298 @@
...
+                file_name=file_name,
+                collection_name=collection_name,
+                status=FileStatus.UPLOADING,
+                file_size=Path(path).stat().st_size,
+                uploaded_at=submitted_at,
+                metadata={**file_metadata, "job_id": job_id},
+            )
+            with self._jobs_lock:
+                self._files[file_id] = info
+            self._write_file_manifest(info)
```
KyleZheng1284: Could the job be registered before the remote manifest writes, or could manifests already written be rolled back if a later write fails? The other adapters create the job first and perform backend work in the background, which matches the `BaseIngestor.submit_job()` contract. Here, a failure on file N can leave files 1 through N-1 stored as `UPLOADING` without a registered job or worker to finish them. A failure-injection test on the second manifest would help cover this case.

sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -0,0 +1,1298 @@
...
+            return False
+        if info.status in {FileStatus.UPLOADING, FileStatus.INGESTING}:
+            raise ValueError(f"Cannot delete file {file_id!r} while it is ingesting")
+
+        self._delete_file_documents(file_id, collection_name)
+        self._delete_document_ids(
+            self._get_validated_search_client(),
+            [_record_id(_RECORD_FILE, collection_name, file_id)],
+        )
+        self._deleted_files.add((collection_name, file_id))
```
KyleZheng1284, line 1263: We should make sure the post-delete bookkeeping runs once the Azure delete calls succeed, even if the search-visibility wait times out? The tombstone is currently added before the wait, so a timeout skips `_files` cleanup, summary restoration, and the collection timestamp update. A retry then returns `False` because the tombstone hides the file. Perhaps the bookkeeping could run in a committed/finally path, with the consistency timeout handled separately.

sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -0,0 +1,1298 @@
...
+        return await asyncio.to_thread(self._retrieve_sync, query, collection_name, top_k, filters)
+
+    def _retrieve_sync(
+        self,
+        query: str,
+        collection_name: str,
+        top_k: int,
+        filters: dict[str, Any] | None,
+    ) -> RetrievalResult:
+        if filters and isinstance(filters, dict) and filters.get("$filter"):
```
KyleZheng1284: It seems like Azure either supports the shared metadata-filter contract or explicitly reject every unsupported non-empty `filters` value? OpenSearch and Foundational RAG apply these filters, while the base retriever interface documents them as supported. Azure currently rejects raw `$filter`, but something like `{"file_name": "report.pdf"}` is silently ignored and runs a broader query than the caller requested.

tests/knowledge_layer_tests/run_adapter_compliance.py
```diff
@@ -19,6 +19,8 @@
     python tests/knowledge_layer_tests/run_adapter_compliance.py --backend llamaindex --quick
     python tests/knowledge_layer_tests/run_adapter_compliance.py --backend foundational_rag --quick
     python tests/knowledge_layer_tests/run_adapter_compliance.py --backend opensearch --quick
+    python tests/knowledge_layer_tests/run_adapter_compliance.py --backend azure_ai_search --quick \
```
KyleZheng1284, line 22: Could we add a full Azure AI Search example to the adapter compliance instructions? The runner currently shows full-mode commands for LlamaIndex, Foundational RAG, and OpenSearch, but only a quick registration check for Azure. A copy-paste command that keeps credentials in environment variables, uses a reviewer-specific index prefix, and disables the TTL thread would make the expected live validation path clear.

docs/source/examples/azure-ai-search.md
```diff
@@ -0,0 +1,85 @@
...
+| Role | Used for |
+|------|----------|
+| `Search Service Contributor` | Create and inspect the shared AI-Q index. |
+| `Search Index Data Contributor` | Upload, query, and delete index documents. |
+
+Assign the roles at the search-service scope. The principal ID is the object ID
+of the system-assigned or user-assigned managed identity running AI-Q.
+
+
+Replace the `knowledge_search` block in a web configuration such as
```
KyleZheng1284: Could we add a checked-in `configs/config_web_azure_ai_search.yml` and update this example to reference it? That would make the UI E2E path copy-pasteable with `./scripts/start_e2e.sh --config_file configs/config_web_azure_ai_search.yml`, while keeping the endpoint and credentials in environment variables.

docs/source/examples/azure-ai-search.md
```diff
@@ -0,0 +1,85 @@
...
+Install the backend dependency:
+
+```bash
+uv pip install -e "sources/knowledge_layer[azure_ai_search]"
+```
+
+Set the environment used by the adapter:
+
+```bash
+export AZURE_SEARCH_ENDPOINT=https://<service>.search.windows.net
```
KyleZheng1284: Could we also add placeholder entries for `AZURE_SEARCH_ENDPOINT`, `AZURE_SEARCH_API_KEY`, and `AIQ_AZURE_SEARCH_INDEX_PREFIX` to `deploy/.env.example`, and mention here that `start_e2e.sh` reloads `deploy/.env`? That would make the Azure E2E config work without extra exports and avoid a valid shell value being overwritten by a stale `.env` value. The example should use placeholders/defaults rather than a real service endpoint, key, or reviewer-specific prefix.

KyleZheng1284:
@harmke everything worked great E2E, could you please address the remaining comments would love to get this into the 2.2 release!

sources/knowledge_layer/src/azure_ai_search/adapter.py
harmke: Addressed in `b86340a`. Since the Azure AI Search backend has not shipped yet, the shared-index layout remains schema version 1. I aligned the schema validation test with version 1, changed the mismatch test to use `_SCHEMA_VERSION + 1`, and removed the documentation that incorrectly described version-1 indexes as legacy. The full Azure unit suite passes.

sources/knowledge_layer/src/azure_ai_search/adapter.py
harmke: Addressed in `4809562`. Once Azure accepts the document deletion, local file cleanup, summary restoration, and the collection timestamp update now complete before the search-visibility wait. If that wait times out, the timeout still propagates, but bookkeeping remains committed and a retry safely reports that the file is already deleted. Added deterministic timeout/retry regression coverage.

sources/knowledge_layer/src/azure_ai_search/adapter.py
harmke: Addressed in `1813715`. Azure AI Search now explicitly rejects every non-empty caller-provided filter dictionary instead of silently ignoring unsupported values and broadening the query. Both raw $filter expressions and structured filters such as `{"file_name": "report.pdf"}` return a clear unsupported-filter error before any Azure request is made. `None` and `{}` continue through the normal collection-scoped retrieval path. Added regression coverage for both rejection cases.

tests/knowledge_layer_tests/run_adapter_compliance.py
harmke: Addressed in `aebac16`. The compliance instructions now include a full Azure AI Search run using environment-based credentials, a reviewer-specific `aiq-${USER}` index prefix, and `start_ttl_cleanup=false`. I validated the path against live Azure: all 14 compliance checks passed, including ingestion, retrieval, file deletion, and collection cleanup.

docs/source/examples/azure-ai-search.md
harmke: Addressed in `aebac16`. Added `configs/config_web_azure_ai_search.yml`.

docs/source/examples/azure-ai-search.md
harmke: Addressed in `aebac16`. Added safe placeholder/default entries for `AZURE_SEARCH_ENDPOINT`, `AZURE_SEARCH_API_KEY`, and `AIQ_AZURE_SEARCH_INDEX_PREFIX` to `deploy/.env.example`, plus optional `AZURE_CLIENT_ID` for user-assigned managed identity. The resulting E2E configuration was validated successfully against live Azure AI Search.

docs/source/customization/knowledge-layer.md
```diff
@@ -152,6 +156,37 @@ functions:
...
+LlamaIndex backend through `AIQ_EMBED_BASE_URL` and `AIQ_EMBED_MODEL`; set
+`AIQ_EMBED_DIM` when changing the model dimensions. Set a deployment-unique
+`AIQ_AZURE_SEARCH_INDEX_PREFIX` when multiple AI-Q deployments share a search
+service.
+
+Azure stores all logical collections in one physical index selected by the
+prefix, schema version, embedding model, and dimension. Collection, file, and
+chunk manifests enforce logical isolation. Retrieval is always hybrid, and
+chunking is fixed at 1024 tokens with 128-token overlap. Schema-version-1
+indexes are ignored and require re-ingestion.
```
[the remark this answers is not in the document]
harmke: Addressed in `b86340a`. Since the Azure AI Search backend has not shipped yet, the shared-index layout remains schema version 1.

tests/knowledge_layer_tests/run_adapter_compliance.py
```diff
@@ -19,12 +19,23 @@
...
     # Full mode - complete ingestion + retrieval test
     python tests/knowledge_layer_tests/run_adapter_compliance.py --backend llamaindex
     python tests/knowledge_layer_tests/run_adapter_compliance.py --backend foundational_rag
     python tests/knowledge_layer_tests/run_adapter_compliance.py --backend opensearch
 
+    export AZURE_SEARCH_ENDPOINT=https://your-service.search.windows.net
+    export AZURE_SEARCH_API_KEY=your-search-admin-key
+    export AIQ_AZURE_SEARCH_INDEX_PREFIX="aiq-${USER}"
+    python tests/knowledge_layer_tests/run_adapter_compliance.py --backend azure_ai_search \
+      --config '{"start_ttl_cleanup":false}'
```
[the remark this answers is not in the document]
harmke: Not in scope for this PR, should align with the other tests.

configs/config_web_azure_ai_search.yml
```diff
@@ -0,0 +1,211 @@
+# This is the Web mode configuration for Azure AI Search.
+# It has the following features:
+# - Knowledge retrieval using Azure AI Search
+# - Web search and Paper search tools by default
```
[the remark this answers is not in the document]
KyleZheng1284: @harmke could we resolve this

KyleZheng1284 (approved):
LGTM

KyleZheng1284:
@coderabbitai review

KyleZheng1284:
@coderabbitai full review

PR #325, opened 2026-07-09
feat(helm): support custom backend config via ConfigMap

### LINKED ISSUE #243: Customize values.yaml file in ai-q blueprint helm chart to use user provided config file
I have created a config.yaml file that connects to my MCP server, and I have already validated it successfully using the NVIDIA AgentIQ Toolkit (nat) command line.

Now, I want to use this configuration within the AI-Q Blueprint Helm chart. I noticed there is an option to use existing config files from the config/ directory, but I do not see a way to specify an external or custom config file during Helm installation.

While I could manually copy the file into the aiq-backend container and use it from there, I would prefer a cleaner approach where the config file can be specified directly through Helm values at installation time.

Ideally, there should be a value that points to a custom config file, or an option to mount the file during deployment. Currently, I only see the following predefined configuration files available in the config/ directory:

CONFIG_FILE: configs/config_web_frag.yml

I would appreciate any guidance or recommended approach for handling custom configuration files in this setup.

### LINKED ISSUE #302: [Docs]: Local inference on K8s not documented
### Documentation Location

https://docs.nvidia.com/aiq-blueprint/latest/customization/swapping-models.html#using-downloadable-nims-self-hosted

### Issue

To do local inference, the configuration cannot be customized easily (the config file needs to be created manually and store it as a configmap)

https://docs.nvidia.com/aiq-blueprint/latest/deployment/kubernetes.html#configuration

### Suggested Fix

Add a section on the docs on how to create a custom configmap with custom tweaks on the docs.

This probably will require changing the helm chart as well to expose a configmap reference instead a file reference at aiq.apps.backend.env.CONFIG_FILE

### Pre-Submission Checks

- [x] I searched existing issues and pull requests.
- [x] This report does not include secrets, credentials, or private data.

e-minguez:
#### Overview

Adds a first-class option to supply a custom AI-Q backend workflow config to the Helm chart, addressing #302. Previously the only way to configure the Kubernetes deployment was selecting one of the config YAMLs baked into the image via `CONFIG_FILE`; using a custom config meant hand-building a ConfigMap and wiring volumes manually.

A new `apps.<app>.config` block (enabled for `backend`) supports two paths:
  - Inline — paste config YAML into `content`; the chart renders a ConfigMap.
  - Existing ConfigMap — reference one you created out-of-band via `existingConfigMap`.

#### Validation

  - helm lint deploy/helm/deployment-k8s/ — passed
  - helm template for default / inline / existing-ConfigMap cases — verified ConfigMap rendering, volume + mount at /app/custom-configs, and single CONFIG_FILE pointing at the mount (baked default dropped when
  enabled); default render unchanged
  - uv run pytest tests/deploy/test_helm_deployment_k8s.py — 9 passed (4 new + 5 existing)
  - uv run ruff check tests/deploy/test_helm_deployment_k8s.py — clean
  - sphinx-build -b html docs/source — build succeeded, no warnings/broken xrefs on the edited pages
  - [x] I certify this contribution under the Developer Certificate of Origin (DCO) and signed my commits with git commit -s or an equivalent sign-off.

`deploy/helm/helm-charts-k8s/aiq/templates/deployment.yaml` — the `config.enabled` logic that drops the baked `CONFIG_FILE` and injects the mount path (the key design decision), plus the new `templates/app-config-configmap.yaml`. Then `tests/deploy/test_helm_deployment_k8s.py` for the covered cases. Note the vendored subchart is repackaged as `charts/aiq-0.0.6.tgz`

#### Related Issues

- Closes #302 
- Closes #243 

## Summary by CodeRabbit

- **New Features**
  - Added optional custom backend configuration for Kubernetes deployments through inline YAML or an existing ConfigMap.
  - Supports configurable configuration filenames and mount paths.
  - Custom settings take precedence over the default backend configuration.

- **Bug Fixes**
  - Added validation for missing or conflicting configuration sources and invalid mount paths.
  - Inline configuration changes trigger automatic pod rollouts; existing ConfigMap updates require a manual restart.

- **Documentation**
  - Expanded Kubernetes, Helm, Docker Compose, and model-swapping guidance with configuration examples and deployment instructions.


deploy/helm/helm-charts-k8s/aiq/templates/app-config-configmap.yaml
```diff
@@ -0,0 +1,17 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "aiq.appFullname" (list $ $appName) }}-config
+  namespace: {{ include "aiq.namespace" $ }}
+  labels:
+    {{- include "aiq.appLabels" (list $ $appName $appConfig) | nindent 4 }}
+data:
+  {{ $appConfig.config.fileName | default "config.yml" }}: |
+    {{- $appConfig.config.content | nindent 4 }}
```
cdgamarose-nv: The backend only reads CONFIG_FILE during startup. If someone changes config.content with helm upgrade, Kubernetes updates the ConfigMap but leaves the existing pods running with the old workflow. Could we add a checksum annotation to trigger a rollout, and document that externally managed ConfigMaps require a restart when their contents change?

deploy/helm/helm-charts-k8s/aiq/templates/app-config-configmap.yaml
```diff
@@ -0,0 +1,17 @@
+{{- range $appName, $appConfig := .Values.apps }}
+{{- if and $appConfig.enabled $appConfig.config }}
+{{- if and $appConfig.config.enabled $appConfig.config.content (not $appConfig.config.existingConfigMap) }}
```
cdgamarose-nv: If config.enabled is true but neither content nor existingConfigMap is set, the chart still mounts aiq-backend-config even though that ConfigMap was never created. Helm reports success, but the backend pod then fails to start. It would be better to fail during rendering with a clear message .

docs/source/deployment/kubernetes.md
```diff
@@ -195,6 +197,84 @@ helm upgrade --install aiq aiq2-web-2.0.0.tgz -n ns-aiq \
...
+**Option A -- Inline (chart renders the ConfigMap):**
+
+```yaml
+# aiq-custom-config-values.yaml
+aiq:
+  apps:
+    backend:
+      config:
+        enabled: true
+        content: |
```
cdgamarose-nv: This example is mounted as the entire CONFIG_FILE; it is not merged with one of the bundled configurations. As written, it only defines an LLM and is missing the functions and workflow needed to run the backend. Could we instead tell users to copy and edit a complete shipped config, or provide a complete example that has been tested end to end?

deploy/helm/helm-charts-k8s/aiq/templates/deployment.yaml
```diff
@@ -227,13 +235,18 @@ spec:
...
+        {{- if or $appConfig.volumeMounts (and $.Values.fusionConfig $.Values.fusionConfig.enabled) (and $appConfig.config $appConfig.config.enabled) }}
         volumeMounts:
         {{- if and $.Values.fusionConfig $.Values.fusionConfig.enabled }}
         - name: {{ $.Values.fusionConfig.volumeName | default "config" }}
           mountPath: {{ $.Values.fusionConfig.mountPath | default "/app/config" }}
           readOnly: true
         {{- end }}
+        {{- if and $appConfig.config $appConfig.config.enabled }}
+        - name: app-config
+          mountPath: {{ $appConfig.config.mountPath | default "/app/custom-configs" }}
```
cdgamarose-nv: The documentation says /app/configs must not be used as the custom mount path, but the chart currently accepts it. Could we either enforce that restriction with a clear Helm error or remove the warning if mounting there is actually supported?

cdgamarose-nv (changes requested):
Thanks for putting this together. I checked the linked issues against the current develop branch, and the request is still valid. It is possible to mount a custom config today using CONFIG_FILE, volumes, and volumeMounts, but doing so is fairly manual and is not clearly documented. The inline/existing-ConfigMap support in this PR would make that workflow much easier and would help address both #302 and #243.
The branch has fallen quite a bit behind develop though, and now has conflicts across the chart, deployment template, docs, and tests. Could you rebase it onto the latest develop? While resolving those conflicts, please keep the current 2.2.0 versions and the newer deployment changes, and replace the old archive.

deploy/helm/helm-charts-k8s/aiq/templates/app-config-configmap.yaml
e-minguez: Good call. Thanks.

deploy/helm/helm-charts-k8s/aiq/templates/app-config-configmap.yaml
e-minguez: Good call. Thanks.

deploy/helm/helm-charts-k8s/aiq/templates/deployment.yaml
e-minguez: Good call. Thanks.

docs/source/deployment/kubernetes.md
e-minguez: Good call. Thanks.

e-minguez:
> Thanks for putting this together. I checked the linked issues against the current develop branch, and the request is still valid. It is possible to mount a custom config today using CONFIG_FILE, volumes, and volumeMounts, but doing so is fairly manual and is not clearly documented. The inline/existing-ConfigMap support in this PR would make that workflow much easier and would help address both #302 and #243. The branch has fallen quite a bit behind develop though, and now has conflicts across the chart, deployment template, docs, and tests. Could you rebase it onto the latest develop? While resolving those conflicts, please keep the current 2.2.0 versions and the newer deployment changes, and replace the old archive.

Thanks for the review. I sent the updated PR with the rebase and the other fixes you suggested. PTAL.

PR #327, opened 2026-07-09
Add agent observability with Monocle

imohammedansari:
## Summary

Adds Monocle observability to the agent. Monocle is an OpenTelemetry-based tracer for LLM applications. With it enabled, each run is recorded as a structured trace: the agent and graph invocations, tool calls, LLM inferences, token usage, and timings. The change is additive and does not alter application logic.

## What this adds

Instrumentation is one setup call plus one dependency:

- `frontends/cli/pyproject.toml`: adds `monocle_apptrace`.
- `frontends/cli/cli.py`: calls `setup_monocle_telemetry(workflow_name="nvidia-aiq", monocle_exporters_list="file")` at startup.

That call auto-instruments the frameworks already in use (LangGraph and the LLM clients), so there is no per-tool or per-call wiring to maintain. Traces are written to `.monocle/` by default.

## What you get

Each run produces a trace of all the agents and tools that were triggered: which agent ran, which tools were called (`web_search_tool`), and what LLM inferences happened. In effect, it's the path the agent took to answer a given question. This is useful for developers building the agent, since they can see how it actually behaved on a run. The same traces are also a good basis for a behavioral test suite, an integration test that asserts on that behavior, and I've opened a companion PR that shows how that works: [behavioral test suite](https://github.com/NVIDIA-AI-Blueprints/aiq/pull/328). You can open the trace files directly, view them in the [Monocle VS Code extension](https://marketplace.visualstudio.com/items?itemName=OkahuAI.monocle-apptrace), or send them to [Okahu](https://www.okahu.ai) for analysis across many runs.

---

## Example trace (Okahu VS Code Extension)

The AI-Q (NAT) research agent driving the model and calling `web_search_tool`, captured as trace spans.

image

## Summary by CodeRabbit

* **New Features**
  * Added optional Monocle-based tracing for AI-Q workflows with environment- or workflow-configured exporter selection (including local file/console and optional Okahu export).
  * Introduced a Monocle workflow telemetry exporter for end-to-end agent execution visibility.
* **Bug Fixes**
  * Prevented Monocle tracing dependencies from loading unless enabled.
  * Flushes pending tracing data before CLI shutdown to avoid dropped final spans.
* **Documentation**
  * Updated observability and deployment docs with enablement steps, captured-data/retention guidance, and configuration options.
* **Chores**
  * Added `monocle` optional dependency extra (`aiq-agent[monocle]`).

---

PS: if Monocle looks useful, a ⭐ helps the project (https://github.com/monocle2ai/monocle). And if you want to turn these traces into tests, that's the companion PR (#328).

### DIFF
--- a/deploy/.env.example
+++ b/deploy/.env.example
@@ -118,6 +118,12 @@ DASK_DISTRIBUTED__LOGGING__DISTRIBUTED=warning
 # LANGCHAIN_PROJECT=aiq-research
 # WANDB_API_KEY=
 
+# Monocle (optional): OpenTelemetry tracing for agents. Requires the `monocle`
+# extra (`uv sync --extra monocle`). See docs/source/deployment/observability.md.
+# MONOCLE_TRACING=true
+# MONOCLE_EXPORTERS=file          # file, console, okahu, s3, blob, gcs (default: file)
+# OKAHU_API_KEY=okh_xxxxxxxx      # required only for the `okahu` exporter
+
 # -----------------------------------------------------------------------------
 # Evaluation (optional)
 # -----------------------------------------------------------------------------
--- a/docs/source/deployment/observability.md
+++ b/docs/source/deployment/observability.md
@@ -255,6 +255,67 @@ general:
         shutdown_timeout: 30000
 ```
 
+## Monocle
+
+[Monocle](https://github.com/monocle2ai/monocle) is an open-source, OpenTelemetry-based tracer for agentic applications. It instruments the frameworks already in use (LangChain, LangGraph, and others) and records each run end-to-end -- LLM calls, agent steps, and tool and MCP invocations, with their inputs, outputs, timings, and token counts. It is packaged as an **optional** tracing backend and is disabled unless you both install it and enable it -- the default install and default behavior are unchanged.
+
+Each run writes one trace file to `.monocle/` in the working directory; open it in the [Monocle VS Code extension](https://marketplace.visualstudio.com/items?itemName=OkahuAI.monocle-apptrace) to inspect the span timeline and token counts. Connect to [Okahu](https://www.okahu.ai), an agent-observability platform, to analyze traces across runs and run trace-based and agentic evaluations (via the `okahu` exporter).
+
+### Install
+
+Monocle ships as an optional extra, so a default install does not pull the OpenTelemetry stack:
+
+```bash
+uv sync --extra monocle
+# or, in an existing environment:
+pip install 'aiq-agent[monocle]'
+```
+
+Enabling Monocle without the extra installed fails fast with an actionable error naming the install command; when Monocle is not enabled, `monocle_apptrace` is never imported.
+
+### Enable
+
+There are two ways to opt in. Both forward the exporter list to `setup_monocle_telemetry(workflow_name="nvidia-aiq", monocle_exporters_list=...)` and Monocle initializes at most once per process.
+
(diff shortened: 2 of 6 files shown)


imohammedansari:
Thanks for the review. Addressed the CodeRabbit findings on the latest push (`36f53c9`, DCO signed-off):

**Fixed**
- **`monocle_exporter.py` — graceful degradation on missing secret:** `okahu` selected without `OKAHU_API_KEY` no longer raises. It's dropped with a warning and the remaining exporters continue; if none remain, Monocle is skipped cleanly. An *unknown* exporter (a config typo, not a missing secret) still fails fast with an actionable message.
- **`monocle_exporter.py` — YAML no-op short-circuit:** when the env gate already initialized Monocle, the `monocle` telemetry exporter now returns the no-op exporter before re-validating the (documented-as-ignored) YAML settings.
- **`frontends/cli/cli.py` — flush before hard exit:** `os._exit(0)` now runs after `flush_monocle_if_enabled()`, force-flushing pending Monocle/OTel spans so the last spans of a run aren't dropped.

**Already resolved in the current revision**
- **Module-level telemetry init / `--help` side effect:** initialization now happens inside `main()` after arg parsing (env gate), never at import; `cli.py:46` no longer references Monocle.
- **`frontends/cli/pyproject.toml` unconstrained dep:** `monocle_apptrace` is no longer a dependency there — it's an opt-in `monocle` extra in the root `pyproject.toml`.

**Version pin (kept, with a note):** `monocle_apptrace>=0.8.8,<0.9` — `0.8.8` is published on PyPI (verified; it's the current latest stable). The automated "unavailable" flag appears to be from stale index data. Happy to adjust the floor if you'd prefer.

**Redaction (out of scope, noted):** trace payloads are exported verbatim by design; off-box exporters now emit a data-egress warning and the docs call this out. Payload redaction is configured at the Monocle layer (custom output processors), so it's out of scope for this opt-in-enablement PR rather than something to hard-code here.

imohammedansari:
Thanks — addressed both follow-ups in `d458cd2`:

- **Off-box egress warning on the YAML path** (`monocle_exporter.py`): extracted the egress warning into a shared `_warn_off_box()` helper and now call it in *both* opt-in paths, so a YAML-configured `okahu`/`s3`/`blob`/`gcs` exporter logs the same "trace data leaving local disk" warning as the env-gated path (no longer silent).
- **Docs** (`observability.md`): corrected the Data-handling note — an unknown exporter still fails fast, but a missing `OKAHU_API_KEY` now *warns and skips* the `okahu` exporter (continuing with the rest; Monocle skipped if none remain), with a reminder to check logs to confirm remote export is actually active.

tanleach:
Thanks for putting this together, Mohammed, and for working through the earlier review feedback. Since this PR was opened, AIQ’s `develop` branch has adopted NeMo Relay as its observability runtime in #452. Relay now owns the agent, LLM, and tool scope hierarchy; local ATOF output; OTEL/OpenInference export; redaction; pricing; CLI lifecycle; and asynchronous-job trace correlation.

Because this PR introduces a second process-global instrumentation and exporter pipeline, merging it now would duplicate that architecture and create separate privacy, lifecycle, configuration, and trace-hierarchy contracts. The branch also conflicts with current `develop`, so we do not think asking you to rebase this implementation would be the right direction.

If Okahu can ingest Relay’s OTLP/OpenInference output, or Monocle can operate as a Relay exporter or consumer without re-instrumenting AIQ, we would be interested in discussing that narrower design. Otherwise, we’re going to close this PR as superseded by #452.

Thank you again for the contribution and for starting the observability discussion.

PR #328, opened 2026-07-09
Add trace-based behavioral tests with Monocle Test Tools

imohammedansari:
## Summary

Adds a behavioral test suite that asserts against the AI-Q research agent's Monocle execution traces: which agent ran, which tools it called, what it returned, and its duration. Additive only, under `tests/monocle/`, with no app-code changes.

## Why

The existing suite (100+ files under `tests/`) is thorough at the unit layer but never runs the assembled agent to check what a real run actually did. This adds that behavioral coverage, so if a later prompt, tool, or route change alters routing, tool calls, or latency, a test catches it.

## How it works

It uses Monocle Test Tools. The offline tests load a recorded trace from file with `with_trace_source("file", trace_path=...)`, which is fast, needs no keys, and is deterministic. They then assert with the fluent API: `called_agent`, `called_tool`, `contains_input` / `contains_output`, `under_duration`. The live test drives the real NAT workflow (via `configs/config_openai_cli.yml`, with an auto-approving HITL callback so a clarification or plan-approval interrupt does not block) and asserts on structure and budget only.

The point of loading from file is that the same assertions run in both modes. You pin a check against a recorded trace while iterating (fast, no keys, deterministic), then run the identical check against a live run for real regression coverage.

NAT traces carry no token metadata (inference spans record only finish reasons), so budgets are duration-based. The live test is opt-in (`RUN_LIVE_NVIDIAAIQ=1`) and runs one at a time, since NAT binds its asyncio singletons to the first event loop.

## Changes (all under `tests/monocle/`)

- `test_nvidiaaiq.py`: 2 offline file-loaded tests (a capabilities intro answered directly; an NVIDIA stock-price lookup via `web_search_tool`), plus 1 live test (stock-price lookup, opt-in via `RUN_LIVE_NVIDIAAIQ=1`).
- `conftest.py`: Monocle setup, `.env` loading, `run_nvidiaaiq()` (auto-approve HITL).
- `traces/`: 2 recorded traces.
- `configs/config_openai_cli.yml`, `requirements.txt`, `README.md`.

## Summary by CodeRabbit

- **New Features**
  - Added a ready-to-use OpenAI CLI configuration with web search, research workflows, citation verification, and configurable checkpoints.
  - Configured shallow vs. deep research behavior with tailored escalation, clarification, and planning.

- **Documentation**
  - Added a Monocle-based trace testing guide, including offline replay and opt-in live run instructions.

- **Tests**
  - Added Monocle trace-based behavioral tests for capability queries and stock-price research, validating tool usage and workflow duration budgets (with optional live verification).

- **Chores**
  - Updated ignore rules to prevent Monocle trace artifacts from being tracked.

---

PS: if Monocle looks useful, a ⭐ helps the project (https://github.com/monocle2ai/monocle). The instrumentation that emits these traces is the companion PR (#327).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,9 @@ env/
 .env
 .env.local
 .env.*.local
+
+# Monocle traces
+.monocle/
 # Logs
 *.log
 logs/
--- a/configs/config_openai_cli.yml
+++ b/configs/config_openai_cli.yml
@@ -1,0 +1,129 @@
+# This is the default configuration for the CLI mode.
+# It has the following features:
+# - Human-in-the-loop clarification and plan approval before deep research starts
+# - Web search and Paper search tools by default
+# - There is no knowledge retrieval
+
+general:
+  telemetry:
+    logging:
+      console:
+        _type: console
+        level: INFO
+    # tracing:
+    #   langsmith: # Optional: LangSmith tracing - requires langsmith API key. Set using `export LANGSMITH_API_KEY=<your-langsmith-api-key>`
+    #     _type: langsmith
+    #     project: nvidia-aiq
+
+llms:
+  nemotron_llm_intent:
+    _type: openai
+    model_name: gpt-4o
+    api_key: ${OPENAI_API_KEY}
+    temperature: 0.5
+    top_p: 0.9
+    max_tokens: 4096
+    num_retries: 5
+
+  nemotron_super_llm:
+    _type: openai
+    model_name: gpt-4o
+    api_key: ${OPENAI_API_KEY}
+    temperature: 0.7
+    top_p: 0.7
+    max_tokens: 16384
+    num_retries: 5
+
+functions:
+  # =========================================================================
+  # Data Source Registry
+  # =========================================================================
+  # Central registry that controls:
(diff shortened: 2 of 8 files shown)


tanleach:
Thanks for the companion testing proposal. The goal of catching agent-behavior regressions from structured traces is valuable, but AIQ’s trace contract has changed since this was opened. `develop` now uses NeMo Relay, and its tests intentionally assert semantic AIQ workflow and agent scopes rather than the generic `LangGraph` identity used by these Monocle fixtures.

The two offline tests here also replay fixed trace files, so they validate those fixtures rather than execute the current application. Only the opt-in live test exercises AIQ, and the suite is not integrated into the repository’s locked test environment or CI.

If we add more assembled-agent behavioral coverage, it should be Relay-native: ideally a deterministic current-workflow run that generates ephemeral ATOF/OTEL and asserts semantic agent and tool behavior without adding a second instrumentation runtime. If Monocle Test Tools can consume Relay ATOF or OTLP/OpenInference directly, please point us to that support and we can evaluate a consumer-only approach.

Given the current architecture, we’re going to close this PR alongside #327 as superseded. Thank you again for the contribution and for demonstrating the testing idea.

PR #330, opened 2026-07-13
feat(ui): inline result charts for agent answers

Manushpm8:
## What this does

Lets the agent present numeric answers (rankings, trends, distributions, gains/losses, single metrics) as inline charts instead of walls of text or tables. The agent writes a small JSON chart spec in its reply; the frontend renders it as a styled inline SVG. No code execution and no new runtime dependencies.

## How it works

1. The agent emits a fenced `chart` (or `chart-carousel`) block of JSON in its answer.
2. `MarkdownRenderer` detects that fence and hands it to the new `ResultChart` module.
3. The spec is validated (Zod), normalized, and rendered as an SVG chart.
4. If a spec is malformed, it falls back to a normal code block, so nothing breaks.

Inline result charts (transparent cards, dark theme)

## Chart types

`bar`, `hbar`, `line`, `area`, `grouped-bar`, `delta` (diverging gains/losses), KPI-only cards, and a pageable `chart-carousel`. One answer can contain several charts.

## Correctness and safety

- Safe-by-default axes: magnitude bars start at zero, axis ticks compact large numbers, and rankings that exceed the row cap show a disclosed "top N of M" note (never a silent drop).
- Missing points render as gaps, never as an implied zero.
- Each chart has a "Show data" toggle and a CSV export; the CSV neutralizes spreadsheet formula injection while preserving real negative values.
- The spec's colors are a fixed enum and all text is escaped React content, so an agent-authored spec has no injection surface.

## Where the numbers come from

The renderer is tool agnostic: it charts whatever numbers a tool returns, so it works today with web search and will work with any future data source. A generic chart contract was added to the shallow researcher and deep-research writer prompts (no dataset-specific coupling) and is guarded by a test.

## Testing

- 100% statement, branch, function, and line coverage on the `ResultChart` module, enforced by a scoped Vitest threshold (global coverage is unaffected).
- Type-check, lint, the full frontend test suite, and the prompt-contract pytest all pass.
- Verified end to end in a browser against a live agent: a real query produced a chart spec that rendered as an inline bar chart with the Show data and CSV controls.

## Summary by CodeRabbit

* **New Features**
  * Added interactive markdown result charts: chart + carousel rendering, KPI-only cards, tooltips, legends, and a toggleable “Chart data” table.
  * Added CSV export/download for chart specs with safe filenames and CSV formula injection protection.
* **Documentation**
  * Updated agent prompt guidance and the ResultChart module README to enforce the chart JSON contract and fenced-block rules.
* **Bug Fixes**
  * Improved handling for malformed/unsupported chart fences, bare JSON detection, degenerate charts (KPI-only), and ranking truncation with “Showing top N of M”.
* **Tests**
  * Expanded Vitest coverage for parsing, normalization, rendering, CSV utilities, and toolbar/carousel behaviors, with strict ResultChart coverage thresholds.


src/aiq_agent/agents/deep_researcher/prompts/writer.j2
```diff
@@ -102,6 +102,36 @@ A figure is earned: generate or embed one only when its data is source-anchored
 - Do not place bare URLs in the report body; URLs belong only in the `## Sources` section.
 - Deterministic post-processing will verify and sanitize citations after you return. Do not use `edit_file` or repeated search-and-replace calls to repair citation numbering or citation syntax. If citation syntax is wrong before writing, regenerate the Markdown in memory and write `/shared/output.md` once.
 
+## Presenting Data (Charts)
+When the report presents numbers that compare multiple things (a ranking or top-N across entities, a distribution or counts across categories, a trend over an ordered/time axis, or gains vs losses), present them as a chart, not just prose or a table. Lead with a one-sentence verdict, then the chart.
+
+- Emit the chart as a fenced code block tagged `chart` holding a SINGLE line of valid JSON (no comments, no trailing commas), right after the sentence that introduces it.
+- Chart to reveal a pattern; use a markdown table only to look up exact values the chart does not already show; use prose to state the verdict. Do not repeat the chart's numbers as a table unless the user asks.
```
KyleZheng1284: **[P1] Keep a portable chart fallback.** These prompts also feed PDF/Markdown exports and API/CLI clients, but only the web `MarkdownRenderer` understands `chart`; PDF currently exports the raw JSON. Please gate chart emission on renderer support or retain a readable table/text fallback. The same applies to the shallow prompt.

frontends/ui/src/shared/components/ResultChart/parse.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+/** Coerce a cell value (number, or a numeric-ish string like "11,463" or "$1.2M") to a number. */
+export function toNumber(value: unknown): number | null {
+  if (typeof value === 'number') return Number.isFinite(value) ? value : null
+  if (typeof value === 'string') {
+    const cleaned = value.replace(/[,$\s%]/g, '')
+    if (cleaned === '') return null
+    const parsed = Number(cleaned)
+    return Number.isFinite(parsed) ? parsed : null
```
KyleZheng1284: **[P1] Correct percent-string parsing.** `toNumber("94%")` returns 94, then percent formatting multiplies by 100, so this accepted value renders as 9,400%. Please make percent parsing format-aware or reject `%`-suffixed strings.

frontends/ui/src/shared/components/ResultChart/normalize.ts
```diff
@@ -0,0 +1,92 @@
...
+  const key = spec.series[0].key
+  const fmt = spec.y?.format ?? 'number'
+  const values = numericValues(spec, key)
+
+  if (values.length <= 1) {
+    if (spec.kpis && spec.kpis.length > 0) return spec.kpis
+    const row = spec.data.find((r) => toNumber(r[key]) != null)
+    if (!row) return [{ label: spec.title, value: '–' }]
+    const cat = String(row[spec.x.key] ?? '')
+    return [{ label: spec.title, value: formatValue(values[0], fmt), sub: cat || undefined, tone: 'accent' }]
```
KyleZheng1284: **[P2] Preserve sparse rows.** `values.length <= 1` can still represent several rows with only one numeric value. That collapses the result to a KPI and drops the missing rows plus data/CSV controls. Could we require `spec.data.length === 1`, or preserve access to the underlying rows?

frontends/ui/src/shared/components/ResultChart/parse.ts
```diff
@@ -0,0 +1,98 @@
...
+  const parsed = ChartSpecSchema.safeParse(parseJson(raw))
+  if (!parsed.success) return null
+  const spec = parsed.data
+
+  // The x key must resolve on at least one row, and at least one series must
+  // carry usable numeric data, otherwise there is nothing meaningful to draw.
+  const hasX = spec.data.some((row) => row[spec.x.key] != null && row[spec.x.key] !== '')
+  if (!hasX) return null
+  const hasNumeric = spec.series.some((s) => spec.data.some((row) => toNumber(row[s.key]) != null))
+  if (!hasNumeric) return null
```
KyleZheng1284: **[P2] Reject or remove empty series.** Using `some` accepts a chart when one series works even if another is entirely non-numeric; that empty series still occupies the legend and layout. Please validate every declared series or filter unusable series before rendering.

frontends/ui/src/shared/components/ResultChart/types.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+export const ChartSpecSchema = z.object({
+  type: z.enum(CHART_TYPES),
+  title: z.string().min(1),
+  subtitle: z.string().optional(),
+  x: z.object({ key: z.string().min(1), label: z.string().optional() }),
+  y: z
+    .object({ label: z.string().optional(), format: z.enum(VALUE_FORMATS).optional() })
+    .optional(),
+  series: z.array(SeriesSchema).min(1).max(6),
```
KyleZheng1284, line 39: **[P2] Constrain multi-series delta charts.** Delta bars are colored only by sign and their series legend is hidden, but this schema still permits up to six series. Multi-series delta bars therefore have no visible series identity. Please restrict delta to one series or add a separate series encoding.

frontends/ui/src/shared/components/ResultChart/README.md
```diff
@@ -0,0 +1,134 @@
...
+### Chart carousel
+
+```json
+{
+  "title": "Revenue Forecast: Baseline vs Prediction",
+  "charts": [
+    { "type": "line", "title": "Baseline vs Prediction", "x": { "key": "month" }, "y": { "format": "currency" },
+      "series": [{ "key": "baseline", "color": "neutral" }, { "key": "prediction", "color": "green" }],
+      "data": [ { "month": "Jan", "baseline": 5200000, "prediction": 5300000 } ] }
+  ]
```
KyleZheng1284, line 96: **[P3] Make the carousel example valid.** This example contains one child chart, but `ChartCarouselSpecSchema` requires at least two, so copying it falls back to raw JSON. Please add a second line chart.

tests/aiq_agent/agents/test_result_chart_prompt.py
```diff
@@ -0,0 +1,92 @@
...
+
+@pytest.mark.parametrize("prompt_key", list(_PROMPTS))
+def test_prompt_carousel_examples_match_the_schema(prompt_key: str) -> None:
+    text = _PROMPTS[prompt_key].read_text()
+    blocks = _CAROUSEL_BLOCK.findall(text)
+    assert blocks, f"{prompt_key} prompt has no ```chart-carousel example"
+    for block in blocks:
+        carousel = json.loads(block)
+        assert carousel["title"]
+        assert len(carousel["charts"]) >= 2
```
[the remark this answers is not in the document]
KyleZheng1284: Reopening because this is still present at `2442700c8b0d054db9097c9dae9c4edc3a230562`: the loop checks type/x/series/data, but still does not assert each child chart's required `title`.

KyleZheng1284 (approved):
Should be good to be merged in to get folks unblocked.

Would like to follow up in a different PR to where we could do more of a enforced check for when a sandbox is configured to reject any inline charts in addition to the prompt itself.

PR #332, opened 2026-07-14
feat(ui): chat trace state and streaming plumbing (stacked on #331)

Manushpm8:
> **Stacked on #331.** This branch builds on #331, so until that merges the "Files changed" view here also includes #331's files. Review this PR by its own commit (`feat(ui): add chat trace fields, folded-step parser, and sidebar state`) or the Commits tab.

## What this does

PR 2 of the UI-only chat port. Adds the chat state and streaming plumbing that the visual PR will read. **No user-visible change** and no predictive/analytical/warehouse features.

- **chat types:** optional `ThinkingStep` fields (`completedAt`, `status`, `argSummary`), `ChatMessage.selectedModel`, and a `plan_approval` prompt type.
- **`intermediate-step-parser`:** generic folded reasoning / reflection / explanation sentinels and predicates, `splitPayload` / `extractFoldedOutput`, and label lookup via the shared tool-label engine.
- **`deep-research-trace` (new):** replays a stored deep run into the same `ThinkingStep` stream so a reloaded report shows an identical trace.
- **`store` + `use-websocket-chat`:** populate the new step fields as the run streams, fail steps on error, and thread the selected model through send.
- **layout store:** a sidebar collapse model added *alongside* the existing panel API, a prompt draft, a selected-model slot, and data-source display helpers.

## Preserved (no regression)

Per-user MCP OAuth, WebSocket token rotation, artifact/file handling, and HITL are all untouched. Every existing store/hook/parser test still passes.

## Testing

Full frontend suite green (1479 tests, +59 new), including the pre-existing token-rotation and OAuth specs. `type-check` and `lint` clean.

## Screenshots

This PR is streaming/state plumbing with no UI of its own; it drives the live phased thinking trace (durations, status, response timer) that #333's components render:

Phased thinking trace driven by the streaming plumbing

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Enhanced chat rendering with KaTeX math, Mermaid diagrams from fenced blocks, and answer-mode inline citations with tooltips.
  - Added per-turn model selection (plus a new plan-approval prompt type) and clearer “thinking step” trace folding.
  - Introduced reusable UI primitives for cards, copy-to-clipboard, collapsible sections, and rich sources (list/strip/cards).

- **Bug Fixes**
  - Improved thinking-step success/error handling and reduced-motion behavior for trace status indicators.

- **Tests**
[...]

PR #331, opened 2026-07-13
feat(ui): shared UI foundation (utilities, source/research primitives, markdown upgrade)

Manushpm8:
## What this does

PR 1 of a UI-only chat redesign. Adds the foundational utilities and shared component library that later PRs build on. **No user-visible change yet** and no predictive/analytical/warehouse features — this is purely additive UI plumbing.

### Utilities (`shared/lib`)
- `cn()` — clsx + tailwind-merge className composition.
- `titleCaseWords()` / `ACRONYMS` — humanize snake_case identifiers into labels.
- `AppMotionConfig` — Motion provider that honors `prefers-reduced-motion`.

### Shared components (`shared/components`)
- `Sources/*` — source chips, list, and reference parsing (web + doc kinds).
- `research/*` — `StatusDot`, `ToolCallRow`, and a generic tool-label engine for the live research trace.
- `Surface/Card`, `Actions/CopyButton`, `CollapsibleBlock`, `Mermaid`.
- `MarkdownRenderer` upgrade: an `answer` variant, superscript citations with hover popovers, KaTeX math, and Mermaid diagrams. Prod's artifact-URL handling is preserved verbatim; chart fences fall through unchanged.

Adds `clsx`, `tailwind-merge`, `motion`, `mermaid`, `katex`, `rehype-katex`, `remark-math`, `unist-util-visit`.

## Testing

- Unit and render tests for every new module; the pure logic (`cn`, `humanize`, `parse-references`, `source-utils`, `research-labels`, `rehype-citations`) is covered thoroughly.
- `npm run type-check` and `npm run lint` clean; full frontend suite green (1420 tests).

## Scope

Deliberately additive so it reviews in isolation. Follow-ups add the chat state/streaming plumbing, then the visual chat components that consume these primitives.

## Screenshots

This PR adds the shared rendering primitives; they have no page of their own, so they are shown here rendering an answer in the chat (the surrounding shell/layout is added in #333):

- Markdown with superscript citations + parsed Sources list:

Answer with citations and source list

- Trace rows (StatusDot, ToolCallRow) in the phased spine:

Trace rows: status dots and tool-call rows

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added Mermaid fenced-block rendering with zoom/pan, fullscreen, and theme-aware diagrams.
  * Enhanced Markdown rendering with KaTeX math and answer-mode citations as source chips with tooltips.
[...]

PR #333, opened 2026-07-14
feat(ui): re-skin the chat experience (stacked on #331, #332)

Manushpm8:
> This PR now contains the **full** chat-UI change on its own. The stacked #331 (shared UI foundation) and #332 (chat trace + streaming plumbing) were consolidated here and closed, so review/merge this single PR. Its diff includes all of their files.

## What this does

PR 3 (final) of the UI-only chat port: brings the chat experience to the premium layout. Verified in a browser against a live agent.

- **Live thinking trace** (`ChatThinking`): the intermediate stream folds into a phased spine with status dots, real durations, human tool-call rows, a response timer, and a "using" source chip. Collapses to a summary when done.
- **Answers** (`AgentResponse`): superscript citations with hover popovers, a styled source list, a copy button, and an answer-reveal animation.
- **Interactive clarifier** (`AgentPrompt`): collapsible, with plan approve/reject, clickable options, keyboard shortcuts, and inline free-text.
- **Layout**: push-style panels, turn grouping with a run lane and pinned-to-bottom autoscroll, a restyled composer with a stop control, an AppBar theme toggle and active data-source state, a collapsible sessions rail, and re-skinned research/data-source panels and cards.

## Preserved (no regression)

Per-user MCP OAuth, human-in-the-loop navigation blocking, and rich artifact/file handling are all kept. The model picker is intentionally omitted (prod exposes no selectable models). No predictive/analytical/warehouse UI is ported.

## Testing

- Full frontend suite green (1487 tests); `type-check` and `lint` clean (0 errors).
- Verified end to end in a browser against a live agent: the composer, the phased thinking trace, and answers with superscript citations + styled source list all render as intended.

## Screenshots

The chat on the restored aiq-rfm theme (composer, push panels, NVIDIA Sans base font), and a response summary with the model badge removed (there is no model selector, so the model icon/name is no longer shown next to the timer):

Chat landing on the aiq-rfm palette

Response summary and answer, no model badge

## Consolidated from #331 and #332 (now closed)

This single PR carries the whole chat-UI change. The previously-stacked #331 and #332 were folded in and closed, so their files and review feedback are all here.

**#331 - shared UI foundation:** the component library and utilities the re-skin builds on - the MarkdownRenderer upgrade (superscript citations via rehype-citations, Mermaid diagrams), the Sources components (SourceList / SourceStrip / kind icons), CollapsibleBlock, CopyButton, Surface/Card, the research-trace primitives (StatusDot, ToolCallRow), and lib helpers (cn, humanize, motion).

**#332 - chat trace + streaming plumbing:** the state/streaming layer the visual components read - `ThinkingStep` fields (completedAt / status / argSummary), the `intermediate-step-parser` (folded reasoning / reflection / explanation), the new `deep-research-trace` replay so a reloaded report shows an identical trace, the store + `use-websocket-chat` wiring, and the layout-store sidebar / draft / model state.

The phased thinking trace below is rendered by #331's trace primitives and driven by #332's streaming plumbing:

Thinking trace: foundation primitives + streaming plumbing

## Summary by CodeRabbit

* **New Features**
  * Added interactive clarification and plan-approval prompts with keyboard shortcuts, cancellation, and collapsible previews.
  * Introduced clearer research step timelines, status indicators, source chips, generated-file sections, and auto-following streamed content.
  * Added Markdown enhancements for math, Mermaid diagrams, inline citations, references, and richer answer formatting.
  * Added copy-answer controls, stop-generation action, prompt drafts, model selection, and light/dark theme switching.
  * Redesigned session and data-source panels with collapsible navigation and improved labeling.

* **Bug Fixes**
  * Improved report loading errors, retry behavior, deep-research recovery, message ordering, and thinking-step failure handling.

Signed-off-by: Manush Maheshwari <manushm@nvidia.com>
Signed-off-by: Manush Maheshwari <91221099+Manushpm8@users.noreply.github.com>


frontends/ui/src/features/chat/components/ChatThinking.tsx
```diff
@@ -4,53 +4,549 @@
...
+}
+
+/**
+ * Map a tool/step function name to the data source it belongs to, so the "Using"
+ * chips reflect the sources the answer ACTUALLY used (from its tool calls) rather
+ * than everything that was enabled. Returns null for non-tool steps.
+ */
+const toolNameToSourceId = (functionName: string): string | null => {
+  const name = (functionName || '').toLowerCase().replace(/^tool:\s*/, '')
+  if (name.includes('web_search') || name.includes('tavily')) return 'web_search'
```
cdgamarose-nv: We shouldn't be hard-coding any data sources here. Thinking traces will omit any other user configured data sources.

frontends/ui/src/features/layout/components/ReportTab.tsx
```diff
@@ -33,17 +37,28 @@ interface ReportTabProps {
...
+  const [showFiles, setShowFiles] = useState(false)
+
+  const sources = useMemo(() => {
+    const all = deepResearchCitations ?? []
+    const cited = all.filter((c) => c.isCited)
+    return (cited.length > 0 ? cited : all).map(mapCitationSource)
+  }, [deepResearchCitations])
+
   const reportContentStr = typeof reportContent === 'string' ? reportContent : ''
+  const reportBody = useMemo(() => stripTrailingReferences(reportContentStr), [reportContentStr])
```
cdgamarose-nv: Markdown references are always stripped, but the replacement is rendered only when structured deepResearchCitations exist. Older jobs and compatible agents that emit only Markdown citations will lost them entirely.

frontends/ui/src/features/layout/components/InputArea.tsx
```diff
@@ -178,6 +185,18 @@ export const InputArea: FC<InputAreaProps> = memo(function InputArea({
   const isDisabledByAuth = !isAuthenticated
   const disabled = isDisabledByAuth || ((isBusy || isResearchSessionInProgress) && !isResponseMode)
 
+  // True while a generation is in flight (not a HITL response), used to swap send for Stop
+  const isProcessing = isBusy && !isResponseMode
```
cdgamarose-nv, line 189: `isBusy` comes from `useIsCurrentSessionBusy`, which returns true whenever `isDeepResearchStreaming` is true — a global store value, not scoped to the current session. So when session A runs deep research, session B shows the Stop button. `isResearchSessionInProgress` correctly scopes ownership (so no "Research in progress" popover) but `isProcessing` is still true, so Stop renders. Clicking it calls `disconnect() `on session B's WebSocket, not session A's SSE stream.

cdgamarose-nv (changes requested):

frontends/ui/src/features/chat/components/ChatThinking.tsx
Manushpm8: Fixed. Removed the hard-coded source list; the trace now derives shown sources dynamically from the enabled data sources (toolMatchesSource + usedSources), so any user-configured source is included. Covered by tests.

frontends/ui/src/features/layout/components/ReportTab.tsx
Manushpm8: Fixed. Trailing references are now stripped only when structured citations exist (hasStructuredCitations gate). Markdown-only jobs and agents keep their citations. Covered by tests.

frontends/ui/src/features/layout/components/InputArea.tsx
Manushpm8: Fixed. Busy state is now scoped to the owning conversation (ownsDeepResearch), so session B no longer shows Stop while session A runs deep research. Covered by tests.

exactlyallan:
@Manushpm8 The inline thinking steps is looking really good. 

Some UI specific suggestions:
- Id suggest adding a break line or some sort of indicator to "group" the agent chat responses to the steps indicator for clarity. 
- The cursor pointer css effect is buggy at many points; make sure its applied to the full button and not just text.
- The left panel menu button should open AND close the panel. 
- The left panel close animation "jumps"; Make sure the icons stay fixed to the same position as when the panel is collapsed. Also dont instantly hide the text.
- Enable the tooltips of the sessions title.
- If the data panel is open and the report tab is clicked, close the data panel first.
- The research panel is missing the left border. 
- Would be good to through and make all lines and toggle consistent with the newer style.

FWIW I had some WIP designs to update the research panel nav to be more concise and similar to the left panel. Ideas are attached for reference, though would make sense in a follow up PR.

Screenshot 2026-08-14 at 1 03 14 PM
Screenshot 2026-08-14 at 1 02 51 PM
Screenshot 2026-08-14 at 1 02 25 PM

frontends/ui/src/features/layout/components/ReportTab.tsx
```diff
@@ -33,17 +37,32 @@ interface ReportTabProps {
...
     })))
   // Resolve the owning job id (active or latest finished) so artifact:// images render.
   const deepResearchJobId = useChatStore(selectResolvedDeepResearchJobId)
 
+  const [showFiles, setShowFiles] = useState(false)
+
+  const sources = useMemo(() => {
+    const all = deepResearchCitations ?? []
+    const cited = all.filter((c) => c.isCited)
+    return (cited.length > 0 ? cited : all).map(mapCitationSource)
```
cdgamarose-nv: Citations can point to the wrong source.

Imagine the research process discovers:

  - Source A first
  - Source B second

  But the finished report uses:

  - [1] Source B
  - [2] Source A

  The UI ignores the finished report’s numbering and rebuilds the list using discovery order. Clicking [1] can therefore open Source A even though the report meant Source B.

  It gets worse when the report mixes websites and internal documents: once the UI sees a website citation, it removes the report’s original Sources section. Internal-document sources
  may then disappear entirely.

  Impact: users can be shown incorrect evidence for a claim.

frontends/ui/src/features/chat/store.ts
```diff
@@ -1224,10 +1230,7 @@ export const useChatStore = create<ChatStore>()(
 
         getThinkingStepsForMessage: (userMessageId: string) => {
           const { thinkingSteps } = get()
-          // Filter out deep research steps - they're displayed in the Research Panel, not ChatThinking
-          return thinkingSteps.filter(
-            (step) => step.userMessageId === userMessageId && !step.isDeepResearch
-          )
+          return thinkingSteps.filter((step) => step.userMessageId === userMessageId)
```
cdgamarose-nv, line 1233: Deep research can run several workers simultaneously, all named researcher-agent. The UI tracks them by that shared name instead of their unique IDs.

  It is like tracking two employees using only the name “Researcher.” When one finishes, the UI may mark the other as finished, attach output to the wrong row, or leave one permanently
  “running.”

  Impact: the research result is probably still correct, but the progress display can lie about what happened.

frontends/ui/src/shared/components/Sources/parse-references.ts
```diff
@@ -0,0 +1,269 @@
...
+      inOrdered = false
+      continue
+    }
+    if (TOP_ORDERED_RE.test(line)) {
+      inOrdered = true
+      continue
+    }
+    const bullet = line.match(BULLET_TOP_RE)
+    if (bullet) {
+      if (inOrdered && bullet[1].length < 3) lines[i] = `   ${line.replace(/^\s+/, '')}`
```
cdgamarose-nv: Blank lines never clear inOrdered, so valid Markdown such as `
```
1. First
2. Second

- Separate topic 

```
is rewritten with three spaces and rendered as a child of item 2. That changes the answer’s meaning instead of merely normalizing malformed output. Determine nesting from actual Markdown block boundaries, or avoid rewriting valid blocks, and add a blank-line-separated top-level bullet regression.

cdgamarose-nv (changes requested):

frontends/ui/package.json
```diff
@@ -24,15 +24,23 @@
     "@nvidia/foundations-react-core": "^0.600.0",
     "@react-pdf/renderer": "^4.3.2",
     "@tailwindcss/postcss": "^4.1.18",
+    "clsx": "^2.1.1",
     "http-proxy": "^1.18.1",
+    "katex": "^0.17.0",
     "marked": "^16.4.2",
+    "mermaid": "^11.16.0",
```
cdgamarose-nv: Upgrade Mermaid before rendering agent-generated diagrams

  The lockfile installs Mermaid 11.16.0. `npm audit` reports five advisories fixed in 11.16.1, including CSS injection outside the diagram and browser denial-of-service through infinite
  rendering loops. Because agent-generated Mermaid is rendered and inserted into the page, this input cannot be treated as trusted. Upgrade the lockfile to at least 11.16.1 and require
  `npm audit --omit=dev` to pass.

  This is especially relevant because the component inserts Mermaid’s generated SVG using dangerouslySetInnerHTML.

frontends/ui/src/features/layout/components/ChatArea.tsx
```diff
@@ -109,77 +216,108 @@ export const ChatArea: FC<ChatAreaProps> = memo(function ChatArea({ isAuthentica
...
+                      <ChatThinking
+                        steps={messageSteps}
+                        isThinking={isStreaming && userMessage.id === currentUserMessageId}
+                        isWaiting={isWaiting}
+                        isInterrupted={isInterrupted}
+                        enabledDataSources={userMessage.enabledDataSources}
+                        messageFiles={userMessage.messageFiles}
+                        model={userMessage.selectedModel}
+                        responseStartedAt={userMessage.timestamp}
+                        responseCompletedAt={completedResponse?.timestamp}
```
cdgamarose-nv: The hidden deep-research tracking message is created when the job starts. When the job finishes, its status is updated but its timestamp is not, and this line then treats that
  original timestamp as the completion time. A five-minute research job can therefore display a duration of only a few seconds. Store a real completion timestamp when the terminal event
  arrives and add a multi-minute-job regression test.

frontends/ui/src/features/layout/components/ReportTab.tsx
cdgamarose-nv: For the fix, please make the verified final report metadata the single source of truth for citation number, URL, and document key. Rebuilding final references from discovery-time  citation events will remain order-dependent and incomplete.

frontends/ui/src/features/chat/store.ts
cdgamarose-nv: Please address this by correlating workflow, tool, and LLM events using their unique event/agent/run IDs. Avoid adding more name-based matching conditions. I also recommend extracting this correlation from the compressed callbacks in use-deep-research.ts into named, directly testable handlers.

frontends/ui/src/shared/components/Sources/parse-references.ts
cdgamarose-nv: Maybe we can stop guessing at Markdown intent. Remove `nestBulletsUnderOrderedItems` from the rendering path rather than adding another special case. The existing `renumberOrderedLists` behavior should also be reconsidered because it similarly changes intentionally separate lists.

cdgamarose-nv (changes requested):
Please fix the requested changes by simplifying these paths rather than adding more heuristic conditions. Citations should come from authoritative final metadata, streaming events should be keyed by unique IDs, and valid Markdown should render as authored. If decomposing ChatThinking/use-deep-research is too large for this PR, please file focused follow-up issues rather than extending the current name/token/regex matching layers.

frontends/ui/src/features/layout/components/ReportTab.tsx
Manushpm8: Fixed in 0cb9e8c: the report's own [N] numbering (parsed via splitReferences) is now the source of truth for number/URL/key; the discovery-order list is only a fallback and is keyed by marker number, not array position. Tests added.

frontends/ui/src/features/chat/store.ts
Manushpm8: Fixed in 8192c94: correlation now keys on unique agent/run ids via an extracted, directly-testable deep-research-correlation module, so concurrent researcher-agent workers stay distinct (no more name-based matching). Tests added.

frontends/ui/package.json
Manushpm8: Fixed in 41c2551: bumped Mermaid to 11.16.1 and regenerated the lockfile, clearing the five advisories.

frontends/ui/src/features/layout/components/ChatArea.tsx
Manushpm8: Fixed in 8192c94: a real completion timestamp (responseCompletedAt) is stored on the terminal event and used for the duration, so multi-minute jobs report the correct elapsed time. Tests added.

frontends/ui/src/shared/components/Sources/parse-references.ts
Manushpm8: Fixed in 0cb9e8c: removed nestBulletsUnderOrderedItems and renumberOrderedLists from the render path entirely, so valid Markdown is no longer rewritten. Regression test for your exact case added.

Manushpm8:
Fixed in the latest commits: citations now come from the report's authoritative [N] metadata, streaming events key on unique agent ids (extracted into a testable correlation module), and the Markdown-rewriting heuristics were removed so valid Markdown renders as authored. No new name/token/regex matching was added. Thanks for the feedback!

Manushpm8:
@exactlyallan thanks for the review! Handled the quick ones here: full-button cursor, the left menu now toggles open and close, research-panel left border, session-title tooltips, and the report tab already closes the data panel. The group indicator, close-animation polish, consistent lines/toggles, and your nav redesign I'll take in a follow-up PR. Appreciate the mockups!

cdgamarose-nv (approved):
Looks good! Thanks for the UI revamp!
There is a test that's failing; fix that and they its good to merge.

PR #336, opened 2026-07-15
fix(knowledge): stabilize Azure Search file lifecycle

harmke:
#### Overview

Azure AI Search is eventually consistent. Upload and deletion results can therefore appear stale immediately after successful writes, causing premature job completion, incomplete cleanup, or inconsistent file state.

This PR:

- Deletes failed-upload chunks using known deterministic chunk IDs, including chunks that have not yet become searchable.
- Waits for stable manifest and chunk visibility before marking an upload job complete.
- Deletes file chunks using deterministic IDs instead of relying on an immediately consistent search.
- Requires three consecutive collection-scoped zero-result reads covering both the file manifest and chunks before reporting deletion success.
- Updates summary, tombstone, and local file bookkeeping only after deletion is confirmed, allowing an in-process retry after a confirmation timeout.
- Adds regression coverage for delayed visibility, stable-read confirmation, collection isolation, and retryable deletion bookkeeping.

The scope is limited to the Azure AI Search adapter and its tests. No API contract or configuration changed, so no documentation update is required.

Known bounded limitations remain: stable reads are a consistency heuristic; confirmation-timeout retry is guaranteed only within the current process and, after restart, depends on the file manifest still being searchable.

#### Validation

```text
.venv/bin/ruff check \
  sources/knowledge_layer/src/azure_ai_search/adapter.py \
  tests/knowledge_layer_tests/test_azure_ai_search.py
Passed

.venv/bin/ruff format --check \
  sources/knowledge_layer/src/azure_ai_search/adapter.py \
  tests/knowledge_layer_tests/test_azure_ai_search.py
Passed

git diff --check
Passed

.venv/bin/pytest \
  -p no:cacheprovider \
  tests/knowledge_layer_tests/test_azure_ai_search.py
56 passed in 5.81s
```

Strict live Azure lifecycle validation was repeated three times with fresh collections. All three runs passed, verifying immediate delete/list/query behavior, cross-collection isolation, deletion of the remaining file, and complete collection cleanup.

- [ ] I updated documentation for user-facing or contributor-facing changes.

Start with `sources/knowledge_layer/src/azure_ai_search/adapter.py`, especially the failed-upload rollback, stable visibility checks, and `delete_file` bookkeeping order.

Regression coverage is in `tests/knowledge_layer_tests/test_azure_ai_search.py`.

#### Related Issues

- Relates to #308

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Azure AI Search ingestion reliability under eventual consistency with stabilized polling for terminal manifest+chunk visibility and exact stable document counts.
  * Finalization is now visibility-checked; finalize failures trigger deterministic rollback (including chunk docs) and correct FAILED/SUCCESS state handling.
  * Deterministic chunk/document identifiers and consistency-safe deletes (including safe precomputed rollback after post-upload failures).
  * Collection deletion now fences concurrent submissions and rejects new jobs when the collection is no longer active.
* **Tests**
  * Upgraded the Azure AI Search fake backend to model delayed/stale visibility and “response lost” failures; added/expanded lifecycle, restart/delete, rollback, and bounds-based polling assertions.


sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -833,6 +876,29 @@ def get_job_status(self, job_id: str) -> IngestionJobStatus:
...
+                label=f"file {detail.file_id!r} manifest",
+                stable_reads=_CONSISTENCY_STABLE_READS,
+            )
+            self._wait_for_search_count(
+                _record_filter(_RECORD_CHUNK, job.collection_name, file_id=detail.file_id),
+                expected_count=detail.chunks_created,
+                label=f"file {detail.file_id!r} chunks",
+                stable_reads=_CONSISTENCY_STABLE_READS,
+            )
+
```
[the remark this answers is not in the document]
harmke: Good point. The per-file checks are intentional so each file's status and chunk count are verified independently. I'd prefer to keep this PR focused on correctness and handle batching separately if the latency becomes an issue.

sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -971,13 +1039,22 @@ def _process_file(
 
         client = self._get_validated_search_client()
         self._upload_documents(client, search_documents)
```
KyleZheng1284: Could we compute `chunk_ids` before calling `_upload_documents()` and include the upload in the rollback block? I reproduced a response-loss case where Azure accepted both chunks, the client raised before returning, and this PR left both chunks searchable while the job was marked failed; the base branch cleaned them up. Since these IDs are deterministic, deleting the full attempted ID set on any upload or post-upload failure should close the gap.

sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -918,10 +980,16 @@ def _process_job(
                     self._cleanup_paths([path])
                 self._update_job(job_id, processed_files=index + 1)
 
-        if failed == len(file_paths):
-            self._fail_job(job_id, f"All {failed} file(s) failed to ingest")
+        try:
+            self._wait_for_job_visibility(job_id)
+        except Exception as error:  # noqa: BLE001
+            self._fail_job(job_id, f"Failed to finalize ingestion: {self._translate_error(error)}")
```
KyleZheng1284: Could we make the finalization-timeout state consistent? I reproduced `_wait_for_job_visibility()` timing out after a successful upload: the job became terminal `FAILED`, while its file and chunks remained `SUCCESS`. A client retrying the failed job can then ingest the same file again. Could we keep finalization retryable/non-terminal, or reconcile the successful files before failing the job, with a test for the chosen contract?

sources/knowledge_layer/src/azure_ai_search/adapter.py
```diff
@@ -1185,13 +1285,17 @@ def delete_collection(self, name: str) -> bool:
...
-        ]
+        }
+        for info in files.values():
+            document_ids.add(_record_id(_RECORD_FILE, name, info.file_id))
+            if info.status == FileStatus.SUCCESS:
+                document_ids.update(_chunk_id(info.file_id, index) for index in range(info.chunk_count))
         if document_ids:
-            self._delete_document_ids(client, document_ids)
+            self._delete_document_ids(client, sorted(document_ids))
         self._delete_document_ids(client, [str(manifest["id"])])
```
KyleZheng1284, line 1396: Could we keep the `DELETING` manifest and repeat the discover/delete pass until the collection is stably empty before removing the manifest? The `_files` overlay only covers this process. I reproduced a restart/second-replica case where a file manifest and chunk were not searchable during this pass: `delete_collection()` returned `True`, removed the collection manifest, and the child documents later remained with no retry anchor.

KyleZheng1284 (commented):
I revalidated these lifecycle edge cases against the current PR head (`13bda78`). Suggestions are inline.

sources/knowledge_layer/src/azure_ai_search/adapter.py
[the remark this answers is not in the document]
KyleZheng1284: I checked the timing here: with three stable reads at 250 ms, these two waits add about 1 second per file on the happy path and up to 9.5 seconds per file on timeout, before Azure request time. Could we preserve the per-file checks with one shared polling loop and per-file counters, or otherwise cap the batch size and track the batching follow-up before resolving this?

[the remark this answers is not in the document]
harmke: @KyleZheng1284 I revisited `_wait_for_job_visibility()` and changed it to check all files through one shared polling loop, while still tracking manifest and chunk visibility per file. We now check all files, wait once, and check them all again instead of waiting separately for every file. This reduces the happy path consistency wait from 1 second per file to 1 second for the entire job.

sources/knowledge_layer/src/azure_ai_search/adapter.py
harmke: Fixed in latest commit. We now record every attempted chunk ID before calling Azure and delete the full attempted set if the upload raises. This also covers the case where Azure accepted the documents but the response was lost.

sources/knowledge_layer/src/azure_ai_search/adapter.py
harmke: Fixed by reconciling the successful files before failing the job. If final visibility confirmation times out, we delete those files, mark their progress as failed, and then fail the job. This prevents a retry from leaving the previous successful copy behind. I also added a test case for this.

sources/knowledge_layer/src/azure_ai_search/adapter.py
harmke: Implemented a bounded version of this in [483291f](https://github.com/NVIDIA-AI-Blueprints/aiq/blob/483291f74de50b2e169e3922761fae18f3580005/sources/knowledge_layer/src/azure_ai_search/adapter.py#L1313-L1370). The collection remains DELETING while we get a stable view of its children, delete them, and require three consecutive zero-result reads before removing the collection manifest. If that confirmation times out, the manifest remains available for retry instead of reporting false success. Test case also added.

KyleZheng1284 (approved):
LGTM

PR #371, opened 2026-07-17
feat(knowledge): add NeMo Retriever Library as a knowledge layer backend

KyleZheng1284:
#### Overview

Add two separately registered NeMo Retriever Knowledge Layer backends to AI-Q:

| Backend | Runtime boundary | Intended use |
|---|---|---|
| `nemo_retriever` | AI-Q calls an independently deployed NeMo Retriever public REST API | Shared Docker Compose, Helm/Kubernetes, or remote enterprise deployments |
| `nemo_retriever_local` | AI-Q, the NeMo Retriever Python library, and embedded LanceDB run in one Python 3.12 process | Experimental laptop-oriented zero deployment with remote inference endpoints |

Both backends implement AI-Q's collection, document, ingestion-job, retrieval, and universal-schema contracts while preserving NeMo Retriever's extraction, embedding, storage, ranking, and native-distance semantics.

#### What changes

- Adds typed service and embedded adapters with distinct backend registrations, preventing cached REST and local instances from being mixed.
- Adds collection and document CRUD, deterministic multi-file ingestion, stable identities, restart reconciliation, deletion, and citation-ready chunk normalization.
- Makes service submission non-blocking: the upstream job ID is returned after job creation while multipart uploads continue on a bounded background executor.
- Bounds service admission with `max_concurrency + max_queued_uploads`; oversized batches return HTTP 413 and temporary saturation returns HTTP 503 before an NRL job is created.
- Preserves NRL's native vector distance in returned chunks. Lower is closer; AI-Q does not manufacture a cross-metric 0–1 score.
- Moves NeMo Retriever settings under backend-owned `backend_config` fields and retains credentials as `SecretStr` until the immediate inference or transport boundary.
- Adds shipped service and local workflow profiles plus operator documentation covering setup, configuration, file formats, security, lifecycle, and troubleshooting.

#### Deployed REST backend

`nemo_retriever` uses NRL's public collection, document, job-scoped ingestion, health, and query routes. AI-Q owns transport behavior, universal-schema conversion, and process-local upload tracking; the deployed NRL stack owns extraction, OCR, tokenization, embedding, indexing, physical VectorDB storage, and collection durability.

The service contract was validated against immutable NRL integration revision [`f3a0b418b7250fa8823ec44dea569b07e2b008cb`](https://github.com/NVIDIA/NeMo-Retriever/commit/f3a0b418b7250fa8823ec44dea569b07e2b008cb), or a released successor containing the same collection-management and tokenizer behavior.

Use `configs/config_web_nemo_retriever.yml`. The public gateway may be local, SSH-forwarded, or exposed through an enterprise ingress. TLS verification remains enabled by default, with optional enterprise CA and bearer-token configuration.

#### Embedded zero-deployment backend

`nemo_retriever_local` does not start an NRL HTTP service, VectorDB server, Docker Compose stack, or Helm deployment. One isolated environment contains editable AI-Q packages plus an unmodified NeMo Retriever checkout pinned to commit `c80f4a5189ee10b98cbdb93e2f853ceb7b699c3b`:

```bash
uv sync --project environments/nemo_retriever_local --frozen
uv run --project environments/nemo_retriever_local --frozen \
  nat serve --config_file configs/config_web_nemo_retriever_local.yml --port 8000
```

The isolated project is constrained to Python `>=3.12,<3.13`; it does not add NeMo Retriever to the normal AI-Q environment. `--frozen` installs the checked-in resolution without changing or re-resolving the lock.

AI-Q owns temporary staging, job/status models, bounded work admission, lifecycle, and universal-schema normalization. NeMo Retriever owns extraction planning and graph construction, inference helpers, canonical schemas, `CollectionWriteContext`, collection catalog and recovery, chunk identity, physical table selection, VDB writes, retrieval, ranking, and distance. AI-Q calls NRL's `IngestVdbOperator`, `RetrieveVdbOperator`, and collection APIs; it does not issue direct LanceDB searches, reproduce NRL schemas, or expose physical table identifiers.

The shipped profile uses threaded Dask workers and inline full research so AI-Q, NRL, LanceDB, and the data-directory lock remain in one process. Document ingestion remains asynchronous. Detached durable research jobs remain a service-backend capability.

Local extraction profiles:

| Profile | Behavior |
|---|---|
| `fast-text` | Text-only extraction and embedding; lowest extraction overhead |
| `auto` | NRL's complete default extraction plan, including text and configured multimodal extraction paths |

`NRL_INFERENCE_API_KEY` configures NRL extraction and embedding inference. The AI-Q agent LLM can use a separate OpenAI-compatible endpoint and credential, such as `NVIDIA_API_KEY`. Document ingestion does not require a generative LLM; normal research retrieves through the registered `knowledge_search` tool and requires a configured agent LLM.

#### Upload and API contracts

- The global upload allowlist remains backward compatible. Enable PowerPoint explicitly with `FILE_UPLOAD_ACCEPTED_TYPES=.pdf,.docx,.pptx,.txt,.md` in the shared frontend/backend environment.
- Pinned NRL plans `.pptx` through its document/PDF branch for both local profiles; DOCX/PPTX conversion requires LibreOffice on `PATH`.
- Request validation is atomic. One unsupported or malformed file rejects the complete request with HTTP 415 and creates no ingestion job.
- Failures after job acceptance remain visible per file and may produce partial success.
- Metadata is included in the default idempotency identity so changed metadata cannot silently reuse stale work.

#### Compatibility and non-goals

- No NeMo Retriever repository modification, vendoring, or patching is included.
- Existing LlamaIndex, Foundational RAG, OpenSearch, and Azure AI Search configuration and runtime paths are unchanged.
- No extraction, embedding, ranking, native-distance, NRL schema, or physical-storage semantics are reimplemented or changed.
- This PR does not add a public headless retrieval endpoint; chat invokes the registered retriever through `knowledge_search`.
- Embedded mode is experimental and single-process. Multiple AI-Q processes sharing one local data directory, Python 3.13, Intel macOS, and local laptop GPU inference are outside this PR.
- Platform promotion requires green Linux, Windows x64, and Apple Silicon CI plus the documented restart, deletion, idempotency, and lock tests.

#### Validation

Automated and local validation on Apple Silicon macOS:

- [x] Focused upload, service-adapter, and local-adapter suites in the normal AI-Q environment: `124 passed, 3 skipped` (NRL is intentionally absent there).
- [x] Isolated Python 3.12 NeMo Retriever suite: `48 passed`, including embedded LanceDB, backend-level retrieval, restart/reconciliation, and native server lifecycle.
- [x] Native credential-free `nat serve` E2E: create collection, ingest, poll, delete document, and delete collection, without an NRL HTTP service or inference credentials.
- [x] Full root Python suite completed successfully.
- [x] Both shipped NeMo Retriever YAML profiles: `0 errors, 0 warnings` from the AI-Q workflow validator.
- [x] Scoped Ruff check and format validation.
- [x] Root and isolated `uv lock --check` validation.
- [x] Documentation build completed; only two pre-existing MCP cross-reference warnings remained.
- [x] `actionlint`, `git diff --check`, and required commit hooks passed, including secret detection and Markdown link validation.

Previously validated against a separately deployed NRL stack:

- [x] Connected AI-Q on macOS to the public NRL gateway.
- [x] Ingested single and multiple PDFs into session collections and retrieved grounded chunks through `knowledge_search`.
- [x] Deleted documents and verified NRL removed the corresponding logical document data.
- [x] Reused an existing collection for subsequent ingestion and retrieval.

Pending before promotion of documented platform support:

- [ ] GitHub Linux, Windows x64, and Apple Silicon matrix checks on the final pushed head.
- [ ] Optional live full-chat rerun against a configured tool-calling OpenAI-compatible LLM; no-LLM backend retrieval remains covered by direct Python integration.

Signed-off-by: Kyle Zheng <126034466+KyleZheng1284@users.noreply.github.com>
Signed-off-by: Kyle Zheng <kyzheng@nvidia.com>

- [x] I certify this contribution under the Developer Certificate of Origin (DCO).

1. `sources/knowledge_layer/src/nemo_retriever/README.md` and the two shipped configs for the service/local operating model.
2. `sources/knowledge_layer/src/nemo_retriever/adapter.py` for REST CRUD, background uploads, admission, status, and retrieval.
3. `sources/knowledge_layer/src/nemo_retriever/_local_client.py` and `local_adapter.py` for embedded NRL orchestration and lifecycle.
4. `_transport.py`, `_models.py`, and `_normalization.py` for the public wire contract, retries, redaction, and universal chunk mapping.
5. `sources/knowledge_layer/src/register.py` for backend-owned configuration and NAT lifecycle finalization.
6. The three NeMo Retriever test modules for service, local, and native integration coverage.

The main design boundary is deliberate: AI-Q provides a consistent Knowledge Layer contract while NeMo Retriever continues to own its operators, collection model, VDB behavior, ranking, and physical storage.

#### Related Issues

- None.

## Summary by CodeRabbit

- **New Features**
  - Added remote and experimental local NeMo Retriever knowledge backends.
  - Added document ingestion, collection management, asynchronous jobs, status tracking, filtering, citations, and distance metadata.
  - Added hosted and zero-deployment research configurations.

- **Bug Fixes**
  - Improved non-blocking uploads and collection operations.
  - Improved retry handling, error reporting, cleanup, and expired-job responses.
  - Added clearer handling for oversized or temporarily unavailable ingestion capacity.

- **Documentation**
  - Expanded setup, configuration, compatibility, operation, and troubleshooting guidance.


sources/knowledge_layer/src/nemo_retriever/_transport.py
```diff
@@ -0,0 +1,224 @@
...
+            self._redact(f"NeMo Retriever {operation} failed with HTTP {status}: {reason}"),
+            status_code=status,
+        )
+
+    def _should_retry(self, response: httpx.Response, attempt: int) -> bool:
+        return response.status_code in _RETRYABLE_STATUS_CODES and attempt < self._max_retries
+
+    def _delay(self, response: httpx.Response, attempt: int) -> float:
+        retry_after = _retry_after_seconds(response.headers.get("Retry-After"))
+        return retry_after if retry_after is not None else self._backoff(attempt)
```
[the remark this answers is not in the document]
charlesbluca: Confirmed—valid on `1c73ba49`. The local successor clamps both exponential backoff and server-provided `Retry-After` values to 30 seconds. Please keep this open until the change lands.

tests/knowledge_layer_tests/test_nemo_retriever_live.py
```diff
@@ -0,0 +1,159 @@
...
+    os.environ.get("AIQ_NRL_LIVE_TESTS") != "1",
+    reason="Set AIQ_NRL_LIVE_TESTS=1 to run against a live NeMo Retriever deployment",
+)
+
+
+def _bool_env(name: str, default: bool) -> bool:
+    value = os.environ.get(name)
+    if value is None:
+        return default
+    return value.lower() in {"1", "true", "yes", "on"}
```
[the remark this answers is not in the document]
charlesbluca: Confirmed—valid on `1c73ba49`. The local successor accepts explicit true/false spellings and raises on unknown values instead of silently treating a typo as disabled certificate verification.

sources/knowledge_layer/src/nemo_retriever/_transport.py
```diff
@@ -0,0 +1,224 @@
...
+                    raise NemoRetrieverTransportError(
+                        self._redact(
+                            f"NeMo Retriever {operation} transport failure after retries: {type(error).__name__}"
+                        )
+                    ) from error
+                time.sleep(self._backoff(attempt))
+                continue
+            if self._should_retry(response, attempt):
+                time.sleep(self._delay(response, attempt))
+                continue
```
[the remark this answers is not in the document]
charlesbluca: Confirmed—valid on `1c73ba49`. The local successor retries safe methods by default and permits write retries only when the caller supplies an explicit idempotency guarantee. The gate applies to both transport failures and retryable HTTP responses.

sources/knowledge_layer/src/nemo_retriever/adapter.py
```diff
@@ -0,0 +1,806 @@
...
+            _FileDescriptor(
+                position=position,
+                path=path,
+                filename=filename,
+                file_size=path.stat().st_size,
+                content_sha256=content_sha256,
+                manifest_entry_id=manifest_entry_id,
+            )
+        )
+    return descriptors
```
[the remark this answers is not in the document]
charlesbluca: Confirmed—valid on `1c73ba49`. The local successor snapshots each source into a temporary immutable file and derives size, digest, manifest identity, and uploaded bytes from that snapshot. This is separate from the submission-blocking issue raised in this review.

sources/knowledge_layer/src/nemo_retriever/README.md
```diff
@@ -0,0 +1,156 @@
...
+```bash
+./scripts/setup.sh
+source .venv/bin/activate
+
+uv run python .agents/skills/aiq-configure-workflow/scripts/validate_config.py \
+  configs/config_web_nemo_retriever.yml
+
+./scripts/start_e2e.sh \
+  --config_file configs/config_web_nemo_retriever.yml \
+  2>&1 | tee /private/tmp/aiq-nrl-e2e.log
```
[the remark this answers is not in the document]
charlesbluca: Confirmed. The local successor changes the example to `/tmp/aiq-nrl-e2e.log`, which works on both Linux and macOS.

tests/knowledge_layer_tests/run_adapter_compliance.py
```diff
@@ -33,6 +35,10 @@
     python tests/knowledge_layer_tests/run_adapter_compliance.py --backend azure_ai_search \
       --config '{"start_ttl_cleanup":false}'
 
+    python tests/knowledge_layer_tests/run_adapter_compliance.py --backend nemo_retriever \
+      --config '{"base_url":"http://127.0.0.1:17670","api_token":"...",\
+                 "scope":"workspace-123"}'
```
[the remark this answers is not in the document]
charlesbluca: Confirmed—valid on `1c73ba49`. The local successor loads `NRL_API_TOKEN` from the environment as `SecretStr`, rejects tokens supplied through CLI JSON, redacts configuration output, and wraps the live-test token as a secret.

sources/knowledge_layer/src/nemo_retriever/adapter.py
```diff
@@ -0,0 +1,806 @@
...
+            {
+                "position": item.position,
+                "filename": item.filename,
+                "content_sha256": item.content_sha256,
+            }
+            for item in descriptors
+        ],
+    }
+    encoded = json.dumps(canonical, separators=(",", ":"), sort_keys=True).encode("utf-8")
+    return f"aiq-{hashlib.sha256(encoded).hexdigest()}"
```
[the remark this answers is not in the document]
charlesbluca: Confirmed—valid on `1c73ba49`. The local successor canonicalizes the scrubbed metadata before job creation and includes that serialization in the default idempotency key, so identical bytes with different metadata no longer collide.

sources/knowledge_layer/src/nemo_retriever/_transport.py
```diff
@@ -0,0 +1,224 @@
...
+        self._owns_client = client is None
+        headers = {"Accept": "application/json", "X-NRL-Scope": scope}
+        if api_token:
+            headers["Authorization"] = f"Bearer {api_token}"
+        timeout = httpx.Timeout(request_timeout_s, connect=connect_timeout_s)
+        limits = httpx.Limits(max_connections=100, max_keepalive_connections=20)
+        verify = _tls_verify(verify_ssl, ca_bundle)
+        self._client = client or httpx.Client(headers=headers, timeout=timeout, limits=limits, verify=verify)
+        if client is not None:
+            self._client.headers.update(headers)
```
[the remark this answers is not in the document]
charlesbluca: Confirmed—valid on `1c73ba49`. The local successor keeps authorization and scope headers transport-local and applies them per request instead of mutating an injected client default headers. This should remain open until that patch is pushed.

sources/knowledge_layer/src/nemo_retriever/_models.py
```diff
@@ -0,0 +1,167 @@
...
+    offset: int
+    limit: int
+    items: list[JobDocumentWire] = Field(default_factory=list)
+
+
+class QueryHitWire(_WireModel):
+    chunk_id: str
+    document_id: str
+    text: str
+    score: float
```
charlesbluca: **Blocking paired-contract mismatch:** NRL `69581680` returns `ranking` metadata and does not return this required `score` field. Runtime validation confirms the native score fields are absent, so `QueryHitWire` rejects a valid NRL query response before normalization. Model the NRL ranking object and convert the complete result set to AIQ's `[0,1]` `Chunk.score` using its declared direction. The local successor's direction-aware conversion demonstrates the required behavior.

sources/knowledge_layer/src/nemo_retriever/adapter.py
```diff
@@ -0,0 +1,806 @@
...
+                metadata=json.dumps(metadata, separators=(",", ":"), sort_keys=True),
+            )
+            return _wire(UploadAcceptedWire, payload, "job document upload")
+
+        accepted_by_position: dict[int, UploadAcceptedWire] = {}
+        try:
+            with ThreadPoolExecutor(max_workers=min(self._settings.max_concurrency, len(descriptors))) as executor:
+                futures = {executor.submit(upload, item): item.position for item in descriptors}
+                for future in as_completed(futures):
+                    accepted_by_position[futures[future]] = future.result()
```
charlesbluca: **Blocking:** `submit_job()` waits for every upload future before returning the job ID. The async submission route calls this synchronous method directly, so a large or stalled upload blocks the request—and potentially the event loop—despite the ingestion interface representing asynchronous jobs. Return after creating/registering the job and perform uploads in tracked background execution, with a test proving submission returns before a blocked upload completes.

sources/knowledge_layer/src/nemo_retriever/_transport.py
```diff
@@ -0,0 +1,224 @@
...
+    @staticmethod
+    def _backoff(attempt: int) -> float:
+        return min(0.5 * (2**attempt), 30.0)
+
+    def _response_error(self, response: httpx.Response, operation: str) -> NemoRetrieverHTTPError:
+        status = response.status_code
+        if operation.startswith("job ") and status in {404, 410}:
+            return NemoRetrieverCompatibilityError(
+                "NeMo Retriever rejected a job-scoped API route with "
+                f"HTTP {status}. Confirm AIQ and the NRL service use compatible collection-management API versions.",
```
charlesbluca, line 134: A 404 or 410 from any operation whose name starts with `job ` is not necessarily an API-version mismatch. Status polling for an unknown or expired job legitimately returns those statuses, but this converts them into upgrade guidance. Detect compatibility through an explicit capability/version check, or limit this mapping to confirmed endpoint absence, and preserve normal not-found/expired resource errors.

configs/config_web_nemo_retriever.yml
```diff
@@ -0,0 +1,186 @@
...
+        name: "Web Search"
+        description: "Search the web for real-time information."
+        tools:
+          - web_search_tool
+          - advanced_web_search_tool
+      - id: knowledge_layer
+        name: "Knowledge Base"
+        description: "Search documents managed by NeMo Retriever."
+        tools:
+          - knowledge_search
```
charlesbluca, line 129: **Blocking configuration path:** this advertises `knowledge_layer` as an explicitly selectable data source, but an async submission selecting that ID returns 422 in the deployed configuration. The same agent successfully invokes `knowledge_search` when no source filter is supplied, so the tool itself is registered but the explicit-selection path is not. Please make the registry ID resolvable by job submission and add a filtered-submission integration test.

charlesbluca (dismissed):
Requesting changes for four remaining paired-contract blockers: the published query wire model does not accept NRL ranking metadata, `submit_job()` blocks through all uploads, job-resource 404/410 responses are over-classified as API incompatibility, and explicit `knowledge_layer` selection fails despite the unfiltered agent using the tool successfully.

The eight existing automated findings are valid and appear addressed in the local successor patch, but should remain open until those changes are pushed. The focused adapter suite passed 31 tests with 1 skip; the full knowledge-layer suite passed 184 tests with 7 skips.

sources/knowledge_layer/src/nemo_retriever/_models.py
```diff
@@ -0,0 +1,167 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
peterychang, line 1: These schemas need to be kept in sync with NRL. Would it be worth move the shared ones into a separate repo that both projects pull from?

sources/knowledge_layer/src/nemo_retriever/_transport.py
```diff
@@ -0,0 +1,224 @@
...
+        limits = httpx.Limits(max_connections=100, max_keepalive_connections=20)
+        verify = _tls_verify(verify_ssl, ca_bundle)
+        self._client = client or httpx.Client(headers=headers, timeout=timeout, limits=limits, verify=verify)
+        if client is not None:
+            self._client.headers.update(headers)
+
+    def _url(self, path: str) -> str:
+        return f"{self._base_url}/{path.lstrip('/')}"
+
+    def _redact(self, value: str) -> str:
```
peterychang, line 113: If we make the api token a `pydantic.SecretStr`, we'll get the redaction for free

sources/knowledge_layer/src/nemo_retriever/adapter.py
peterychang: Would the solution here be to return the job_id immediately while the rest of the code executes in the background? What should happen in the case that a request using the job_id is made while the uploads haven't been completed or accepted yet on the backend?

sources/knowledge_layer/src/register.py
```diff
@@ -281,6 +281,52 @@ class KnowledgeRetrievalConfig(FunctionBaseConfig, name="knowledge_retrieval"):
         gt=0,
         description="Embedding dimensions; defaults to AIQ_EMBED_DIM and must match existing indexes",
     )
+    # NeMo Retriever service options
```
peterychang: just a comment: Its really nasty that the knowledge layer config needs to track every possible option for every subsystem. We really should pull this apart so each subsystem can track its own config options

src/aiq_agent/knowledge/schema.py
```diff
@@ -65,6 +65,14 @@ class Chunk(BaseModel):
         description="The main text. If this is a visual/table, this MUST be the summary or caption.",
     )
     score: float = Field(0.0, ge=0.0, le=1.0, description="Similarity score (0.0 to 1.0).")
+    distance: float | None = Field(
```
peterychang, line 68: does distance have a range? can it be negative?

frontends/aiq_api/src/aiq_api/routes/collections.py
```diff
@@ -135,3 +159,57 @@ async def health_check(
         except Exception as e:
             logger.error(f"Health check failed: {e}")
             raise HTTPException(status_code=503, detail=str(e))
+
+    @router.post(
```
peterychang: Is there a reason we're exposing the search to the api? Is AIQ expected to act as a data source for other systems in this case?

sources/knowledge_layer/src/nemo_retriever/_local_client.py
```diff
@@ -0,0 +1,1083 @@
...
+    data_dir: Path
+    scope: str
+    profile: str
+    page_elements_invoke_url: str | None
+    ocr_invoke_url: str | None
+    table_structure_invoke_url: str | None
+    embed_invoke_url: str | None
+    embed_model_name: str | None
+    embed_model_provider_prefix: str | None
+    inference_api_key: str | None = field(repr=False)
```
peterychang: nit: SecretStr?

sources/knowledge_layer/src/register.py
KyleZheng1284: yes I definitely agree this is getting pretty messy quickly

sources/knowledge_layer/src/nemo_retriever/_local_client.py
```diff
@@ -0,0 +1,1083 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
peterychang, line 1: It looks like most of this file is implementing custom workflows over the vdb directly. Does the NRL repo not implement this functionality itself?

peterychang:
A single unsupported filetype will fail the entire upload batch
image

peterychang:
Locking error
Steps:
1. Upload file
2. Run deep research

image

Edit:
The error appears to be caused by 2 different processes calling to get a local runtime with the same directory. Since the key-runtime mapping isn't shared across processes, we get this error

```
Process: MainProcess, PID: 19225, TID: 132365202329920
Key: /home/pc/git/aiq/.aiq-data/nemo_retriever, Entry: None
call_chain:
- _get_retriever (/home/pc/git/aiq/sources/knowledge_layer/src/register.py:611)
  - get_retriever (/home/pc/git/aiq/src/aiq_agent/knowledge/factory.py:178)
    - __init__ (/home/pc/git/aiq/sources/knowledge_layer/src/nemo_retriever/local_adapter.py:77)
      - _initialize_local_runtime (/home/pc/git/aiq/sources/knowledge_layer/src/nemo_retriever/local_adapter.py:40)
        - acquire_local_runtime (/home/pc/git/aiq/sources/knowledge_layer/src/nemo_retriever/_local_client.py:1067)
...
Process: Dask Worker process (from Nanny), PID: 19276, TID: 138847791591744
Key: /home/pc/git/aiq/.aiq-data/nemo_retriever, Entry: None
call_chain:
- _get_retriever (/home/pc/git/aiq/sources/knowledge_layer/src/register.py:611)
  - get_retriever (/home/pc/git/aiq/src/aiq_agent/knowledge/factory.py:178)
    - __init__ (/home/pc/git/aiq/sources/knowledge_layer/src/nemo_retriever/local_adapter.py:77)
      - _initialize_local_runtime (/home/pc/git/aiq/sources/knowledge_layer/src/nemo_retriever/local_adapter.py:40)
        - acquire_local_runtime (/home/pc/git/aiq/sources/knowledge_layer/src/nemo_retriever/_local_client.py:1067)
```

KyleZheng1284:
> Member

> A single unsupported filetype will fail the entire upload batch image


did you run this on `auto` mode for NRL pptx should be accepted as type, I think this might be more of a aiq issue? Particularly in which the default API allowlist is only `.pdf,.docx,.txt,.md `so the request is rejected before reaching NRL.

tests/knowledge_layer_tests/test_nemo_retriever_live.py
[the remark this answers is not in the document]
KyleZheng1284: Fixed: service boolean settings now accept only recognized true/false values and reject invalid input before TLS is configured.

sources/knowledge_layer/src/nemo_retriever/adapter.py
KyleZheng1284: Fixed: submit_job() now returns after registering the job, reports deterministic UPLOADING entries during background transfer, and replaces them with stable NRL document IDs as each upload is accepted

sources/knowledge_layer/src/nemo_retriever/_transport.py
KyleZheng1284: This advertises knowledge_layer as an explicitly selectable data source, but an async submission selecting that ID returns 422

configs/config_web_nemo_retriever.yml
KyleZheng1284: Fixed: knowledge_layer now resolves to knowledge_search during source validation, with a filtered route regression test covering the former 422 response

sources/knowledge_layer/src/nemo_retriever/_transport.py
KyleZheng1284: the token remains SecretStr through public and backend configuration and is unwrapped only when the HTTP transport is constructed, whose errors explicitly redact it.

sources/knowledge_layer/src/register.py
KyleZheng1284: Fixed for the new NRL paths: service and local options now live under backend-owned backend_config

sources/knowledge_layer/src/nemo_retriever/_local_client.py
KyleZheng1284: inference_api_key remains a SecretStr in local settings and runtime state and is unwrapped only for the immediate NRL inference invocation

sources/knowledge_layer/src/nemo_retriever/_local_client.py
KyleZheng1284: NRL does own the underlying storage and retrieval behavior. The local adapter instantiates NRL’s LanceDB with the same configuration as its VectorDB service and delegates to NRL’s collection APIs, schemas, CollectionWriteContext, IngestVdbOperator, and RetrieveVdbOperator. AI-Q only adds the staging, job-status, bounded-worker, locking, and lifecycle behavior required by the Knowledge API. 

It does not query LanceDB tables directly, reproduce NRL schemas, or select physical tables. We use the lower-level collection operators because the pinned NRL revision does not yet expose a single embedded collection facade, and its high-level Retriever does not currently consume the collection tuple result correctly

frontends/aiq_api/src/aiq_api/routes/collections.py
KyleZheng1284: This is intentionally a deployment-scoped endpoint for headless ingestion and retrieval validation without an LLM; chat calls the registered retriever directly, and the docs clarify that it is not a per-user authorization boundary.

sources/knowledge_layer/src/nemo_retriever/_models.py
KyleZheng1284: These models intentionally validate a narrow, forward-compatible REST wire contract rather than copying NRL storage schemas; I suggest keeping the pinned compatibility tests here and revisiting a shared or generated client when NRL publishes one.

sources/knowledge_layer/src/nemo_retriever/_transport.py
KyleZheng1284: only job creation and upload treat 404/410 as compatibility failures; job-status polling now preserves normal missing or expired resource errors.

src/aiq_agent/knowledge/schema.py
KyleZheng1284: moved it to be non optional

KyleZheng1284: Correction: both pinned NRL paths use LanceDB’s default L2 metric, so `distance` remains optional for backends that omit it but, when present, must be finite and non-negative, has no fixed upper bound, and uses lower-is-closer semantics.

tests/knowledge_layer_tests/test_nemo_retriever_adapter.py
```diff
@@ -0,0 +1,792 @@
...
+        client=httpx.Client(transport=mock_transport),
+    )
+    return {
+        "base_url": "https://nrl.example.test",
+        "scope": "workspace-123",
+        "api_token": SecretStr(token),
+        "max_retries": retries,
+        "max_concurrency": 2,
+        "_transport": transport,
+    }
```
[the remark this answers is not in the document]
KyleZheng1284: No change is needed: these are caller-owned `MockTransport` clients with no socket pool, and the focused suite passes with `ResourceWarning` promoted to an error.

sources/knowledge_layer/src/nemo_retriever/adapter.py
```diff
@@ -0,0 +1,694 @@
...
+        self._settings = _settings(self.config)
+        self._transport = _transport_for(self.config, self._settings)
+
+    @property
+    def backend_name(self) -> str:
+        return _BACKEND_NAME
+
+    def close(self) -> None:
+        """Release the transport created for this retriever instance."""
+        self._transport.close()
```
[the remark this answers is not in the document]
KyleZheng1284: Fixed without adding reference counting: REST retrievers no longer expose `close()`, service ingestor shutdown drains only its owned executor, and no non-owning adapter or direct-query request can close the process-shared transport.

src/aiq_agent/knowledge/factory.py
```diff
@@ -274,6 +274,19 @@ def clear_active_ingestor() -> None:
...
+    """Remove one expected cached ingestor during a backend lifecycle finalizer.
+
+    The identity check prevents an older finalizer from evicting a replacement
+    instance created for the same backend.
+    """
+    if _INGESTOR_INSTANCES.get(backend) is not ingestor:
+        return False
+    del _INGESTOR_INSTANCES[backend]
+    logger.info("Released singleton ingestor instance for backend: %s", backend)
+    return True
```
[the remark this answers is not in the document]
KyleZheng1284: No code change is needed: `get_ingestor()` never replaces a present cache entry and the supported lifecycle has one synchronous initializer and finalizer, while a generic lock would broaden every backend without solving multi-consumer reference ownership.

src/aiq_agent/fastapi_extensions/routes/documents.py
```diff
@@ -52,8 +54,9 @@ def add_document_routes(router: APIRouter):
         responses={
             400: {"description": "No files provided"},
             404: {"description": "Collection not found"},
-            413: {"description": "Upload size or file-count limit exceeded"},
+            413: {"description": "Upload size, file-count, or ingestion-capacity limit exceeded"},
             415: {"description": "Unsupported, malformed, or mismatched file content"},
+            503: {"description": "Document ingestion is temporarily at capacity"},
```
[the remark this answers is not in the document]
KyleZheng1284: Fixed in c506f815: the alternate document route now offloads every synchronous ingestor operation and reuses the existing cancellation-safe submission helper; the non-blocking contract test now covers both route implementations. Validation: 43 focused tests, 80 FastAPI-extension tests, and 2,344 full-suite tests passed.

tests/aiq_agent/fastapi_extensions/test_upload_security.py
```diff
@@ -685,3 +820,86 @@ async def save_upload(*_args, on_temp_path_created, **_kwargs):
...
+            ingestor=ingestor,
+        )
+    )
+    try:
+        assert await asyncio.to_thread(ingestor.submission_started.wait, 1)
+        await asyncio.sleep(0)
+        assert not safety_fired.is_set(), "submit_job blocked the event loop until the deadlock fuse fired"
+        assert not ingestor.release_submission.is_set(), "submit_job blocked the event loop"
+        ingestor.release_submission.set()
+        response = await asyncio.wait_for(request_task, 1)
```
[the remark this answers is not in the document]
KyleZheng1284: Fixed in c506f815: the one-second synchronization waits are now five-second CI safety bounds while the existing ten-second deadlock fuse remains unchanged. The focused, module, and full Python suites all pass.

peterychang:
> > Member
> 
> > A single unsupported filetype will fail the entire upload batch image
> 
> did you run this on `auto` mode for NRL pptx should be accepted as type, I think this might be more of a aiq issue? Particularly in which the default API allowlist is only `.pdf,.docx,.txt,.md `so the request is rejected before reaching NRL.

I'm not sure, I ran the default config for the local nrl instance. I'll check the configs and try running it again

Edit:
Yes, my env config was wrong. Works with the correct config

frontends/aiq_api/src/aiq_api/routes/collections.py
```diff
@@ -135,3 +159,57 @@ async def health_check(
...
+        except Exception as exc:
+            logger.warning(
+                "Knowledge collection lookup failed (backend=%s, error_type=%s)",
+                ingestor.backend_name,
+                type(exc).__name__,
+            )
+            raise HTTPException(status_code=502, detail="Knowledge backend query failed") from exc
+
+        if collection is None:
+            raise HTTPException(status_code=404, detail=f"Collection '{request.collection_name}' not found")
```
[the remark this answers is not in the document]
KyleZheng1284: This is a valid concern. The current Knowledge API is deployment-scoped, and AI-Q does not yet have a collection-ownership model. Adding a query-only check here would be incomplete, so tenancy should be designed consistently across collection creation, listing, upload, query, and deletion. I’m keeping this thread open until that scope is agreed.

sources/knowledge_layer/src/nemo_retriever/_local_client.py
```diff
@@ -0,0 +1,1083 @@
...
+        return CollectionInfo(
+            name=name,
+            description=item.get("description"),
+            file_count=len(files),
+            chunk_count=sum(file.chunk_count for file in files),
+            created_at=_timestamp(item.get("created_at")),
+            updated_at=_timestamp(item.get("updated_at")),
+            backend=_BACKEND_NAME,
+            metadata=scrub_metadata(item.get("metadata") or {}),
+        )
```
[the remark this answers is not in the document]
KyleZheng1284: Agreed that scanning every document to calculate collection counts is a performance limitation. The pinned NRL collection API does not currently expose aggregate document and chunk counts, and AI-Q should not query LanceDB directly or invent an NRL response contract. We’ll retain the correct public-API pagination for now and adopt aggregate counts when NRL exposes them.

peterychang (approved):

tanleach:
/ok to test [d04ddbf](https://github.com/NVIDIA-AI-Blueprints/aiq/pull/371/commits/d04ddbf1d66a1eef8fa0e330de2e957c1ab27f13)

frontends/aiq_api/src/aiq_api/routes/collections.py
KyleZheng1284: Removed in `1e715815`. Chat already invokes the registered retriever through `knowledge_search`, so this PR no longer adds a public retrieval API; backend-level integration tests retain no-LLM retrieval coverage.

PR #389, opened 2026-07-27
refactor(deep-research): single-source the chart contract via an on-demand visualization skill

Manushpm8:
#### Overview

The inline chart-spec contract lived in the always-on `writer.j2` prompt (and was duplicated in the sandbox `## Figures` block). That is prompt weight every deep-research writer paid on every turn, whether or not a chart was warranted, which works against keeping the writer prompt lean.

This change single-sources that contract into the on-demand `visualization` chart skill and loads it only when a chart is actually needed:

- Moves `chart-generation` into its own sandbox-independent `visualization` skill collection (out of the sandbox-only `research` collection) and makes its `SKILL.md` two-mode: the existing sandbox path (matplotlib PNG via `execute`, `artifact://` reference) plus a new inline path holding the full `chart` / `chart-carousel` / KPI-only spec contract, field reference, examples, and the portable markdown-table fallback.
- Shrinks `writer.j2` from an embedded chart contract to one skills-gated pointer: sandbox writers get the per-job run command, non-sandbox writers are pointed at the inline spec in the skill, and a no-skills `{% else %}` branch degrades gracefully to a markdown table. No `chart` JSON syntax remains in the prompt.
- Wires the `visualization` skill into every deep-research config so inline charts keep working everywhere. This is behavior-preserving: no config loses inline charts. `visualization` is not sandbox-gated, so it is never added to `require_sandbox`.

The skill machinery is independent of the sandbox: `skills_enabled` is driven by the `deep_research_skills` config, and the `/skills/` route is a filesystem-backed read, so the writer can load the chart skill with or without a sandbox.

Signed-off-by: Manush Maheshwari <manushm@nvidia.com>

#### Validation

**Unit and prompt tests** (`tests/aiq_agent/agents/deep_researcher/` plus `test_result_chart_prompt.py`): 399 passed, 4 skipped. Coverage asserts the chart contract lives in the skill (both modes) and the shallow prompt, that `writer.j2` no longer embeds it (`"```chart" in rendered_prompt` is `False`), that the writer delegates to the skill, and that the skill examples validate against the UI zod schema.

**Config validation**: all 11 deep-research configs parse and load; every `deep_research_agent.skills` reference resolves to a defined `deep_research_skills` function (no dangling refs).

**Live end-to-end run** on the default model (`nvidia/nemotron-3-super-120b-a12b` via `integrate.api.nvidia.com`), non-sandbox writer with the `visualization` skill wired and zero chart syntax in the base prompt. The writer navigated the skill route and read the contract before emitting a chart:

```
ls /skills                                            -> ['/skills/visualization/']
ls /skills/visualization                              -> ['/skills/visualization/chart-generation/']
read_file /skills/visualization/chart-generation/SKILL.md   (SKILL.md content)
read_file .../SKILL.md offset=100 limit=200                 (Inline Mode contract)
write_file /shared/output.md  ->  ```chart {"type":"hbar", ...} ```
```

The emitted spec passed the real UI `ChartSpecSchema` / `parseChartSpec` and rendered through the actual `ChartBlock` -> `ResultChart` component (not a fallback code block). Because the chart syntax exists only in the skill file, this confirms the on-demand skill mechanism produces charts end to end on the default model.

- `src/aiq_agent/agents/deep_researcher/skills/visualization/chart-generation/SKILL.md` (the two-mode contract, the single source of truth).
- `src/aiq_agent/agents/deep_researcher/prompts/writer.j2` (the skills-gated pointer that replaced the embedded contract).
- One wired config, e.g. `configs/config_cli_default.yml` (the `deep_research_skills` block + `skills:` reference pattern, applied to all deep-research configs).

#### Related Issues

- Relates to #

## Summary by CodeRabbit

* **New Features**
  * Deep research writer now routes chart generation through the visualization capability via a shared skills mapping across supported configurations.
  * Writer conditionally renders “Figures and Charts” when available, with two delivery modes: sandbox PNG artifacts (when execution is available) or inline chart specifications.
* **Documentation**
  * Added/updated migration guidance noting that chart-generation moved from the research collection to visualization.
* **Tests**
  * Expanded and refactored coverage for both chart delivery modes, writer degradation when visualization isn’t enabled, and the sandbox chart artifact contract.


src/aiq_agent/agents/deep_researcher/prompts/writer.j2
```diff
@@ -75,15 +75,19 @@ When synthesizing:
 - Do not expose internal `evidence_judgment` scores or rationale in the final answer unless the user explicitly asks for methodology.
 - Err on the side of more useful information rather than less, while staying focused on the requested answer shape.
 
-{% if execution_enabled %}## Figures
+{% if skills_enabled %}## Figures and Charts
```
KyleZheng1284: Could we gate this on whether the writer has the `visualization` skill? 

It seems like the global `skills_enabled` may be true even when only the researcher has skills, leaving the writer unable to load `chart-generation`. Perhaps a writer-specific `chart_skill_enabled` value plus a regression test should cover this cleanly.

src/aiq_agent/agents/deep_researcher/skills/visualization/chart-generation/SKILL.md
```diff
@@ -2,18 +2,59 @@
 name: chart-generation
```
KyleZheng1284, line 2: Moving this skill from research to visualization changes existing valid configurations that assign only research to the writer. 

Is there enough documentation for this change? Since we are requiring the migration to [research, visualization]

src/aiq_agent/agents/deep_researcher/prompts/writer.j2
```diff
@@ -75,15 +75,19 @@ When synthesizing:
...
+{% if execution_enabled %}
+A sandbox is available, so the skill renders a PNG artifact. Write the plotting script to `{{ sandbox_workdir }}/make_chart.py` and run it exactly as `python3 {{ sandbox_workdir }}/make_chart.py {{ sandbox_artifact_dir }}` so the image plus its `manifest.json` land in the only directory the runtime harvests, then embed it once where it is discussed with `![<caption>](artifact://<filename>)`. Never put a literal `<sandbox_workdir>` or `<sandbox_artifact_dir>` placeholder into code or a command, and never paste plotting code, a sandbox path, or base64 image data into the report.
+{% else %}
+There is no sandbox, so the skill emits an inline `chart` (or `chart-carousel`) JSON spec that the web app renders, followed by a compact portable table of the same values.
+{% endif %}
+A figure is earned: chart only source-anchored, reasonably complete data. If a series is mostly undisclosed or mixes metric definitions, present the table (which shows the gaps) and state the limitation in one sentence instead of a misleading chart.
```
KyleZheng1284: Could we tighten the ownership boundary here? I feel like the writer prompt should expose only the writer’s actual visualization capability and runtime context, while `chart-generation` owns the full inline/artifact procedure.

 If sandbox → artifact is a strict contract, it should be enforced outside the prompt, and the `research` → `visualization` move should be documented as a configuration migration.

KyleZheng1284 (approved):
LGTM, I like the idea of routing it through a visualization skill then checking if sandbox is configured or not

KyleZheng1284:
**Clarify or enforce the sandbox manifest checkpoint contract**

@Manushpm8 
In live sandbox job `ff467a9b-fc8f-465d-ad11-478b51125564`, `revenue_bar.png` and `revenue.csv` were exported successfully. However, they appeared only at terminal harvest time and had scan-default metadata (`title: null`, `caption: null`, `inline: false`). This suggests that the manifest-driven `harvest_after_execute()` checkpoint did not capture valid manifest entries, and the terminal directory scan recovered the files instead.

This is not a request to expose `manifest.json` as a downloadable artifact, nor is Markdown embedding required. The concern is that the chart skill currently describes the manifest as required for reliable checkpoint harvesting.

Could we either:

1. Add an integration test proving that a sandbox chart execution writes a valid manifest and is captured by `harvest_after_execute()` before terminal scanning; or
2. Clarify in the skill that the manifest is preferred metadata/checkpointing and that terminal directory-scan recovery is an accepted successful path?

If the manifest remains required, the test should assert that the PNG is durable immediately after `execute` and that manifest metadata survives into its artifact record. The final scan should remain as the failure-recovery path.

cdgamarose-nv (approved):
LGTM!

AjayThorve:
can you pull in latest develop, CI should be fixed

Manushpm8:
@AjayThorve done, pulled in latest develop (`8ff182a`). It picks up the crypto-50 bump that clears the MCP audit gate, and the PR is mergeable now. It just needs a `/ok to test 8ff182a` to re-run the heavy CI. Thanks!

KyleZheng1284:
/ok to test https://github.com/NVIDIA-AI-Blueprints/aiq/pull/389

KyleZheng1284:
/ok to test https://github.com/NVIDIA-AI-Blueprints/aiq/commit/9731fae4c58792ca266a616d17a2e3750236b4df

KyleZheng1284:
/ merge

KyleZheng1284 (approved):

PR #439, opened 2026-08-12
fix: enforce shallow citation integrity

AjayThorve:
#### Overview

- preserve the existing `release/2.2` shallow-research prompt and add one targeted citation preflight reminder
- require every published researched answer to retain both a verified inline citation and a verified source section
- preserve deterministic single-source recovery; for ambiguous multi-source drafts, allow one bounded, tool-free citation repair and fail closed if integrity is still missing
- support both public-URL sources and uploaded-document citation keys such as `report.pdf, p.15`
- keep the behavior provider-agnostic so the default Lightning and frontier Luna profiles share the same hardened runtime path

Nemotron 3.5 Lightning can intermittently return a correct answer after successful research without inline citation markers or a source section. Prompting can reduce that failure rate, but it is not a product invariant. This change keeps the current research behavior intact and makes an uncited researched answer unpublishable.

Signed-off-by: Ajay Thorve <athorve@nvidia.com>

#### Validation

- [x] `uv run ruff check .`
- [x] `uv run ruff format --check .`
- [x] `uv run pytest -q tests/ sources/`: **2157 passed, 13 skipped**
- [x] prompt diff against `release/2.2` is one citation-preflight line; query rewriting, tool selection, source hierarchy, and answer style are unchanged
- [x] uploaded-document integration: an uncited multi-file `knowledge_search` summary invokes the bounded repair and publishes only with an inline marker plus verified filename/page citation key
- [x] citation-integrity parsing preserves valid answer lines that begin with an inline marker instead of treating them as reference definitions
- [x] Inference Hub shallow-workflow smoke on the final prompt/runtime: **5/5 Lightning** and **5/5 Luna** reports passed verified inline-plus-source checks
- [x] Lightning exercised the bounded repair path in **2/5** cases; both repaired reports passed the final integrity gate
- [x] Luna used the same shared prompt/runtime path and passed **5/5** without repair

The live smoke is a provider compatibility and citation-reliability check, not a replacement for the existing FreshQA quality benchmark.

1. `src/aiq_agent/agents/shallow_researcher/agent.py`: the final publication invariant, deterministic single-source fallback, and bounded multi-source repair
2. `tests/aiq_agent/agents/shallow_researcher/test_agent.py`: fail-closed, timeout, uploaded-document, source-heading, prefix-citation, and repair-path regression coverage
3. `tests/aiq_agent/test_default_model_profiles.py`: default Lightning and frontier Luna sharing the same hardened path
4. `src/aiq_agent/agents/shallow_researcher/prompts/researcher.j2`: the single targeted prompt addition

#### Related Issues

- No public issue; fixes the shallow citation regression reported during 2.2 RC validation.

## Summary by CodeRabbit

* **Improvements**
  * Research reports now provide more reliable inline citations and complete reference sections.
  * Missing or invalid citations are automatically repaired when possible using identified sources.
  * Citation repair is bounded to prevent delays, and reports with unresolved issues are prevented from being published with unsupported claims.
  * Citation handling now supports a broader range of source, heading, and reference formats.
  * Research responses continue to follow guidance for source prioritization, premise verification, concise answers, unverifiable claims, and uploaded-document research.

### DIFF
--- a/src/aiq_agent/agents/shallow_researcher/agent.py
+++ b/src/aiq_agent/agents/shallow_researcher/agent.py
@@ -113,10 +113,18 @@ def _has_citation_integrity(report_text: str, valid_citations: Sequence[dict[str
     if source_heading is None:
         return False
     # Source-definition lines contain [N] labels but are not inline citations.
-    # Remove both the heading and definitions before accepting markers anywhere
-    # else, including a short answer that a provider placed after the section.
-    prose = report_text[: source_heading.start()] + report_text[source_heading.end() :]
-    prose = _REFERENCE_ENTRY_LINE_RE.sub("", prose)
+    # Remove only the contiguous definition block after the heading; matching
+    # globally would also erase valid prose that begins with an inline marker.
+    tail_lines = report_text[source_heading.end() :].splitlines(keepends=True)
+    first_definition = 0
+    while first_definition < len(tail_lines) and not tail_lines[first_definition].strip():
+        first_definition += 1
+    after_definitions = first_definition
+    while after_definitions < len(tail_lines) and _REFERENCE_ENTRY_LINE_RE.fullmatch(
+        tail_lines[after_definitions].rstrip("\r\n")
+    ):
+        after_definitions += 1
+    prose = report_text[: source_heading.start()] + "".join(tail_lines[after_definitions:])
     return any(int(number) in valid_numbers for number in _INLINE_CITATION_RE.findall(prose))
 
 
--- a/src/aiq_agent/agents/shallow_researcher/agent.py
+++ b/src/aiq_agent/agents/shallow_researcher/agent.py
@@ -62,15 +62,75 @@ AGENT_DIR = Path(__file__).parent
 
 _SOURCE_SECTION_HEADING_RE = re.compile(
     r"^[^\S\n]*(?:"
-    r"#{1,6}[^\S\n]+(?:Sources|References)\b[^\n]*"
-    r"|\*\*(?:Sources|References):?\*\*[^\n]*"
-    r"|(?:Sources|References):[^\n]*"
-    r"|(?:Sources|References)"
+    r"#{1,6}[^\S\n]+(?:Sources|References):?"
+    r"|\*\*(?:Sources|References):?\*\*:?"
+    r"|(?:Sources|References):?"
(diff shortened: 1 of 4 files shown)


tanleach:
after going back and forth with my review agent I found the following:

```
Single-source fallback can silently erase legitimate answer content

 Locations

 - src/aiq_agent/agents/shallow_researcher/agent.py:63-70
 - src/aiq_agent/agents/shallow_researcher/agent.py:82-87
 - src/aiq_agent/agents/shallow_researcher/agent.py:525-527

 The new heading regex treats any Markdown heading beginning with Sources or References as a source section:

 ```regex
   #{1,6}\s+(?:Sources|References)\b[^\n]*
 ```

 _append_minimal_citation() then discards everything from the first match onward.

 I reproduced this input:

 ```markdown
   ## Sources of renewable energy
   Solar is renewable.
 ```

 The fallback produced:

 ```markdown
    [1]

   **References:**
   - [1] Article - https://example.com
 ```

 The complete answer was deleted, and the resulting bare citation can still satisfy the final structural check.

 Similar losses occur with:

 ```markdown
   Sources: market revenue
   The business grew.
 ```

 and with a real source block followed by trailing answer text.

 Required correction

 A source section should be recognized only when:

 - It uses a canonical exact heading, and
 - It is followed by an actual reference-definition block or is an empty trailing heading left by verification.

 When replacing a source section, remove only the heading and associated reference definitions. Preserve all surrounding
 answer text.

 Add regressions for:

 - ## Sources of renewable energy
 - ## References to previous work
 - Plain prose beginning with Sources:
 - Answer text following a real source block```

AjayThorve:
@tanleach Thanks — I reproduced this and the finding was valid. Fixed in `fd7739c1` and tightened further in `b568befc`:

- only exact canonical `Sources`/`References` headings followed by a real contiguous definition block (or an empty trailing heading) are treated as source sections
- replacement removes only that section and preserves surrounding answer text
- source-like prose and headings such as `## Sources of renewable energy`, `## References to previous work`, and `Sources: market revenue` are preserved
- trailing and marker-first answer text after a real reference block is preserved

The requested regressions are included. Validation is green: 224 citation-focused tests, the full suite at 2165 passed / 13 skipped, and all PR checks including CodeRabbit and DCO.

tanleach (approved):

PR #444, opened 2026-08-13
fix(shallow): improve citation reliability

AjayThorve:
#### Overview

Nemotron 3.5 Lightning usually omits citations in its initial shallow-research draft. The rc10 citation-integrity fix therefore invoked a second, one-shot LLM repair on nearly every request. On NVIDIA Build, that repair frequently exceeded its hard-coded 60-second timeout and converted otherwise successful research into a failed workflow.

This change removes the latency-sensitive citation-repair pass and makes citation-complete generation part of the primary shallow-research flow:

- provide a registry-backed, numbered source catalog immediately after tool evidence and require exact inline citations plus matching reference rows;
- allow the first post-evidence response to either synthesize from relevant evidence or make one final, rewritten-query search when the result is empty, failed, irrelevant, conflicting, or does not support the question's premise;
- cap shallow research at two tool calls regardless of the broader configured iteration budget;
- track evidence per invocation, clear additive turn evidence at the run boundary, and prevent stale session sources from suppressing a retry or validating the current answer;
- handle a failed first tool call as the first bounded attempt so one retry remains available; and
- fail closed when the final draft lacks verified citations, without auto-attaching a single unrelated source.

The upfront and post-search prompts also require the model to verify assumed events, relationships, counts, titles, and entities, and to directly correct unsupported premises rather than substitute a related fact.

Citation verification remains authoritative. This PR removes the repair LLM call; it does not weaken the publication invariant or publish an uncited fallback.

Signed-off-by: Ajay Thorve <athorve@nvidia.com>

#### Validation

Focused shallow-research tests:

```text
$ uv run pytest -q tests/aiq_agent/agents/shallow_researcher
73 passed in 2.13s
```

Broader citation and source-registry tests:

```text
$ uv run pytest -q tests/aiq_agent/agents/shallow_researcher tests/aiq_agent/common/test_citation_verification.py tests/aiq_agent/common/test_data_source_registry.py
260 passed in 2.06s
```

Targeted Ruff validation:

```text
$ uv run ruff check src/aiq_agent/agents/shallow_researcher tests/aiq_agent/agents/shallow_researcher
All checks passed!

$ uv run ruff format --check src/aiq_agent/agents/shallow_researcher tests/aiq_agent/agents/shallow_researcher
10 files already formatted
```

Commit hooks passed for the final two commits, including Ruff, formatting, merge-conflict, large-file, whitespace, and secret checks.

NVIDIA Build forced-shallow benchmark using the same 15 prompts, submitted serially through the async-job API with `agent_type=shallow_researcher`:

- rc10 baseline: 6/15 successful and cited (40%), average 53.21s, p90 151.56s;
- final candidate: 14/15 successful and cited (93.3%), average 25.31s, p50 25.51s, p90 45.81s, max 49.85s; and
- the only failure was an immediate fail-closed `CitationIntegrityError`, not a repair timeout.

This benchmark validates the release-blocking publication reliability and latency path. It is not a claim of semantic-answer parity: source relevance remains model-dependent, with the bounded retry and premise-verification prompt reducing but not eliminating irrelevant evidence selection.

The local Inference Hub endpoint override used during comparison testing is intentionally excluded from this PR.

- [ ] I updated documentation for user-facing or contributor-facing changes. Not applicable: this changes internal shallow-agent generation and verification behavior without changing a public API, configuration contract, or contributor workflow.

Start with `src/aiq_agent/agents/shallow_researcher/agent.py`, specifically the post-tool relevance/synthesis decision, the two-call budget, and current-turn registry reconstruction in `run()`. Then review the behavioral coverage in `tests/aiq_agent/agents/shallow_researcher/test_agent.py` for irrelevant results, tool failures, stale session/input evidence, malformed citations, and fail-closed behavior.

The central design decision is to make the primary generation path citation-complete while retaining exactly one bounded research retry. There is no citation-repair LLM call and no uncited fallback.

#### Related Issues

- Follow-up to #439


AjayThorve:
Cross-endpoint regression validation with the exact PR image (`4f6da553`) is complete.

| Endpoint | Earlier candidate | PR image | Latency |
| --- | ---: | ---: | --- |
| Inference Hub | 13/15 (86.7%) | 15/15 (100%) | avg 16.13s, p50 14.60s, p90 19.68s |
| Local pinned NVFP4 vLLM | 13/15 (86.7%) | 27/30 across two runs (90%) | avg 10.94s, p50 9.12s, p90 16.43s |

All requests were forced through `shallow_researcher`, serial, using the same 15 prompts and endpoint-specific configs. Inference Hub performed 15 searches for 15 requests; local performed 30 searches for 30 requests. There were no repair calls and no redundant tool retries.

The three local failures were fail-closed citation-integrity failures and were prompt-independent across repeats: prompts that failed in one run passed in another. This matches the stochastic format-adherence behavior seen in the earlier local baseline rather than a deterministic regression.

Citation success therefore shows no regression and remains above the 80% acceptance gate on both endpoints. Latency is somewhat higher than the earlier sample (IH p50 10.15s -> 14.60s; local p50 7.85s -> 9.12s), so this evidence should not be used to claim latency neutrality.

src/aiq_agent/agents/shallow_researcher/agent.py
```diff
@@ -183,6 +182,18 @@ def _append_minimal_citation(report_text: str, source: SourceEntry) -> str:
...
+def _format_synthesis_source_catalog(sources: Sequence[SourceEntry]) -> str:
+    """Render registry-backed reference rows for the first synthesis draft."""
+    rows: list[str] = []
+    for number, source in enumerate(sources, 1):
+        if source.url:
+            title = " ".join((source.title or source.url).split())
+            rows.append(f"- [{number}] {title} - {source.url}")
+        elif source.citation_key:
+            rows.append(f"- [{number}] {source.citation_key}")
+    return "\n".join(rows)
```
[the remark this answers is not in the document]
AjayThorve: Fixed in f353a0cd. Catalog display fields are canonicalized onto one printable line, URL identities containing whitespace/control characters are rejected rather than transformed, and the catalog is explicitly fenced and labeled as untrusted source data. The regression coverage includes newline/control-bearing citation data and an unsafe URL. Focused result: 83 passed; Ruff and changed-file pre-commit hooks pass.

tests/aiq_agent/agents/shallow_researcher/test_agent.py
```diff
@@ -307,27 +315,8 @@ def test_default_prompt_has_structural_citation_contract(self, mock_llm_provider
...
-        )
-
-        async with asyncio.timeout(0.1):
-            with pytest.raises(CitationIntegrityError, match="citation_integrity_lost"):
-                await agent._repair_missing_citations(
-                    [HumanMessage(content="Draft")],
-                    [SourceEntry(url="https://example.com/source")],
-                )
+        assert "Your draft is discarded" in agent.system_prompt
+        assert "Do not use alternative markers" in agent.system_prompt
```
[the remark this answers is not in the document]
AjayThorve: Fixed in f353a0cd. Removed the prompt-prose assertions and replaced them with structural checks for message topology, unbound synthesis, catalog boundaries/content, and the cited final-answer behavior. Focused result: 83 passed; Ruff and changed-file pre-commit hooks pass.

src/aiq_agent/agents/shallow_researcher/agent.py
```diff
@@ -183,6 +183,34 @@ def _append_minimal_citation(report_text: str, source: SourceEntry) -> str:
...
+            # instead of transforming the source identity.
+            if any(char.isspace() or unicodedata.category(char).startswith("C") for char in source.url):
+                continue
+            title = canonicalize_text(source.title or source.url)
+            if title:
+                rows.append(f"- [{number}] {title} - {source.url}")
+        elif source.citation_key:
+            citation_key = canonicalize_text(source.citation_key)
+            if citation_key:
+                rows.append(f"- [{number}] {citation_key}")
```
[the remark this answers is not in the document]
AjayThorve: Fixed in 8c035c57 with a narrower identity-preserving approach. Safe citation keys are copied byte-for-byte so the prompt and SourceRegistry compare the same value; keys containing control characters, non-space whitespace, or surrounding whitespace are excluded from the catalog. This avoids changing the shared registry matching contract. The focused test covers rejection of the newline/control-bearing key and preservation of repeated internal spaces. Result: 83 passed; Ruff and changed-file pre-commit hooks pass.

tanleach:
Stale session sources can produce false citations
 
 agent.py:385-445 builds the synthesis catalog from every source accumulated in the persistent session registry. That
 registry is intentionally reused across turns by chat_researcher/register.py:691-695.

 Independent reproduction:

 - Prior turn registered an old weather URL.
 - Current CUDA search returned no results.
 - The old source made source_catalog non-empty, so the current empty-result retry was skipped.
 - The returned answer was:

 ```text
   CUDA is a parallel-computing platform [1].

   ## Sources
   - [1] Old weather report - https://weather.example/old
 ```

 This is an ordinary multi-turn path, not merely a hostile-model scenario. It silently turns a failed current search into a
 confidently false citation.

src/aiq_agent/agents/shallow_researcher/agent.py
```diff
@@ -403,17 +357,91 @@ async def agent_node(state: ShallowResearchAgentState) -> dict[str, Any]:
...
+            # registry identity proves provenance, not that the source supports
+            # the answer.
+            research_limit = min(self.max_tool_iterations, 2)
+            has_tool_result = bool(processed_history and isinstance(processed_history[-1], ToolMessage))
+            research_exhausted = iterations >= research_limit
+            relevance_retry_available = has_tool_result and not research_exhausted
+            source_catalog = ""
+            if has_tool_result:
+                active_registry = get_session_registry() or self.source_registry
+                source_catalog = _format_synthesis_source_catalog(active_registry.all_sources())
```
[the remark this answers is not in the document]
AjayThorve: Fixed in 41e9c735. The graph now accumulates sources captured by the current shallow turn; both the post-tool catalog and citation verification use that turn-scoped allowlist whenever research ran. Prior session sources remain usable only for a no-tool conversational follow-up. Added a regression with a stale prior-turn URL plus two empty current searches that proves the retry still occurs and the stale source cannot validate the final draft.

AjayThorve:
> Stale session sources can produce false citations
> 
> agent.py:385-445 builds the synthesis catalog from every source accumulated in the persistent session registry. That registry is intentionally reused across turns by chat_researcher/register.py:691-695.
> 
> Independent reproduction:
> 
> * Prior turn registered an old weather URL.
> * Current CUDA search returned no results.
> * The old source made source_catalog non-empty, so the current empty-result retry was skipped.
> * The returned answer was:
> 
> ```
>   CUDA is a parallel-computing platform [1].
> 
>   ## Sources
>   - [1] Old weather report - https://weather.example/old
> ```
> 
> This is an ordinary multi-turn path, not merely a hostile-model scenario. It silently turns a failed current search into a confidently false citation.

great catch, fixed this now

src/aiq_agent/agents/shallow_researcher/agent.py
```diff
@@ -561,6 +605,14 @@ async def run(self, state: ShallowResearchAgentState) -> ShallowResearchAgentSta
...
         # Post-process: verify citations against source registry
         validated_result = dict(result)
+        if validated_result.get("tool_iterations", 0) > 0:
+            # Session registries intentionally span conversation turns. Once
+            # this turn performed research, verify only sources returned by
+            # this turn so stale evidence cannot suppress retries or validate
+            # an unrelated follow-up answer.
+            registry = SourceRegistry()
+            for source in validated_result.get("turn_sources", []):
+                registry.add(source)
```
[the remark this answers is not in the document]
AjayThorve: Fixed in 50ef8367. `run()` now clears the additive `turn_sources` field before graph invocation, so a reused input state cannot promote prior-run evidence into the current turn. The regression pre-populates both the session registry and input `turn_sources`, then verifies two empty searches still retry once and fail with `EmptySourceRegistryError`. Focused shallow tests: 73 passed; broader citation/data-source tests: 260 passed.

src/aiq_agent/agents/shallow_researcher/models/state.py
```diff
@@ -46,3 +50,4 @@ class ShallowResearchAgentState(BaseModel):
     available_documents: list[AvailableDocument] | None = None
     collection_name: str | None = None
     tool_iterations: int = 0
+    turn_sources: Annotated[list[SourceEntry], add] = Field(default_factory=list)
```
KyleZheng1284: **[P1 suggestion] Keep rejected search evidence out of the final answer**

The run-boundary reset fixes sources leaking across separate runs, but this additive reducer also keeps sources from both search attempts inside one run. If the first result is citable but rejected as irrelevant and the retry returns no sources, the rejected first result still appears in the synthesis catalog and citation allowlist. I reproduced this with a CUDA answer citing a San Francisco weather page after an empty retry.

Could we make source ownership attempt-specific—for example, replace `turn_sources` on each ToolNode attempt, or promote only the evidence accepted for synthesis? A regression should cover a citable-but-irrelevant first result followed by an empty retry and verify that the run fails with `EmptySourceRegistryError` without publishing a final report.

src/aiq_agent/agents/shallow_researcher/agent.py
```diff
@@ -466,7 +510,10 @@ async def agent_node(state: ShallowResearchAgentState) -> dict[str, Any]:
 
         builder.set_entry_point("agent")
 
-        tool_node = ToolNode(self.tools)
+        # Convert tool exceptions into ToolMessages so a failed first search
+        # consumes its attempt but can still use the one bounded relevance
+        # retry. If the retry also fails, the empty registry fails closed.
+        tool_node = ToolNode(self.tools, handle_tool_errors=True)
```
KyleZheng1284: **[P1 security suggestion] Redact tool exceptions before sending them back to the model**

With the locked LangGraph version, `handle_tool_errors=True` turns `repr(exception)` into `ToolMessage` content. This graph then includes that message in the next external LLM request. A custom, MCP, per-user, or validation tool can include credentials or internal endpoints in its exception text; a focused reproduction confirmed that dummy secret and endpoint markers reached the next model call.

Could we use a constant safe error message or a redacting handler while preserving the tool message's error status? Please add a test where a tool raises an exception containing dummy sensitive markers, verify the rewritten retry still works, and assert that neither marker appears in the next LLM request.

src/aiq_agent/agents/shallow_researcher/agent.py
```diff
@@ -561,6 +615,14 @@ async def run(self, state: ShallowResearchAgentState) -> ShallowResearchAgentSta
 
         # Post-process: verify citations against source registry
         validated_result = dict(result)
+        if validated_result.get("tool_iterations", 0) > 0:
+            # Session registries intentionally span conversation turns. Once
+            # this turn performed research, verify only sources returned by
+            # this turn so stale evidence cannot suppress retries or validate
+            # an unrelated follow-up answer.
+            registry = SourceRegistry()
```
KyleZheng1284, line 623: **[P1 contract suggestion] Do not publish a shallow-research draft when this turn has no evidence**

This publication behavior predates the PR, so I am not treating it as a newly introduced regression. It does, however, conflict with the PR's stated guarantee that it will not publish an uncited fallback. When the current-turn registry is empty, the code below stores the last model response in `generated_answer`; `public_response`, chat/eval callers, and the async job runner can then expose it as a report even though the workflow failed.

Could the shallow empty-evidence path omit `generated_answer` and return only the typed remediation response? This should be scoped to shallow research so existing deep-research behavior is not changed unintentionally. Tests should cover source-disabled and failed/empty-search runs, including the public response and absence of an async final-report artifact.

src/aiq_agent/agents/shallow_researcher/agent.py
```diff
@@ -422,7 +454,12 @@ async def agent_node(state: ShallowResearchAgentState) -> dict[str, Any]:
                     return {"messages": [response], "tool_iterations": iterations}
 
                 llm = self._get_llm()
-                llm_with_tools = llm.bind_tools(self.tools) if self.tools else llm
+                llm_with_tools = llm
+                if self.tools:
+                    if has_tool_result:
+                        llm_with_tools = llm.bind_tools(self.tools, parallel_tool_calls=False)
+                    else:
+                        llm_with_tools = llm.bind_tools(self.tools)
```
KyleZheng1284: **[P2 suggestion] Enforce the tool-call limit before any tools execute**

The initial binding still allows the provider's default parallel-call behavior, and this path does not validate the returned calls before ToolNode runs them. A model response containing three tool calls therefore executes all three before the iteration counter can notice. I reproduced three tool side effects from one initial response despite the advertised two-call total limit.

Shipped profiles currently set `parallel_tool_calls: false`, but custom profiles and direct callers do not necessarily do so. Could we also bind the initial call with parallel calls disabled and defensively require exactly one allowed source-tool call before returning the response to ToolNode? A regression should return multiple calls and assert that no tool side effect occurs.

src/aiq_agent/agents/shallow_researcher/agent.py
```diff
@@ -403,17 +357,95 @@ async def agent_node(state: ShallowResearchAgentState) -> dict[str, Any]:
 
             processed_history = list(messages)
 
+            # Shallow research gets at most two tool calls: the initial search
+            # and one rewritten-query retry when the first result is unusable.
+            # Keep the model's relevance decision on the first post-tool turn;
+            # registry identity proves provenance, not that the source supports
+            # the answer.
+            research_limit = min(self.max_tool_iterations, 2)
```
KyleZheng1284: **[P2 suggestion] Align the two-call runtime cap with the configuration and documentation contract**

This new clamp makes every `max_tool_iterations` value above two ineffective, while the configuration schema still defaults to five, shipped profiles and examples use five, troubleshooting recommends increasing it, and the prompt says "two calls per tool" rather than two calls total. Existing callers therefore receive behavior different from what the public configuration describes.

Could we choose one supported contract and align all surfaces? If the PR intends an absolute two-call total, the smallest compatibility-friendly option is to document that contract, change the default/examples/prompt to two total, and continue accepting legacy values above two so existing YAML still loads. Please add parameterized tests for values 1, 2, and 5 that demonstrate the chosen semantics.

src/aiq_agent/agents/shallow_researcher/agent.py
```diff
@@ -246,7 +256,6 @@ def __init__(
         system_prompt: str | None = None,
         max_llm_turns: int = 10,
         max_tool_iterations: int = 5,
-        citation_repair_timeout: float = 60.0,
```
KyleZheng1284, line 249: **[P3 compatibility suggestion] Consider retaining this constructor keyword as a deprecated no-op**

Removing `citation_repair_timeout` means a direct caller written against the preceding 2.2 release candidate now receives `TypeError`, even though the repair pass itself is intentionally removed. I found no shipped configuration, public documentation, or production repository caller, so this is an optional compatibility safeguard rather than a merge blocker.

If direct agent constructors are covered by the 2.2 compatibility policy, could we keep the keyword with its existing default, mark it deprecated/ignored, and add a construction-only test? That preserves compatibility without restoring the latency-sensitive repair call.

KyleZheng1284 (commented):
I revalidated these paths against the current PR head with focused deterministic reproductions. The inline comments are written as implementation suggestions. Findings 1, 2, 4, and 5 arise from this PR; finding 3 is an existing shallow-research behavior that conflicts with this PR's stated no-uncited-fallback guarantee; finding 6 is an optional compatibility safeguard for direct callers of the preceding release candidate.

src/aiq_agent/agents/shallow_researcher/agent.py
AjayThorve: this wsa just added in last commit, so safe to remove, as we are getting rid of the citation repair feature

src/aiq_agent/agents/shallow_researcher/agent.py
AjayThorve: Fixed in 80845e6b. `ToolNode` now uses a constant safe error handler instead of serializing `repr(exception)`, while retaining `ToolMessage.status == "error"` so the bounded retry still works. The regression injects dummy secret and internal-endpoint markers, verifies neither reaches the next LLM request, and completes successfully on the rewritten retry.

KyleZheng1284 (approved):
lgtm

AjayThorve:
Based on offline discussion, deciding to close this.

PR #457, opened 2026-08-20
Feature/ds agent fdabench

arc144:
#### DCO sign-off for the squash commit

Signed-off-by: Eduardo Rocha de Andrade <du-ra92@hotmail.com>

#### Validation

- [x] I ran the relevant local checks or explained why they are not applicable.
- [x] I added or updated tests for behavior changes.
- [x] I updated documentation for user-facing or contributor-facing changes.
- [x] I confirmed this PR does not include secrets, credentials, or internal-only data.
- [x] I certify this contribution under the Developer Certificate of Origin (DCO) and signed my commits with `git commit -s` or an equivalent sign-off.
- [x] I replaced the DCO sign-off placeholder with my GitHub commit identity and kept the required angle brackets around the email address.

#### Related Issues

- Relates to #

## Summary by CodeRabbit

* **New Features**
  * Added an autonomous Data Science workflow for interactive analysis, headless evaluation, data retrieval, web search, and persistent Python exploration.
  * Added direct Data Science and FDABench-Lite configuration profiles, hybrid integration, source-aware reporting, citations, and configurable analysis limits.
  * Added a secure, request-scoped Python sandbox for persistent scientific analysis.

* **Bug Fixes**
  * Improved external data-service timeout, retry, and authentication handling.
  * Added support for overriding embedding credentials with `AIQ_EMBED_API_KEY`.
  * Improved workflow and tool availability validation.

* **Documentation**
  * Expanded setup, configuration, prompt, tool, sandbox, and Data Science workflow guidance.

PR #465, opened 2026-08-25
feat: run the data science agent as an async job

ashan-nv:
#### Overview

Runs the Data Science Agent as an async job so a run survives a browser close or
refresh, and routes chat requests to it through the existing Hybrid slot.

A DS run routinely takes several minutes — longer than a tab can be relied on to
stay open. This reuses the existing async job infrastructure (Dask worker, job
store, event store, SSE replay) rather than adding anything new; the agent was
simply never plugged into it.

**Stacked on #457.** This branch is based on that PR's head, so the diff against
`develop` currently includes its commits. Only these 3 are mine:

| Commit | Change |
| :-- | :-- |
| `3132b9b` | Run the DS agent as an async job |
| `191db8b` | Docs: resumable job stream endpoint |
| `17821d3` | Route chat DS requests to an async job |

Please review against `arc144:feature/ds-agent-fdabench`, and merge only after
#457. Happy to rebase onto `develop` once it lands.

#### What changed

**Async execution** — registers `data_science` in the async agent registry; adds
a runner branch so the Dask worker builds the agent with its configured limits
instead of falling through to a generic constructor that silently dropped them;
forces `headless` (a worker has no channel to ask a clarifying question).

**A real bug this surfaced** — the job runner constructs the agent directly and
never calls the NAT registrar, so `validate_research_source_configuration` was
being skipped entirely. An empty source selection previously ran for minutes and
failed at finalization; it now fails immediately with the typed error. Deep
research already guards this inside `agent.py` for the same reason.

**Chat routing** — `hybrid_research_node` awaited DS inline and went straight to
`END`, which would have tied the answer to the open websocket and lost it on a
refresh. It now submits a job when a submitter is configured, mirroring
`deep_research_node`, and still runs inline for the synchronous CLI. Routing
requires three pieces together, including `context_aware_intent_router` — the
plain `intent_classifier` has no `hybrid_research` route, so DS is unreachable
without it.

**UI** — one line to accept a `data_science` escalation kind. The UI branches on
kind only for conversation renaming, so DS reuses the existing streaming,
reconnect, and report rendering unchanged.

Signed-off-by: Ashan Panduwawala <apanduwawala@nvidia.com>

#### Validation

Verified end to end against a live GSF deployment (not mocks):

- Submit returns a `job_id` in **~70ms**; killing the stream mid-run leaves the
  job `running`; reattaching at `/stream/<last_id>` resumes at the next event
  rather than restarting; the report is retrievable later on a fresh connection;
  a client that missed the entire run replays all events.
- Real GSF work throughout — `catalog_search` and `text_to_sql` both executed,
  one run returned 51 rows, answers consistent across runs.
- A database-scoped chat request returned
  `{"type":"job_escalation","kind":"data_science","job_id":...}` with a matching
  `Submitted data_science job` in the backend log.
- **End to end in the web UI.** A chat message routed to DS, submitted a job,
  and streamed to the browser. Refreshing mid-run reattached to the same job and
  resumed with nothing lost:

  ```
  Submitted data_science job 74dc30c9-...
  SSE: Replay complete for job 74dc30c9-... at event_id=431; switching to live mode
  SSE: Fetched 2 events (after_id=431)
  ```

  431 buffered events replayed on reconnect, then live consumption continued.
  The same close/reopen behaviour was separately confirmed on the deep-research
  path, which shares the job store, SSE endpoints, and UI reconnect code.

Commands:

```
uv run pytest tests/aiq_agent/agents/chat_researcher      # 234 passed
uv run pytest frontends/aiq_api/tests/test_data_science_async_job.py  # 13 passed
uv run pytest tests/aiq_agent/agents/data_science         # 40 passed
uv run ruff check . && uv run ruff format --check .
uv run python .agents/skills/aiq-configure-workflow/scripts/validate_config.py \
  configs/config_web_data_science_hybrid.yml              # exit 0
```

One pre-existing failure on the base branch, unrelated and not introduced here:
`test_submit_collision.py::test_submit_records_conversation_id_on_job_access`
patches `aiq_api.jobs.submit._current_conversation_id`, which does not exist.

- [x] I certify this contribution under the Developer Certificate of Origin (DCO).
- [x] I replaced the DCO sign-off placeholder with my GitHub commit identity.

`frontends/aiq_api/src/aiq_api/jobs/runner.py` for the async construction branch,
and `src/aiq_agent/agents/chat_researcher/agent.py` for the inline→job change.

#### Known gaps

- **Routing threshold needs per-deployment tuning.** Unscoped chat requests
  reach DS only when catalog coverage clears `catalog_confidence_threshold`. A
  test deployment measured 0.5 against a representative question: at the 0.6
  default it fell back to shallow, and at 0.4 it routed to DS correctly. The
  config ships the 0.6 default and exposes the knob via
  `AIQ_CATALOG_CONFIDENCE_THRESHOLD`. Scoped requests (`database_name`) always
  take Hybrid regardless of coverage.
- **`stateful_python` sandbox path is unit-tested only** — verifying it needs an
  OpenShell image and policy file.
- **Separate bug, not fixed here:** `password: ${GSF_PASSWORD}` in the DS configs
  and `configuration-reference.md:390` is wrong. That field takes the *name* of
  an env var; `${...}` expands to the secret, so GSF silently registers as
  unavailable. The GSF package's own tests use the bare name.
- **Cosmetic:** the UI banner says "deep research" for any job kind, so a DS run
  is announced as deep research.

#### Related Issues

- Relates to #457

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous Data Science Agent jobs with persisted results, streaming, reconnection, and database-scoped execution.
  * Added hybrid research escalation to Data Science Agent jobs, with inline execution retained when async processing is unavailable.
  * Added `data_science` to the public agent registry, REST API documentation, and chat escalation workflow.
  * Added configurable web and hybrid Data Science profiles.

* **Bug Fixes**
  * Improved sandbox detection and validation of selected research sources.

* **Documentation**
  * Documented asynchronous Data Science Agent execution and API job submission.

### DIFF
--- a/configs/config_web_data_science.yml
+++ b/configs/config_web_data_science.yml
@@ -61,7 +61,10 @@ function_groups:
     auth:
       mode: password
       email: ${GSF_EMAIL}
-      password: ${GSF_PASSWORD}
+      # This field takes the NAME of the env var holding the password, not the
+      # password itself. Using ${GSF_PASSWORD} here resolves to the secret and
+      # GSF then looks up an env var by that value, silently disabling the tools.
+      password: GSF_PASSWORD  # pragma: allowlist secret
     include:
       - catalog_search
       - text_to_sql
--- a/configs/config_web_data_science.yml
+++ b/configs/config_web_data_science.yml
@@ -1,0 +1,125 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Web profile for the Data Science Agent, served as an async job.
+#
+# A DS run can take several minutes, so the agent is submitted to the async job
+# API instead of being held open on a request. Clients POST
+# /v1/jobs/async/submit with agent_type "data_science", then stream progress
+# from /v1/jobs/async/job/{job_id}/stream. Closing or refreshing the browser
+# does not cancel the run: reconnecting to the same job_id replays the stored
+# events and returns the final report.
+#
+# Async jobs run in a Dask worker that reloads this file, so NAT_CONFIG_FILE
+# must point at it and NAT_DASK_SCHEDULER_ADDRESS must be set.
+#
+# Required: INFERENCE_NVIDIA_API_KEY, AIQ_INFERENCE_BASE_URL,
+# AIQ_DATA_SCIENCE_MODEL, TAVILY_API_KEY, GSF_BASE_URL, GSF_EMAIL,
+# GSF_PASSWORD, RAG_SERVER_URL, COLLECTION_NAME.
+
+general:
+  use_uvloop: true
+  telemetry:
+    logging:
+      console:
+        _type: console
+        level: INFO
+
+  front_end:
+    _type: aiq_api
+    runner_class: aiq_api.plugin.AIQAPIWorker
+    db_url: ${NAT_JOB_STORE_DB_URL:-sqlite+aiosqlite:///./jobs.db}
+    expiry_seconds: 86400  # 24 hours (min: 600, max: 604800/7 days)
+    cors:
+      allow_origin_regex: 'http://localhost(:\d+)?|http://127.0.0.1(:\d+)?'
+      allow_methods:
+        - GET
+        - POST
+        - DELETE
+        - OPTIONS
+      allow_headers:
+        - "*"
+      allow_credentials: true
+      expose_headers:
+        - "*"
+
+llms:
+  data_science_llm:
+    _type: openai
+    model_name: ${AIQ_DATA_SCIENCE_MODEL}
+    base_url: ${AIQ_INFERENCE_BASE_URL}
+    api_key: ${INFERENCE_NVIDIA_API_KEY}
+    reasoning_effort: medium
+    max_tokens: 32768
+    num_retries: 2
+    request_timeout: 600
+
+function_groups:
+  gsf:
+    _type: gsf
+    base_url: ${GSF_BASE_URL}
+    auth:
+      mode: password
+      email: ${GSF_EMAIL}
+      password: ${GSF_PASSWORD}
+    include:
+      - catalog_search
+      - text_to_sql
+    connect_timeout_seconds: 10
+    read_timeout_seconds: ${GSF_READ_TIMEOUT_SECONDS:-600}
+    max_retries: 2
+    default_max_rows: 1000
+
+functions:
+  data_sources:
+    _type: data_source_registry
+    sources:
+      - id: structured_data
+        name: "Enterprise Structured Data"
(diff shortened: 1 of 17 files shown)


src/aiq_agent/agents/chat_researcher/register.py
```diff
@@ -604,8 +634,34 @@ async def _submit_deep_job(state: ChatResearcherState) -> str:
...
+                the async worker builds agent state from the query, and the agent
+                re-runs its own bounded catalog discovery.
+                """
+                principal = require_verified_principal()
+                owner = principal.email or principal.sub
+                input_text = _resolve_submission_query(state)
+                if state.clarifier_result:
+                    input_text = f"{input_text}\n\n## Clarification Context\n{state.clarifier_result}"
+
+                return await submit_agent_job(
```
cdgamarose-nv, line 652: This serializes only the latest query and selected sources. A request with `database_name` always routes Hybrid, but the worker reconstructs `DataScienceAgentState` with `database_name=None`, no catalog context, and no conversation history. Database-scoped or contextual follow-up questions can therefore query the wrong/default scope or lose their meaning. Pass structured initial state through the submission/runner boundary and test database-scoped and multi-turn requests.

configs/config_web_data_science.yml
```diff
@@ -0,0 +1,125 @@
...
+    request_timeout: 600
+
+function_groups:
+  gsf:
+    _type: gsf
+    base_url: ${GSF_BASE_URL}
+    auth:
+      mode: password
+      email: ${GSF_EMAIL}
+      password: ${GSF_PASSWORD}
```
cdgamarose-nv: `GSFPasswordAuthConfig.password` expects an environment-variable name and later calls os.environ.get(config.auth.password). `${GSF_PASSWORD}` expands to the secret: punctuation causes schema rejection, while identifier-like values are treated as another missing environment variable and GSF is registered as unavailable. Use `password: GSF_PASSWORD  # pragma: allowlist secret`, matching the Hybrid profile

src/aiq_agent/agents/chat_researcher/register.py
ashan-nv: Good catch — fixed the `database_name` bit, it now gets passed through `submit_agent_job` → `run_agent_job` → `_run_agent` and onto agent state, with tests for the scoped and unscoped cases (`81edbec`).

Left catalog context (the agent just re-derives it, one extra ~5s call) and conversation history (single `input_text` is how every agent job works today, deep research included) as follow-ups rather than reworking the submission contract in this PR.

configs/config_web_data_science.yml
ashan-nv: Fixed in `81edbec` — good catch, I had this right in the Hybrid profile and missed it here.

cdgamarose-nv (approved):
Lgtm!

PR #467, opened 2026-08-25
fix(deep-research): normalize planner verification hints

freshyjmp:
## Summary
- add an optional `Constraint.verification` field to the structured `ResearchPlan` contract
- preserve planner-generated checks for the writer instead of rejecting or discarding them
- retain the normal one-retry structured-response fallback behavior for provider responses returned as JSON text

## Validation
- `uv run pytest tests/aiq_agent/agents/deep_researcher/test_custom_middleware.py -k StructuredResponseTextFallbackMiddleware -q` (11 passed)
- `uv run ruff check src/aiq_agent/agents/deep_researcher/models/subagent_contracts.py src/aiq_agent/agents/deep_researcher/custom_middleware.py tests/aiq_agent/agents/deep_researcher/test_custom_middleware.py`

Observed incident signature: planner retries exhausted after otherwise-valid `ResearchPlan` payloads added one `verification` check to every constraint (six fields in the original capture; seven in the live reproduction).

## Summary by CodeRabbit

- **New Features**
  - Research constraints can now include an optional verification instruction to support concrete checks.
  - Structured research plans accept valid JSON content and preserve supported verification details.

- **Bug Fixes**
  - Improved recovery and validation of structured research-plan responses.
  - Unsupported fields are safely discarded, while invalid or unrecognized responses remain rejected.

- **Tests**
  - Added coverage for JSON normalization, schema validation, supported-field preservation, and unsupported-field removal.

### DIFF
--- a/src/aiq_agent/agents/deep_researcher/custom_middleware.py
+++ b/src/aiq_agent/agents/deep_researcher/custom_middleware.py
@@ -46,7 +46,6 @@ from aiq_agent.common.citation_verification import is_non_citable_status_output
 from aiq_agent.common.logging_utils import log_content_metadata
 
 from .models import ResearchNotes
-from .models import ResearchPlan
 from .resource_limits import DeepResearchResourceLimits
 from .resource_limits import StateBudgetLedger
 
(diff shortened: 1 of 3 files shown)


src/aiq_agent/agents/deep_researcher/custom_middleware.py
```diff
@@ -132,7 +132,11 @@ def _promote(self, response: ModelResponse) -> ModelResponse:
         if not isinstance(message, AIMessage) or message.tool_calls or not isinstance(message.content, str):
             return response
         try:
-            structured = self.schema.model_validate_json(message.content)
+            payload = json.loads(message.content)
```
cdgamarose-nv, line 135: `json.loads()` accepts isolated surrogate escapes such as "\ud800", and `model_validate()` promotes that payload. `PlanPersistenceMiddleware` then crashes with `UnicodeEncodeError` instead of allowing the structured-response correction retry; the former `model_validate_json()` path rejects it correctly. Retain `model_validate_json()` now that the schema supports verification, or explicitly reject invalid Unicode.

PR #468, opened 2026-08-26
feat(research): Add web page fetch tool

smasurekar:
#### Overview

Add `fetch_url_tool`, a NAT source tool that opens exact HTTP(S) URLs through Tavily Extract and returns locally windowed, line-numbered page text. The implementation preserves complete extraction before windowing, rejects non-URL inputs, detects soft 404s, returns non-raising failures, and registers only successfully fetched pages as citable sources.

Register the tool under the `web_search` source in all Tier 1 and registry-based Tier 2 profiles. The frontier-model profile uses its existing explicit tool lists. Install the plugin in development and runtime environments, document the new reader capability, and keep the frozen MCP profile unchanged.

Signed-off-by: smasurekar <smasurekar@nvidia.com>

#### Validation

Start with `sources/web_page_fetch/src/register.py` for URL validation, Tavily extraction, failure handling, and NAT registration, then `sources/web_page_fetch/src/formatting.py` for windowing and scoped citation parsing. Review `configs/config_cli_default.yml` and `configs/config_web_frag.yml` for Tier 1 registry wiring, followed by `deploy/Dockerfile` and `tests/deploy/test_runtime_image_packages.py` for runtime discovery.

#### Related Issues

- N/A

## Summary by CodeRabbit

* **New Features**
  * Added an optional Tavily web-fetch tool with URL validation, configurable limits, line-based navigation, result windowing, and citations.
  * Added configuration for API keys, timeouts, extraction depth, and character budgets.
  * Added handling for duplicate URLs, failed pages, redirects, and partial results.
* **Documentation**
  * Added installation, deployment, configuration, usage, and security guidance, including prompt-injection and data-exfiltration considerations.
* **Bug Fixes**
  * Improved citation handling so unsupported content can continue through fallback processing.


tests/aiq_agent/common/test_citation_verification.py
```diff
@@ -689,6 +689,20 @@ def _parse_custom(content: str, tool_name: str) -> list[SourceEntry]:
...
+        knowable at registration time, without swallowing every other tool's results.
+        """
+
+        def _declines_everything(content: str, tool_name: str) -> list[SourceEntry] | None:
+            return None
+
+        register_source_parser(lambda name: True, _declines_everything)
+        entries = extract_sources_from_tool_result("some_tool", "See https://fallback.example/page")
+        assert [entry.url for entry in entries] == ["https://fallback.example/page"]
+
```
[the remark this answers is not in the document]
smasurekar: Already handled: `tests/aiq_agent/common/test_citation_verification.py:39` has a module-level autouse fixture that snapshots _PARSER_REGISTRY and restores it in place after every test in this file, so `TestParserDispatcher` is already covered.

sources/tavily_web_fetch/src/formatting.py
```diff
@@ -0,0 +1,329 @@
...
+
+    Identity comes from the preamble this module always writes first, not from the tool name: the
+    callable name is the operator's YAML key, which is not available when this parser is
+    registered. Testing for the section marker instead would be too loose -- another tool quoting
+    the marker in a search snippet would have its output claimed, and its own sources discarded.
+    Returning ``None`` declines the content so another parser, or the generic URL fallback,
+    handles it.
+    """
+    if not content.lstrip().startswith(_PREAMBLE):
+        return None
```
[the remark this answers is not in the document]
smasurekar: Fixed: the parser now checks the workflow's own config to see whether the tool that produced the output is really this tool, instead of trusting the preamble text. Any other tool's output is declined, even if it starts with the preamble and contains a marker, new test added, and verified against a real workflow.

docs/source/customization/configuration-reference.md
```diff
@@ -173,6 +173,45 @@ functions:
 | `max_content_length` | `int` | `None` | Truncate each result's content to this many characters. Reduces token usage. |
 | `api_base_url` | `str` | `None` | Optional custom or proxy-compatible Tavily API base URL. A non-empty value is passed to the Tavily client; `None` uses the client's default endpoint. |
 
+### `tavily_web_fetch`
+
+Opens pages you already have the URL for, through the [Tavily](https://tavily.com/) Extract API.
+Search tools find candidate pages; this tool reads one. It is not wired into any shipped config -- add the function
+block and reference it from a `data_source_registry` source (or an agent's `tools` list) to enable it.
```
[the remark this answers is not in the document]
smasurekar: Not applicable — `git grep -ln fetch_url_tool HEAD` matches no file under `configs/`, every shipped `web_search` source lists search tools only, and "Tier 1/Tier 2" profiles don't exist in this repo, so the "not wired into any shipped config" text is correct as written.

sources/tavily_web_fetch/src/register.py
```diff
@@ -0,0 +1,452 @@
...
+
+
+def _instance_is_ours(tool_name: str) -> bool:
+    """Return whether ``tool_name`` is a YAML key built from this tool's config.
+
+    The name is unavailable while this package is being built -- NAT assigns it after the build
+    function returns -- but it is available by the time a parser runs, which is what makes this
+    check possible at all. Anything unknown is treated as not ours.
+    """
+    for owner in _owning_builders:
```
KyleZheng1284: **Blocking suggestion:** Could we scope this ownership check to the workflow/function instance that actually produced the tool result?

`_owning_builders` is process-global, so this loop returns `True` when *any* live builder has a `TavilyWebFetchToolConfig` under the same YAML key. Because the parser is also registered globally with a catch-all matcher, two simultaneous workflows can collide:

1. Workflow A registers the real fetch tool as `fetch_url_tool`.
2. Workflow B has an unrelated tool with the same key and emits the published fetch preamble plus a well-formed `<fetched_page>` marker.
3. A's builder makes `_instance_is_ours("fetch_url_tool")` return `True` for B's result.
4. B's forged URL is accepted into the verified citation sources even though the fetch tool never visited it.

I reproduced this through the real async `tavily_web_fetch` lifecycle and production `extract_sources_from_tool_result`; the result was:

```text
real_dispatch_cross_workflow_sources=['https://attacker.example/forged']
```

I think this is merge-blocking because the check is the authorization boundary for citation provenance, and tool-name equality is not unique across concurrent workflows. One possible fix is to carry the producing workflow/function identity into parser dispatch, or otherwise keep parser ownership per workflow/session, rather than allowing any live builder to vouch for the name. A regression with two concurrent workflows using the same key—only one backed by `TavilyWebFetchToolConfig`—would demonstrate the boundary.

KyleZheng1284 (commented):
I found one blocking citation-ownership issue during the live E2E and concurrency review. Details and a suggested direction are attached inline.

sources/tavily_web_fetch/src/register.py
smasurekar: Addressed. Ownership is now scoped to the workflow run that produced the result.

`_owning_builders` and `_instance_is_ours` are removed, so the collision breaks at step 3: nothing vouches by name. Each call records the pages it read under its own run (`_run_ledgers` keyed `{workflow_run_id: {result: pages read}}`, id from `Context.get().workflow_run_id`); no other run's ledger is consulted, and citable URLs come from that record rather than from re-parsing the output (`parse_fetched_pages` removed). Your repro returns `[]` in place of `['https://attacker.example/forged']`.

Regression coverage:

- `test_a_forgery_from_a_workflow_without_this_tool_is_rejected`: the case you specified, two live workflows sharing the key with only one backed by `TavilyWebFetchToolConfig`.
- `test_a_second_workflow_cannot_replay_this_ones_recorded_result`: a second run replaying the first run's byte-identical result.
- `test_concurrent_workflows_cannot_vouch_for_each_others_results`: two runs live concurrently under the same key, neither claiming the other's result.
- `test_a_call_outside_a_workflow_run_is_not_citable`.

The last three fail when `_current_run_id()` is stubbed to a constant, so they pin the boundary itself rather than a particular forged string.

Two caveats:

1. **Replay still works within a single run.** Another tool in the same session can copy the fetch output and get that record. But the page was already fetched and credited in that same session, so nothing new is added and no unvisited page can be named.

2. **The run id is not a secret we control.** NAT usually generates a random run id, but `runtime/session.py:666` will instead use one supplied in a `workflow-run-id` request header, so a request carrying someone else's id is treated as part of their session. It means guessing a random UUID, and the header is NAT's tracing behavior rather than something this package can turn off.

Are you okay with both caveats as they stand, or would you like either one addressed in this PR?

sources/tavily_web_fetch/src/register.py
```diff
@@ -0,0 +1,548 @@
...
+        SourceEntry(url=url, title=title or None, source_type="url", tool_name=tool_name) for url, title in recorded
+    ]
+
+
+async def _extract(tool, urls: list[str], *, extract_depth: str, timeout_seconds: int) -> tuple[list[dict], str]:
+    """Call Tavily Extract without relevance filtering and return results or an error."""
+    try:
+        # Do not forward query: local windowing must operate on the complete extraction so misses are recoverable.
+        payload = {"urls": urls, "extract_depth": extract_depth}
+        response = await asyncio.wait_for(tool.ainvoke(payload), timeout=timeout_seconds)
```
KyleZheng1284: **Blocking suggestion: keep the inner Tavily call from registering outbound links as sources.**

This nested `TavilyExtract` invocation inherits the workflow callbacks. AI-Q then sees the inner tool name (`tavily_extract`) as a search tool and independently extracts every URL from its raw result. I reproduced the full downstream path with the production callback: fetching `https://example.com/` emitted `citation_source` artifacts for both the fetched page and the never-fetched `https://iana.org/domains/example`; parent-report event resolution then included both URLs. The new run-scoped citation parser correctly keeps only the fetched page, so that fix does not protect this separate durable-artifact path, and the observed behavior contradicts the README promise that outbound links are not registered.

The smallest fix I recommend is to prevent the inner provider adapter from inheriting user-facing callbacks (for example, pass an explicitly empty callback config here after verifying LangChain behavior), because the outer `fetch_url_tool` call already owns lifecycle reporting and authoritative citations. Please add an integration regression using `AgentEventCallback` that proves an outbound URL inside `raw_content` never becomes a `citation_source`. If nested provider observability must remain, the alternative is to make the callback consume the authoritative source parser/registry instead of generically scraping every URL from nested output.

sources/tavily_web_fetch/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+The URLs an agent fetches, and the content returned for them, transit third-party infrastructure.
+Review the provider's published data retention, privacy, and model-training terms for your own
+account and contract; terms negotiated by anyone else do not apply to your deployment.
+
+### Secrets and Logging
+
+The API key is held as a `SecretStr` and read from `TAVILY_API_KEY` when not set in config. Without
+a key the tool registers a stub that returns an error string, so a missing secret degrades
+gracefully instead of crashing. Extraction failures log the exception class name only -- never the
```
KyleZheng1284: **Blocking security/documentation suggestion: disclose Relay's full-payload logging.**

This statement is accurate for this module's Python logger, but it is not the complete logging contract. The tracked web workflow enables append-mode Relay ATOF, and Relay defaults `enable_full_payloads` to `true`. In our real UI run, the tool-start event retained the complete 114-character requested URL and the tool-end event retained all 10,922 characters returned to the model. That matters because this tool accepts arbitrary URLs, including signed download URLs, query credentials, internal document identifiers, or user information; the built-in PII detectors are not guaranteed to recognize arbitrary signed-URL tokens.

My preferred fix is explicit security documentation because full-payload observability is system-wide and appears intentional: state that the requested URL and returned content may be stored or exported, warn against credential-bearing URLs, and document the supported controls (`enable_full_payloads: false`, disabling ATOF/export, or request-level telemetry redaction) plus access/retention responsibilities. If that retention is not intentional for this tool, add tool-specific redaction before export instead. A focused test should pin whichever contract is chosen.

KyleZheng1284 (commented):
Thanks for addressing the earlier cross-workflow parser-ownership collision. Two separate end-to-end boundaries would still probably need resolution before merge: durable citation artifacts emitted by the nested Tavily callback, and the undisclosed Relay full-payload logging contract.

sources/tavily_web_fetch/src/register.py
smasurekar: Confirmed and fixed as recommended. The adapter now takes an explicitly empty callback config, so it emits no lifecycle and `fetch_url_tool` keeps sole ownership of reporting and citations. Verified against the installed `langchain-core`: `ensure_config` overrides on `v is not None`, `_configure` gates on truthiness, so `[]` displaces the inherited manager.

Regression in `tests/test_callback_integration.py`: a real `BaseTool` named `tavily_extract` under a live `AgentEventCallback` with your `example.com` / `iana.org` pair (a `MagicMock` would pass either way). Asserts the outbound URL never becomes a `citation_source`, and that the adapter emits no `tool.start`/`tool.end` at all so a LangChain config-merge change fails loudly. A positive control proves the harness can fail. Both guards failed before the fix.

One residual: `_is_search_tool` matches the operator-chosen YAML key, so naming the block `tavily_fetch` would scrape the *outer* output. Documented (README `Citation Scoping` + config reference) rather than changing shared callback behavior. Happy to take your parser-registry alternative instead.

sources/tavily_web_fetch/README.md
smasurekar: Disclosed as preferred, but the test corrected two things - one of them a control from your comment.

**`enable_full_payloads: false` does not remove this data.** Measured against a live Relay runtime: the trace is byte-identical to default. AI-Q passes tool payloads as explicit scope values, which that setting does not govern. Now a documented negative claim with a test.

**Activation was the undisclosed part.** Request privacy is per-request via `x-aiq-telemetry-redact: true`; `request_privacy_attributes` enables nothing on its own. Documented in neither the README nor `observability.md`; now in both. Fail-open interaction: `redaction.enabled: false` silently disarms the header.

Measured: retained by default; retained with `enable_full_payloads: false`; retained under privacy when redaction is off; removed under privacy; no trace with `atof.enabled: false`.

`tests/test_relay_fetch_tool_retention.py` (6 tests) fetches through the real tool and reads the ATOF back for a signed-URL token and a page-content canary - present by default, absent under the control, each non-control pinned as a negative. One test walks header -> tag -> redacted trace, so the documented instruction is what is tested.

Both behaviors are system-wide, so I documented rather than changed them. The fail-open one looks like a separate defect - say the word and I'll file it.

PR #472, opened 2026-08-27
feat: add Harbor evaluation integration

1ytic:
#### Overview

Add a standalone Harbor evaluation integration for AI-Q under `frontends/evaluation`.

- Add the host-side `AiqHarborAgent` runner, AI-Q workflow profiles, Harbor job configs, and image build helper.
- Add Harbor adapters and task verifiers for DeepSearchQA and DeepResearch Bench II.
- Add focused unit tests, workspace packaging and lock metadata, and contributor documentation for generating tasks and running evaluations.

The evaluation workspace is intentionally independent from the root AI-Q uv workspace while reusing the canonical AI-Q runtime image build inputs.


frontends/evaluation/adapters/deepresearch-bench-ii/src/deepresearch_bench_ii/task-template/environment/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM aiq-harbor:local
```
cdgamarose-nv, line 1: FROM aiq-harbor:local resolves a mutable local tag. I built and ran a trial, and Harbor recorded the task checksum and aiq-agent==2.2.0, but neither the base image ID nor Git revision appeared in the job artifacts; a different checkout can therefore produce indistinguishable result provenance. Tag the image with an immutable revision/digest, reject or identify dirty builds, reference that immutable tag from both task templates, and persist it in the run metadata.

frontends/evaluation/adapters/deepsearchqa/src/deepsearchqa/task-template/tests/verifier.py
```diff
@@ -0,0 +1,350 @@
...
+    logging.basicConfig(level=logging.INFO)
+    verifier_dir = Path("/logs/verifier")
+
+    try:
+        metadata = json.loads(args.metadata_path.read_text(encoding="utf-8"))
+        answer_text = args.answer_path.read_text(encoding="utf-8").strip()
+        if not answer_text:
+            raise ValueError("Agent answer file is empty.")
+        result = evaluate(metadata, answer_text)
+    except Exception as exc:
```
cdgamarose-nv, line 333: This catches missing credentials, API failures, and parsing failures, writes a zero reward, and exits successfully. I reproduced this in the built container with no GEMINI_API_KEY: the verifier exited 0 with grader_valid: 0, so an external user can mistake 900 infrastructure failures for valid zero-scoring answers. Preserve the diagnostic artifact but exit nonzero whenever grading is invalid; apply the same contract to DRB-II’s result_payload["error"] path.

frontends/evaluation/adapters/deepresearch-bench-ii/src/deepresearch_bench_ii/task-template/tests/verifier.py
```diff
@@ -0,0 +1,171 @@
...
+    dimensions = aggregate_scores.compute_dimension_averages(result_payload)
+    rubric_item_count = 0
+    scores = result_payload.get("scores", {})
+    if isinstance(scores, dict):
+        for dimension in ("info_recall", "analysis", "presentation"):
+            dimension_scores = scores.get(dimension)
+            if isinstance(dimension_scores, dict):
+                rubric_item_count += len(dimension_scores)
+
+    return {
```
cdgamarose-nv, line 94: The result records chunking settings but omits GEMINI_MODEL and which API/provider handled the request; the DeepSearchQA result has the same gap. Since both model and endpoint are configurable, two materially different judges currently produce results with identical recorded provenance. Record the non-secret model ID, provider/endpoint identity, scorer revision, and grader dependency version in grading.json.

frontends/evaluation/pyproject.toml
```diff
@@ -0,0 +1,63 @@
...
+]
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/aiq_harbor"]
+
+[tool.pytest.ini_options]
```
cdgamarose-nv, line 29: The workspace’s default pytest configuration discovers only the 20 tests/ cases and silently omits both adapter suites. After installing the adapters, collecting all suites together also fails because both adapter modules are named test_adapter.py without importlib mode; root CI excludes this standalone workspace entirely. Include both adapter paths, enable --import-mode=importlib, ensure all workspace packages are installed, and add a frozen-lock CI lane for this project.

frontends/evaluation/adapters/deepsearchqa/src/deepsearchqa/task-template/tests/test.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -euo pipefail
+
+python -m pip install --no-cache-dir --quiet google-genai==1.73.1 requests
```
cdgamarose-nv, line 4: Every DeepSearchQA trial installs and downgrades google-genai from PyPI at verification time; the 900-task default therefore performs this network-dependent installation 900 times.
DRB-II also installs an unpinned requests. Put the grader dependencies in a locked verifier/task-image layer so a completed agent run cannot become unverifiable because PyPI is unavailable or dependency resolution changes.

frontends/evaluation/configs/deepsearchqa.yaml
```diff
@@ -0,0 +1,27 @@
+jobs_dir: jobs
+n_attempts: 1
+timeout_multiplier: 1.0
+n_concurrent_trials: 10
```
cdgamarose-nv, line 4: This runs 900 tasks at 10-way concurrency, while each task requests 4 GB of memory, implying up to roughly 40 GB plus simultaneous model/search/grader traffic. This may exceed many external Docker allocations and common API rate limits, potentially after incurring substantial cost. Wondering if we should default to smaller concurrency and document dataset size, resource/cost expectations, and how to opt into larger runs.

frontends/evaluation/src/aiq_harbor/runner.py
```diff
@@ -0,0 +1,787 @@
...
+_DEEP_RESEARCH_SUBAGENTS = frozenset(
+    {
+        "planner-agent",
+        "researcher-agent",
+        "source-router-agent",
+        "writer-agent",
+    }
+)
+_MISSING = object()
+_LLM_RUNTIME_FIELDS = (
```
cdgamarose-nv, line 51: `runner.py `manually lists which settings to save for each LLM and function type. Does this mean every time AI-Q adds a new function type or configuration field, this adapter must also be updated?
The workflow will still run because NAT loads the complete config, but the evaluation record will silently omit any setting that is not in these lists. This is already happening:
`paper_search` and `deep_research_agent` have no entries, while fields such as `parallel_tool_calls` and `chat_template_kwargs` are also missing.

Because the uploaded config is not preserved after the Harbor environment is deleted, we may be unable to determine the exact configuration used for a completed evaluation. That
makes the result difficult to reproduce.

Can we avoid maintaining these lists and instead save a sanitized version of the complete resolved config, with secrets removed? This would allow new AI-Q fields and function types
to work without requiring changes to the Harbor adapter.

cdgamarose-nv (changes requested):
Will you be taking care of removing the legacy NAT eval benchmarks as well?

frontends/evaluation/adapters/deepresearch-bench-ii/src/deepresearch_bench_ii/task-template/environment/Dockerfile
1ytic: Yes, I keep this "local" version intentionally to simplify dataset dependencies. It simplify development and research, you can make changes in AIQ, rebuild docker image and use same Harbor dataset. But if you insist we can use latest git revision, which also kind of chicken-egg problem.

frontends/evaluation/adapters/deepsearchqa/src/deepsearchqa/task-template/tests/verifier.py
1ytic: Yes it is. grader_valid metric used exactly to monitor failed verifications. If I'm not mistaken harbor evaluation will failed if verifier failed.

frontends/evaluation/adapters/deepresearch-bench-ii/src/deepresearch_bench_ii/task-template/tests/verifier.py
1ytic: I think this is ok. It replicates the official verification script. I don't think we should record additional metadata.

frontends/evaluation/adapters/deepsearchqa/src/deepsearchqa/task-template/tests/test.sh
1ytic: Yes, and I believe this is a common practice in Harbor framework. Agent and verifier can install dependencies.

frontends/evaluation/configs/deepsearchqa.yaml
1ytic: It's up to user to change this. In our local testing it was fine.

frontends/evaluation/src/aiq_harbor/runner.py
1ytic: This used only for logging, but I agree, we need align it with used tools. I'll update this part now, but I'm not sure how to make it general.

frontends/evaluation/pyproject.toml
1ytic: I add these benchmarks tests and add CI root coverage.

PR #475, opened 2026-08-28
fix: improve shallow agent source selection prompt

niyatisingal:
#### Overview

Based on evaluation results on an internal custom dataset (financebench-hybrid) that mixes questions needing a web search, questions that should be answered from the knowledge base, and questions that require both sources, the shallow agent strongly prefers the web tool over knowledge search.

That bias came from the source hierarchy in the shallow researcher prompt: knowledge_search was framed as “my files / uploaded docs,” and web search was the default for general facts. Public-sounding questions therefore went to Tavily even when the answer was in the indexed collection.

This change updates source selection so that if a knowledge or document tool is listed, the agent uses it first for anything that could live in that collection (named entities, historical records, exact figures, definitions), including public-sounding questions and even when the user never says “my files.” Web search is reserved for live news and for cases where the knowledge tool is missing, empty, or off-topic. Tools are matched by capability, not by a fixed name.

With these prompt changes, overall F1 score improved from ~0.37 to ~0.49 (with nemotron-3.5-lightning model) on the custom dataset.

Signed-off-by: Niyati Singal <nsingal@nvidia.com>

#### Validation

## Improvements

- Research responses now select information sources based on the question and available tools.
- Questions about your knowledge corpus prioritize relevant knowledge and document sources, with primary documents preferred over general web results.
- Questions about uploaded files use the available document-search option and avoid web or academic sources.
- When document-search tools are unavailable, research automatically falls back to web search and provides cited responses.


src/aiq_agent/agents/shallow_researcher/prompts/researcher.j2
```diff
@@ -95,5 +97,5 @@ The user has uploaded the following documents to the knowledge base:
 - **{{ doc.file_name }}**: {{ doc.summary or "No summary available" }}
 {% endfor %}
 
-**CRITICAL**: When the user asks to summarize, describe, explain, or query their uploaded documents, you MUST use the **knowledge_search** tool — NOT web search. Only **knowledge_search** has access to these uploaded documents.
+**CRITICAL**: When the user asks about these documents, you MUST use the knowledge/document tool — NOT web search. Only that tool can read the uploaded files. The knowledge tool is still appropriate for other questions that may be in the same collection.
```
tanleach: Will this unintentionally only use documents uploaded and never allowed for any web search to go hand in hand? Have we validated that asking a question about a document that might require some web search still works?

tanleach (changes requested):

src/aiq_agent/agents/shallow_researcher/prompts/researcher.j2
niyatisingal: yes, I validated on the custom dataset that for a question requiring both web search and knowledge search, the shallow agent indeed calls both to get the final answer (after these changes)

PR #478, opened 2026-08-31
fix: Add retry in case of empty answers

niyatisingal:
#### Overview

1. Shallow researcher: If synthesis has no visible text after sanitization, retry once without tools and with thinking off; a second empty result raises shallow_research_empty_synthesis.
2. Synthesis prompt: Instructs the model to drop fake tool-call XML/JSON and keep only the prose report.
3. Citation verification: sanitize_report() strips leaked tool-call XML/JSON while keeping surrounding prose; markup-only answers still become empty.
4. Post-processing: Citation checks use last_msg.text instead of raw content.
5. Tests/docs: Cover the retry, fail-closed path, sanitizer cases, and the new behavior.

Signed-off-by: Niyati Singal <nsingal@nvidia.com>

#### Validation

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved research responses when synthesis returns empty, times out, or contains only tool-call content.
  - Automatically retries synthesis with fallback handling and reports a clear error when no usable content can be generated.
  - Removes leaked tool-call markup from XML and JSON responses while preserving surrounding report content.
  - Added configurable synthesis timeouts and improved handling of cancelled or delayed requests.

- **Tests**
  - Expanded coverage for retries, timeout handling, fallback behavior, error handling, and content sanitization.


src/aiq_agent/agents/shallow_researcher/agent.py
```diff
@@ -329,6 +391,80 @@ def _get_llm(self) -> BaseChatModel:
...
+            if task in done:
+                return task.result()
+            task.cancel()
+            settle_timeout = min(SYNTHESIS_CANCEL_SETTLE_TIMEOUT, timeout)
+            done, _pending = await asyncio.wait({task}, timeout=settle_timeout)
+            if task in done:
+                if task.cancelled():
+                    raise TimeoutError
+                return task.result()
+            task.add_done_callback(_discard_later_result)
```
cdgamarose-nv: This can leave model calls running after the request has failed. If cancellation does not settle within one second, we stop waiting for the task but do not actually stop it. The retry loop can then start another generation while the previous one is still consuming a connection and model capacity. I don’t think we should solve a timeout by detaching the work; the timeout needs to cancel the provider request, or the request should remain responsible for the active task.

src/aiq_agent/agents/shallow_researcher/agent.py
```diff
@@ -436,17 +572,76 @@ async def agent_node(state: ShallowResearchAgentState) -> dict[str, Any]:
...
+                            except TimeoutError:
+                                logger.warning(
+                                    "Forced synthesis timed out after %.3fs without a thinking budget; "
+                                    "treating as empty content (attempt=%d/%d)",
+                                    self.synthesis_attempt_timeout,
+                                    attempt + 1,
+                                    SYNTHESIS_EMPTY_CONTENT_RETRIES + 1,
+                                )
+                                continue
+                        if _visible_message_text(response):
```
cdgamarose-nv: This is not checking the content that the user will actually see. A response can pass _visible_message_text() and still become empty when sanitize_report() runs later. For example, a response containing only a bare URL passes this check and is then sanitized to an empty string. Could we use the final sanitizer itself to determine whether the response is visible? LangChain’s existing message.text property can also handle the structured-content extraction here.

src/aiq_agent/agents/shallow_researcher/agent.py
```diff
@@ -63,6 +64,18 @@
 # Path to this agent's directory (for loading prompts)
 AGENT_DIR = Path(__file__).parent
 
+# Extra synthesis LLM calls after the first empty visible answer.
+SYNTHESIS_EMPTY_CONTENT_RETRIES = 5
+SYNTHESIS_RETRY_REASONING_BUDGET = 16384
```
cdgamarose-nv: The fixed reasoning budget does not work for all shipped configurations. Several shallow-research profiles use max_tokens: 8192, so a reasoning budget of 16,384 does not prevent reasoning from consuming the entire available completion budget. Other providers may reject this parameter and fall back to the original behavior. A single retry with thinking disabled would be simpler and would address the actual failure mode more directly.

src/aiq_agent/agents/shallow_researcher/agent.py
```diff
@@ -436,17 +572,76 @@ async def agent_node(state: ShallowResearchAgentState) -> dict[str, Any]:
...
+                                    attempt + 1,
+                                    SYNTHESIS_EMPTY_CONTENT_RETRIES + 1,
+                                )
+                                continue
+                        if _visible_message_text(response):
+                            return {"messages": [response], "tool_iterations": iterations}
+
+                    raise RuntimeError(
+                        "shallow_research_empty_synthesis: forced synthesis returned empty content "
+                        f"after all the retries"
```
cdgamarose-nv: This currently fails Ruff

cdgamarose-nv (changes requested):
I think this may be more complicated than the bug requires. We already handle the same case in src/aiq_agent/agents/data_science/agent.py. Can we try to follow that pattern here and benchmark the results?
- Use message.text and sanitize_report() to check whether the response has visible content.
- If it is empty, retry once without tools and with thinking disabled.
- If the retry is still empty, raise shallow_research_empty_synthesis.

That prevents blank responses while letting us remove the retry loop, timeout/cancellation machinery, hard-coded constants, custom content parser, and broad JSON stripping.

The long comments and docstrings should also be trimmed and tests should be reduced to what we need: normal response, empty response retries once, successful retry, and second empty response fails. General model timeouts should be handled separately in the shared invocation layer

src/aiq_agent/agents/shallow_researcher/agent.py
niyatisingal: Removed this

src/aiq_agent/agents/shallow_researcher/agent.py
niyatisingal: Corrected it

src/aiq_agent/agents/shallow_researcher/agent.py
niyatisingal: Removed this

src/aiq_agent/agents/shallow_researcher/agent.py
niyatisingal: Removed this

